PR #11, opened 2018-07-23
Test suite ported from json-ld.org.

gkellogg:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/11.html" title="Last updated on Jul 24, 2018, 9:23 PM GMT (5c10380)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/11/466065d...5c10380.html" title="Last updated on Jul 24, 2018, 9:23 PM GMT (5c10380)">Diff</a>


davidlehn:
I'm guessing the .htaccess file here won't work.  What's the plan for parts that use those features?

gkellogg:
If we sync it to w3.org and update the expected base IRIs in some cases, it would work. Just keeping it there to preserve for now.

davidlehn:
If we're moving all these files, might be a good time to consider using subdirs.  Tools like github are not so great at handling a 1000s+ files per dir.

gkellogg:
> If we're moving all these files, might be a good time to consider using subdirs. Tools like github are not so great at handling a 1000s+ files per dir.

Sure, easy enough, I'll take care of it.

PR #20, opened 2018-08-16
Fix manifest baseIri URLs.

davidlehn:
Remove duplicate extra trailing path segments so relative resource URLs
can be properly constructed.

This was causing issues in jsonld.js and makes baseIri usage uniform with the other manifests.

### DIFF
--- a/tests/fromRdf-manifest.jsonld
+++ b/tests/fromRdf-manifest.jsonld
@@ -4,7 +4,7 @@
   "@type": "mf:Manifest",
   "name": "Transform RDF to JSON-LD",
   "description": "Transform RDF to JSON-LD tests take N-Quads input and use object comparison.",
-  "baseIri": "https://w3c.github.io/json-ld-api/tests/fromRdf/",
+  "baseIri": "https://w3c.github.io/json-ld-api/tests/",
   "sequence": [
     {
       "@id": "#t0001",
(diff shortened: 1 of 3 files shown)


tests/expand-manifest.jsonld
```diff
@@ -547,7 +547,7 @@
       "name": "expandContext option",
       "purpose": "Use of the expandContext option to expand the input document",
       "option": {
-        "expandContext": "0077-context.jsonld"
+        "expandContext": "expand/0077-context.jsonld"
```
gkellogg: I think I accepted this too soon. The context sets the type for `expandContext` to `xsd:string`, and this is supplied as an argument to the processor, so the processor would consider it as an API option, and expand relative to the input file. To do it this way would mean changing the type for `expandContext` to `@id`, and the test runner would be expected to turn these into absolute IRIs. We need to do one or the other.

The API describes this as a **USVString** or **dictionary**. Another way to handle it would be for the test runner to open this and provide the resulting dictionary as an argument to the processor, which is typically what happens with input and context files. We need to decide, and document in the README.

BigBlueHat: > We need to decide, and document in the README.

Should we revert this and address the issue here? or just file another issue?

gkellogg: I can fix this in another PR, but I'd like to get @davidlehn's buy-in first.

davidlehn: Ok, This line should likely be reverted.  The jsonld.js test code was looking for that option and considering it as a path relative to baseIri, reading in the file, coverting to json, and passing the json to the api.  But internally the lib handles loading the original string just fine.  So just removing the option handling nonsense fixes it.

PR #29, opened 2018-08-26
Test NQuads character escapes in fromRdf and toRdf.

gkellogg:
Fixes json-ld/json-ld.org#634.


gkellogg:
@davidlehn there may be some assumptions in here about when a JSON serializer will insert escape sequences, and when just use the original character. Let me know what you think, and we can ignore things that are subject to the vagaries of JSON serializers. The primary thing was to test NQuads handling in jsonld.js IIRC.

davidlehn:
Looks like nt17 test was left out of toRdf.  With latest in-progress jsonld.js the only tests that failed were fromRdf 16 and 17.  Output was escaped 0x6f vs a plain char.  Is there a rule about when to coerce unicode escapes into plain chars?

gkellogg:
ToRdf nt17 left out intentionally, as JSON-LD doesn’t have anything like the upper-case U escape for Unicode that N-Wuads has.

Output for fromRdf tests was based on what the Ruby JSON serializer does; I’m not sure how/if we should require this, unless a spec reference is explicit.

gkellogg:
From RFC8259 (JSON):

> The representation of strings is similar to conventions used in the C
   family of programming languages.  A string begins and ends with
   quotation marks.  All Unicode characters may be placed within the
   quotation marks, except for the characters that MUST be escaped:
   quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F).

Arguably, the JSON inputs that use unescaped control characters should be changed to escaped in the toRdf tests. For 16 and 17, it's not necessary that they be escaped or unescaped, AFAIK; perhaps we should either eliminate the tests as duplicative, or use a different character that must be escaped.

davidlehn:
With latest updates, and a quick patch for support of that PositiveSyntaxTest type, jsonld.js work-in-progress branch passes these tests.  I'm guessing no failures means it worked.  There could probably be NegativeSyntaxTests but maybe that's a bit too much scope creep unless there is something specific worth testing.

PR #36, opened 2018-09-22
Improved examples

gkellogg:
Some of the tabbed examples from the syntax document, but generally not as necessary here. Improves other examples and fixes some problems.


davidlehn:
In the now-live version:
- Example 3 is the only one with tabs, and if you click on a tab then back to original, it's empty.
- Are the YAML files used anywhere?
- Not to start a fight, but any reason for .yml vs .yaml? ;-)

gkellogg:
That’s the only example where it seemed to make sense to use the tabs.

YAML files aren’t used, but will eventually be source for a publication on YAML-LD.

.yml seems to be the convention, for example, .travis.yml.

gkellogg:
Something else is off on example 3, as you should be able to click back, probably a mis-spelled class name.

davidlehn:
A web search shows more YAML extension confusion than convention.  Old YAML FAQ says .yaml.  Many projects use .yaml (appengine, kubernetes, etc) but often support .yml.  And yet also many projects use .yml, maybe a misguided idea that DOS compatibility is important in this case?  I dunno.

gkellogg:
I'm happy to change, but the files are only for our internal purposes, at this point. Change would go in common/extract-examples.rb:

https://github.com/w3c/json-ld-api/blob/20940b6740a59c29f1aaaf05b82f84d31f5dd742/common/extract-examples.rb#L504

PR #41, opened 2018-10-01
Algorithm updates and tests to allow `@container: @set` on @type

gkellogg:
or an alias in 1.1. Remains an error to redefine `@type` in 1.0.

For w3c/json-ld-syntax#34.

### DIFF
--- a/index.html
+++ b/index.html
@@ -5492,7 +5492,7 @@
     <li>The <a href="#generate-blank-node-identifier">Generate Blank Node Identifier algorithm</a>
       has been updated to remove the specifics of how new blank node
       identifiers are created.</li>
-    <li>Values of <code>@type</code>, or an alais of <code>@type</code>, may now have their <code>@container</code> set to @set
+    <li>Values of <code>@type</code>, or an alais of <code>@type</code>, may now have their <code>@container</code> set to <code>@set</code>
       to ensure that <code>@type</code> members are always represented as an array. This
       also allows a term to be defined for <code>@type</code>, where the value MUST be a <a>dictionary</a>
       with <code>@container</code> set to <code>@set</code>.</li>
--- a/index.html
+++ b/index.html
@@ -1125,7 +1125,7 @@
     <section class="informative">
       <h3>Overview</h3>
 
-      <p><a>term definitions</a> are created by
+      <p><a>Term definitions</a> are created by
         parsing the information in the given <a>local context</a> for the
         given <a>term</a>. If the given <a>term</a> is a
         <a>compact IRI</a>, it may omit an <a>IRI mapping</a> by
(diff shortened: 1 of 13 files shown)


index.html
```diff
@@ -5489,6 +5492,10 @@ <h2>Changes since JSON-LD Community Group Final Report</h2>
     <li>The <a href="#generate-blank-node-identifier">Generate Blank Node Identifier algorithm</a>
       has been updated to remove the specifics of how new blank node
       identifiers are created.</li>
+    <li>Values of <code>@type</code>, or an alais of <code>@type</code>, may now have their <code>@container</code> set to @set
```
davidlehn: Wrap `@set` in `<code>`.

tests/compact-manifest.jsonld
```diff
@@ -869,6 +869,33 @@
...
+      "@id": "#t0105",
+      "@type": ["jld:PositiveEvaluationTest", "jld:CompactTest"],
+      "name": "Compact @type with @container: @set using an alias of @type",
+      "purpose": "Ensures that a single @type value is represented as an array",
+      "input": "compact/0105-in.jsonld",
+      "context": "compact/0105-context.jsonld",
+      "expect": "compact/0105-out.jsonld",
+      "option": {"processingMode": "json-ld-1.1", "specVersion": "json-ld-1.1"}
+    }, {
+      "@id": "#t0106",
```
davidlehn: Won't this test be a negative test in 1.0 mode?

tests/expand-manifest.jsonld
```diff
@@ -1180,6 +1180,13 @@
       "purpose": "Verifies that an exception is raised in Expansion when an invalid set or list object is found",
       "input": "expand/e041-in.jsonld",
       "expect": "invalid set or list object"
+   }, {
+      "@id": "#te042",
+      "@type": [ "jld:NegativeEvaluationTest", "jld:ExpandTest" ],
+      "name": "Keywords may not be redefined",
+      "purpose": "Verifies that an exception is raised on expansion when processing an invalid context attempting to define @container on a keyword",
+      "input": "expand/e042-in.jsonld",
+      "expect": "keyword redefinition"
```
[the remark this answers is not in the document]
gkellogg: I added a test to be sure that a 1.0 processor would fail if `@type` were so defined. We didn’t have good relevant tests before.

tests/compact-manifest.jsonld
gkellogg: I think the tests are reasonable to check both 1.0 and 1.1 modes. The negative test is in expansion, to make sure a 1.0 processor fails. A 1.0 processor compacting with an alias of `@type` will allow a `@container` definition but ignore it when actually compacting.

davidlehn: Ah, I thought it failed in 1.0.

tests/expand-manifest.jsonld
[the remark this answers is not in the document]
gkellogg: If `processingMode` is not set in the test definition, it defaults to 1.0. expand/e042-in.jsonld is the following:

```json
{
  "@context": {
    "@type": {"@container": "@set"}
  },
  "@type": "http://example.org/type"
}
```

which a 1.0 processor should fail on.

tests/compact-manifest.jsonld
gkellogg: We actually never checked that keyword aliases should only be for simple term definitions; we just ignore everything else. We could consider tightening these up in 1.1, and require that keyword aliases only be simple term definitions.

tests/expand-manifest.jsonld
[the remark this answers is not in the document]
davidlehn: Yeah, I was forgetting it defaulted to 1.0.

PR #43, opened 2018-10-06
Echidna configuration and Travis configuration for publication

gkellogg:
when pushed to "publications" branch.


gkellogg:
The last thing to do is to remove the "dry-run" configuration, which we can do after the first time pushing to the "publications" branch to verify it's working properly without doing a dry run.

PR #44, opened 2018-10-08
Fix duplicate name in `tests/expand-manifest.jsonld`

stephank:
The expand fixtures otherwise appear clean, but haven't tested the other test categories yet. This sort of thing currently fails in certain JSON parsers (serde in my case).


iherman:
+1 to merge.

This is a purely editorial change, so we can bypass the IPR issue (that being said, @stephank, it may be worth going through the process that you were notified about, in case you plan to propose substantial changes in future). I let @gkellogg decide and act on this.

Thanks @stephank

PR #59, opened 2019-01-23
RFC: benchmarks

davidlehn:
This is a first try at something to be used for benchmarks.
- Benchmarks are needed as we add new features. It's hard to know what effect they will have on performance without having easy to run benchmarks to do measurements.
- We can start some nice performance competition between implementations. And maybe some trash talking too.
- Using the same manifest format that the test runner uses will hopefully allow implementations to resuse code as well as make it easy to test the benchmarks for correctness ahead of time.
- There's already code in jsonld.js that can run benchmarks on manifests. It wasn't too hard to add and has already been useful for some work-in-progress optimizations.
- I wasn't sure where this should go, so I put it alongside other tests. You can reuse the normal test suite data in benchmarks easily (see b002).  Maybe most benchmarks should just be run as part of the main test suite?  I figured the normal test suite would cover what's in the benchmarks so no need to run those again.
- There's a new `loader` option in the manifests to help with mocking a document loader that can pre-load files. Unsure if that's the best way to do that. Work-in-progress.

Thoughts?

If you want to try the jsonld.js version:
```
# correctness test
JSONLD_TESTS=`pwd`/../json-ld-api/tests/benchmarks/b001-manifest.jsonld npm t
# ... 13 passing

# benchmark
JSONLD_TESTS=`pwd`/../json-ld-api/tests/benchmarks/b001-manifest.jsonld JSONLD_BENCHMARK=1 npm t
# ...
compact empty object x 82,458 ops/sec ±1.14% (87 runs sampled)
      ✓ #tb001-c001 Compact empty object (6100ms)
compact empty array x 83,191 ops/sec ±0.69% (89 runs sampled)
      ✓ #tb001-c002 Compact empty array (5996ms)
expand empty object x 206,883 ops/sec ±0.66% (87 runs sampled)
      ✓ #tb001-e001 Expand empty object (5856ms)
# ... etc ...
```


gkellogg:
I wonder if these would be better in the CG. I’m not aware of other working groups that consider performance benchmarking.

davidlehn:
In an un-scribed side discussion at the Feb F2F it was determined that benchmarks, stress tests, and similar should be hosted elsewhere.  They are important for implementations but not necessarily best handled by the WG.  I'll move this PR over to json-ld.org.

PR #60, opened 2019-01-25
Add support for sealed contexts and terms.

gkellogg:
Adds algorithm steps and tests.

For issue w3c/json-ld-syntax#20.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/60.html" title="Last updated on Feb 11, 2019, 10:58 PM UTC (94053f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/60/253b48c...94053f4.html" title="Last updated on Feb 11, 2019, 10:58 PM UTC (94053f4)">Diff</a>


index.html
```diff
@@ -1007,7 +1011,11 @@ <h3>Algorithm</h3>
         <li>
           For each item <var>context</var> in <var>local context</var>:
           <ol>
-            <li>If <var>context</var> is <code>null</code>, set <var>result</var> to a
+            <li>If <var>context</var> is <code>null</code>
+              <span class="changed">and <var>from sealed term</var> is <code>true</code>
+                  or the <var>active context</var> contains no <a>term definition</a>
```
pchampin: This should be "IF context is null  AND ( 'from sealed term' is true OR the active context contains... )"
with those parenhthesis.
The current phrasing does not make the parenthesis explicit, and it might be misleading because the usual precedence of AND and OR implies something else.

PROPOSED: "IF context is null  AND EITHER 'from sealed term' is true OR the active context contains..."

index.html
```diff
@@ -1694,8 +1711,11 @@ <h3>Algorithm</h3>
               has a <a>local context</a>, set <var>active context</var> to the result
               to the result of the
               <a href="#context-processing-algorithm">Context Processing algorithm</a>,
-              passing <var>active context</var> and the value of the
-              <var>term</var>'s <a>local context</a> as <var>local context</var>.</li>
+              passing <var>active context</var>, the value of the
+              <var>term</var>'s <a>local context</a> as <var>local context</var>,
+              <span class="changed">and <var>from sealed term</var> set to the value
```
pchampin: Wow... that would mean that a "sealed type" could redefine the context of the value of a "sealed property", even if the type and the property are sealed by different contexts?

```json
{
  "@context": [
    {
      "@sealed": true,
      "a_prop1": "http://example.org/a#prop1"
    },
    {
      "@sealed": true,
      "b_class1": {
        "@id": "http://example.org/b#class1",
        "@context": null
       }
    }    
  ],
  "a_prop1": {
      "@type": "b_class1",
      "a_prop1": "should this property be unsealed?"
  }
}
```

This is kind of counter-intuitive for me, but it's probably because I'm not familiar yet with "type-scoped contexts"...

gkellogg: Indeed, but it falls out of the logic. But, @dlongley’s notion that sealing doesn’t hold under non sealed terms my alieviate this; perhaps we need to be say that sealed terms can cause their values to be unsealed, and not simply wipe out the context.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-02-01-json-ld#section4).

- No actions or resolutions
Review Pierre-Antoine’s proposal for sealed context processing<br/>
Benjamin Young: <a href="https://lists.w3.org/Archives/Public/public-json-ld-wg/2019Jan/0016.html">https://lists.w3.org/Archives/Public/public-json-ld-wg/2019Jan/0016.html</a><br/>
Ivan Herman: <a href="https://github.com/w3c/json-ld-syntax/pull/119">https://github.com/w3c/json-ld-syntax/pull/119</a><br/>
Gregg Kellogg: <a href="https://github.com/w3c/json-ld-syntax/issues/20">https://github.com/w3c/json-ld-syntax/issues/20</a><br/>
Ivan Herman: -> the proposed text for the syntax draft: <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/119.html#sealed-contexts">https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/pull/119.html#sealed-contexts</a><br/>
Benjamin Young: API PR <a href="https://github.com/w3c/json-ld-api/pull/60">https://github.com/w3c/json-ld-api/pull/60</a><br/>
Gregg Kellogg: I can represent pchampin’s issue, since he can’t call in due to internet connectivity problems<br/>
… There’s an issue, and discussion in the PR<br/>
… the principle issue in #20, and pchamin’s work is around syntax<br/>
… basically, he suggests sealing contexts, not terms, though it’s not design limited that way<br/>
… and sealing a context, the term definitions are tagged with an identifier associated with that context<br/>
… an implementation, for each term def., you know which context it comes from (for sealed contexts)<br/>
… so if you attempt to update that within a sealed node, you can only update it if the update is coming from the sealing context<br/>
… this would happen for embedded context<br/>
… so if I define a term foo, and that context wants to update a term in that context, you can see the terms that the updated terms came from the same context<br/>
… that would allow terms defined in other sealed contexts to not interfere<br/>
… so you cannot create a term that interferes with a different sealed contexts.<br/>
… there is a list of active sealed contexts active during expansion…<br/>
… that effectively unseals all the other contexts when processing<br/>
… if you were to descend into another property not associated, that would have the effect of unsealing all contexts.<br/>
… if you use a term from a context, even if it’s unsealed, you now enter a mode where the context is sealed.<br/>
… what this means for sealing contexts or terms, since you’re tagging the context info per-term, it seems straightforward to do that per-term. There is no support for sealing part of a term, say the ID, but not the container<br/>
… no use cases<br/>
Rob Sanderson: does the seal ID need to be different than the context ID? is it specified explicitly?<br/>
… does this mean that the top context might be sealed, but an embedded one might be unsealed?<br/>
Ivan Herman: he also put a text into the syntax document which describes things and makes it clear and easy to understand<br/>
… there was also an issue: what happens if you do something that is not kosher? one possibility is a warning and you ignore, the other is to throw an error<br/>
… and some of the examples will change based on that answer. It’s a decision<br/>
Benjamin Young: This felt a lot like one context being last<br/>
… it would be super if we could have our use cases written up, both verifiable claims and web of things<br/>
… have use cases<br/>
Dave Longley: I am thinking about this and reviewing it, and looking at it, I’m not sure how it would impact caching of active contexts.<br/>
… I haven’t thought through how it works…can we keep track of an ancestry tree?<br/>
… whatever we put in the spec, whatever matches the output, it’s OK. Still thinking through it<br/>
Ivan Herman: I am worried that what’s described is an implementation strategy, not a requirement for a spec<br/>
Rob Sanderson: +1 to Ivan<br/>
Ivan Herman: we should make sure that we don’t define the process, but instead what we need to have as normative<br/>
Dave Longley: +1 to Ivan<br/>
Ivan Herman: informatively, we can put in examples. What counts is the test cases<br/>
Benjamin Young: we currently don’t track term overriding and who overwrote what in what order<br/>
Dave Longley: +1 … that would be a helpful tool<br/>
Benjamin Young: so if you have two contexts and one overrides it, there’s no awareness of that in the code. So in a context dev, and to keep sealing in, it would be helpful to know what happened and in which order. that could clear a lot of confusion<br/>
Gregg Kellogg: My feeling tis that we’ve been overly prescriptive to ensure interoperability<br/>
… embedded contexts cerate a new ID and inherit from the one embedded<br/>
… so I think we need to try and describe that in a way that stays as close as possible to the behavior, and the mental model has to be described in the syntax doc to understand orders of operations<br/>
… but adding more complexity threatens to degrade performance<br/>
Rob Sanderson: in the proposal, does that include externally referenced, but scoped contexts? So if I had a doc that had sealing at the top level, and I scoped in schema.org within a term, would that then seal schema.org?<br/>
Dave Longley: {"@context": {"sealed": true, "foo": {"@context": "<a href="http://schema.org"}}}?">http://schema.org"}}}?</a><br/>
Rob Sanderson: does that seal by reference schema.org? if you had it as an externally reference, sealed context.<br/>
… what dave said in the chat<br/>
… what happens?<br/>
Gregg Kellogg: my feeling is that scoped context is not explicitly sealed. So, no. schema.org would not be sealed in that example.<br/>
… what you need to know is there is a reference, and we might restrict that to being things in a local/embedded context, have the ability to affect things that were defined in the outer context, but not implicitly sealed<br/>
… does the sealing act valid for everything, or are there parts like vocab that are not sealed?<br/>
Gregg Kellogg: three top-level directives: version, vocab, language, and base<br/>
… context is not a member of a context, it’s a member of a term definition<br/>
Dave Longley: i’ll just say — my view of “sealed contexts” is that someone made a specification for a bunch of terms (a “sealed context”) and JSON-only devs will read that spec and apply the rules therein — not running any JSON-LD processing after that… and we should consider how sealed contexts according to that line of reasoning.<br/>
Gregg Kellogg: the body is an object that contains term definitions, default vocabs…a term definition can itself be sealed, and if that includes an embedded context, and if that’s sealed, then the context that it defined in not implicitly sealed<br/>
… I can’t update the context with a new context…otherwise the use of context within a context is in the document, and….<br/>
… . too many uses of the word “context”.<br/>
Dave Longley: so if we want to allow people to write specs that say: “everything under this term will use schema.org’s context” … that’s what JSON devs would expect, no changes<br/>
Dave Longley: we could of course, require those spec writers to redefine all those terms manually if need be<br/>
Benjamin Young: the top level directives…is it imagined that those would be sealed?<br/>
Gregg Kellogg: we don’t seal base, vocab. if you want to seal, seal a term. if the default vocab is schema.org, and can’t override it…<br/>
Dave Longley: the main use is a spec, like W3c, and they’re defining terms in the spec, and will provide the context, and will only apply the rules in the spec.<br/>
Rob Sanderson: I agree with that too, but then to me it’s going to be clearer if the individual terms are sealed, rather than the terms defined in a context that’s sealed, but other features (e.g. base and vocab) are not sealed in that context<br/>
Dave Longley: first, you’re using JSON-LD to enable extensibility and semantics. that means that you want other terms to be created.<br/>
… so I think that losing .vocab won’t work. Sealed contexts should not be constrain things not defined.<br/>
… we should think of how they should behave within that domain.<br/>
… if we want to figure out how it should work, we should think about it in the context of a spec–if there’s a term not in the spec, you can ignore them.<br/>
… none of these term definitions will change. If you’ followed the path in the spec, nothing will change.<br/>
… that’s what the feature should guarantee<br/>
… everything else is outside of what people who care bout JSON will care about.<br/>
Benjamin Young: +1<br/>
Dave Longley: think about this domain through the lens of JSON-only developers<br/>
Rob Sanderson: Also +1<br/>
Ivan Herman: what would happen if Schema.org sealed that context?<br/>
Dave Longley: all that would happen is that the terms wouldn’t change.<br/>
Ivan Herman: using schema.org, some of the terms are underspecified. So what we do is add our own context that defines, with the same URL, we add additional constraints to the term.<br/>
… this would prevent this<br/>
Dave Longley: if you’re redefining these terms, this would lock that possibility.<br/>
Ivan Herman: even if I use the URL, but I specify the type..<br/>
Dave Longley: that’s a separate issue, but an important one<br/>
Ivan Herman: this happens a lot with schema.org a lot. Often I want to make them more precise<br/>
Benjamin Young: maybe, given large vocabs, sealing many terms, this maybe an API thing, but not a syntax thing? Can the processor seal/not-sealed to create a processor that says that this context is sealed<br/>
… but the developer chooses<br/>
… and then these terms are sacrosanct<br/>
… these are the terms my processor must understand<br/>
Gregg Kellogg: echoing ivan, schema.org author: Values are expected to be an org or person, but that’s defined without a type. JSON-LD processor will allow a string value there<br/>
Dave Longley: i’ll just say that’s interesting (doing this via an API mode/flags) Benjamin … and worth considering further.<br/>
Gregg Kellogg: not a URI.<br/>
Dave Longley: interesting, but i don’t think will work :)<br/>
Gregg Kellogg: that’s a case for updating it for author is explicitly a URI.<br/>
Rob Sanderson: +1 to dlongley, agree it needs to be syntactic<br/>
Dave Longley: a good idea, but there will be people who won’t know to read the spec and flip on the flags. it needs to be in the data.<br/>
Benjamin Young: we’ll see you in a week or less. look through it, add your notes, etc.<br/>
</details>

gkellogg:
I've updated this based on @pchampin's [email](https://lists.w3.org/Archives/Public/public-json-ld-wg/2019Jan/0016.html), and my [response](https://lists.w3.org/Archives/Public/public-json-ld-wg/2019Feb/0001.html).

In the Context Processing Algorithm:

The algorithm now has a `from term` optional parameter, which is either set from the `active property` in the Expansion algorithm, or `null`, depending (as described below).

When parsing a context from an IRI, the IRI is used as the `context id`, otherwise some new unique value. (Note, we don’t actually use `context id`, so it could be anything, but it might be useful for warning reporting.

Let `current seal` be the `context id` associated with `from term` in the active context, if any. An attempt to clear the active context by setting to null is ignored and a warning is issued if `current seal` is set.

Pass `sealed` from the value the `@sealed` member in the local context, if any, `current seal`, and `context id` to the Create Term Definition algorithm.

In the Create Term Definition algorithm:

6) If `current seal` is set, and the active context has an existing term for `term`, and that term is sealed, issue a warning and skip the term definition.

12) Set `sealed` in _definition_ to `context id` if the value of the `@sealed` member in _value_ is true, or the value of _sealed_ if there is no such member.

In the Expansion algorithm:

6) pass `active property` as `from term` when calling the Context Processing Algorithm. — prevents overriding sealed terms from the same context

7.1)  pass `active property` as `from term` when calling the Context Processing Algorithm. — prevents overriding sealed terms from the same context (for @type scoped contexts)

9.5) do not pass `from term` when calling the Context Processing Algorithm. — allows a scoped context for a property to redefine terms

9.8.2.1) pass `active property` as `from term` when calling the Context Processing Algorithm. — prevents overriding sealed terms from the same context (for type maps)

I think this achieves what we want. Expansion is called with the active property, which may be associated with a sealed context, except when processing scoped contexts on property (allows a sealed term to make changes). If the active property is not sealed, the context processing algorithm will allow changes to sealed terms.

index.html
```diff
@@ -992,24 +993,37 @@ <h3>Overview</h3>
...
 
       <p>This algorithm specifies how a new <a>active context</a> is updated
-        with a <a>local context</a>. The algorithm takes three input variables:
-        an <var>active context</var>, a <var>local context</var>, and an  <a>array</a>
-        <var>remote contexts</var> which is used to detect cyclical context inclusions.
-        If <var>remote contexts</var> is not passed, it is initialized to an empty
-        <a>array</a>.</p>
+        with a <a>local context</a>. The algorithm takes two required
+        <span class="changed">and two optional</span>
+        input variables:
```
pchampin: I would replace the colon by a period, and introduce the next sentence by "The requires inputs are".

index.html
```diff
@@ -992,24 +993,37 @@ <h3>Overview</h3>
...
+        an <var>active context</var> and a <var>local context</var>.
+        <span class="changed">The optional inputs are <var>from term</var></span>,
```
pchampin: May be add "which defaults to `null`"

index.html
```diff
@@ -992,24 +993,37 @@ <h3>Overview</h3>
...
-            <li>If <var>context</var> is <code>null</code>, set <var>result</var> to a
+            <li class="changed">Initialize <var>context id</var> to a unique identifier.
+              <span class="note"><var>context id</var> is used to identify sealed terms
+                created from the same context.</span></li>
+            <li>If <var>context</var> is <code>null</code>
+              <span class="changed">and <var>current seal</var> is <code>null</code></span>,
+              set <var>result</var> to a
               newly-initialized <var>active context</var> and continue with the
               next <var>context</var>.
+              <span class="changed">If <var>current seal</var> is <code>null</code>,
```
pchampin: should be "is not `null`"

index.html
```diff
@@ -992,24 +993,37 @@ <h3>Overview</h3>
...
+                processors SHOULD issue a warning that an attempt to clear a sealed context was attempted.</span>
```
pchampin: replace "was attempted" by "was made". "an attempt was attempted" is a bit ugly :)

index.html
gkellogg: The wording was meant to indicate that both of those inputs default to `null`, but being more explicit works.

index.html
gkellogg: Comment made irrelevant by subsequent changes.

index.html
```diff
@@ -992,9 +993,12 @@ <h3>Overview</h3>
...
       <p>This algorithm specifies how a new <a>active context</a> is updated
-        with a <a>local context</a>. The algorithm takes three input variables:
-        an <var>active context</var>, a <var>local context</var>, and an  <a>array</a>
+        with a <a>local context</a>. The algorithm takes three required
+        <span class="changed">and one optional</span>
+        input variables:
+        an <var>active context</var>, a <var>local context</var>, and an <a>array</a>
         <var>remote contexts</var> which is used to detect cyclical context inclusions.
+        <span class="changed">The optional input is <var>from sealed term</var> defaulting to <code>false</code>.</span>
         If <var>remote contexts</var> is not passed, it is initialized to an empty
```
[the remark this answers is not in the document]
gkellogg: Fixed.

gkellogg:
Latest changes should be consistent with our F2F resolutions.

PR #67, opened 2019-03-01
Test `<subject> <rdf:first> <rdf:nil>` triple.

davidlehn:
Adding test to cover a jsonld.js failure: https://github.com/digitalbazaar/jsonld.js/issues/294.
Is this output correct?

### DIFF
--- a/tests/fromRdf-manifest.jsonld
+++ b/tests/fromRdf-manifest.jsonld
@@ -189,6 +189,13 @@
       "input": "fromRdf/0025-in.nq",
       "expect": "fromRdf/0025-out.jsonld"
     }, {
+      "@id": "#t0026",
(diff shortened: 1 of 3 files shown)


gkellogg:
Travis failing, as it looks like a change to common/common.js might not have been picked up. Shouldn't matter for merge.

davidlehn:
@gkellogg So this is the correct output?  The ruby code does the same thing but I didn't dig into the spec algorithm to see if it's what should be output or just what happens to be output.

gkellogg:
Yes, the output looks correct. We improved the algorithm since 1.0 to cover odd cases of list conversion. Step 4.5.8 in the RDF to JSON-LD algorithm sets the trivial case for an empty list, and 5.3.3 should create new lists with the tail appended. If there are no new members, then the list is output with an empty array.

PR #72, opened 2019-03-20
Support JSON Literals.

gkellogg:
Uses JCS normatively for RDF output, with atrisk issue.

For w3c/json-ld-syntax#4.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/72.html" title="Last updated on Mar 26, 2019, 10:39 PM UTC (64a3fe4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/72/42d1159...64a3fe4.html" title="Last updated on Mar 26, 2019, 10:39 PM UTC (64a3fe4)">Diff</a>


gkellogg:
JCS has some failure modes for unrepresentable numbers (e.g., `NaN`, `+-Infinity`). Not sure if it's worth testing these, or calling it out in the algorithm.

iherman:
Minor thingy: after example 22, "unnecssary"->"unnecessary".

index.html
```diff
@@ -2372,7 +2392,10 @@ <h3>Algorithm</h3>
           <a href="#value-compaction">Value Compaction algorithm</a>,
           passing <var>active context</var>, <var>inverse context</var>,
           <var>active property</var>,and <var>element</var> as <var>value</var> is
-          a <a>scalar</a>, return that result.</li>
+          a <a>scalar</a>,
+          <span class="changed">or the <a>term definition</a> for <var>active property</var>
+            has a <a>type mapping</a> of <code>@json</code>,</span>
+          return that result.</li>
```
pchampin: "that result" is ambiguous here. It originally meant "the result of the *Value Compaction Algorithm*, but now the algorithm is not called in all cases.

pchampin:
> Also, I am still not convinced by the role of JSON c14n. I could argue that, e.g., in example 22, the user may expect to see the JSON content _verbatim_ in the generated Turtle, and this canonical version may come as a surprise.

Implementations of the algorithm will usually *not* have access to the verbatim form. In Python for example, the algorithm gets a dict, where order and "unnecssary" spaces have already been tampered with... So even without c14n, the user will usually not get the JSON content verbatim in the generated Turtle.

iherman:
> Implementations of the algorithm will usually not have access to the verbatim form. In Python for example, the algorithm gets a dict, where order and "unnecssary" spaces have already been tampered with... So even without c14n, the user will usually not get the JSON content verbatim in the generated Turtle.

Yeah, that is probably true. I did not think of that.

gkellogg:
> I wonder whether we should not define this in the syntax document first, as well as formally define the RDF JSON datatype. The latter also means we will have to decide whether we put that as a separate document or include it in the syntax, what namespace we use (I would argue we would have to try to put this into the RDF namespace), etc. Those issues may influence this document.

My process is typically to implement the feature, then describe the processing steps and the tests for it. Once we're in agreement with the general direction, then to do the syntax. Certainly, the datatype needs to be normatively described. We can bikeshed on `rdf:` vs `jsonld:` namespaces. Keeping it together with the JSON-LD syntax document seems reasonable to me. If RDF Concepts were a living doc, we could put it there.

> Also, I am still not convinced by the role of JSON c14n. I could argue that, e.g., in example 22, the user may expect to see the JSON content _verbatim_ in the generated Turtle, and this canonical version may come as a surprise.

As @pchampin says, we can't rely on the original syntactic representation, as it's completely lost when transforming into the internal representation (unlike XML/HTML when parsed into an infoset, where whitespace remains). Moreover, once numbers are parsed, any difference between, e.g., `0`, `0.0`, or `0.0e0`, `0.0E+0`, ... is lost. For this, we count on ES6 to describe serializations, which always uses `E+-` with no leading zeros, but doesn't use exponential notation if the value is within certain ranges. Whether or not we can rely on JCS, we'll need to spec something equivalent for basic interoperability, not to mention signing and querying.

iherman:
I must admit I still fail to see why JSON c14n would be _normatively_ necessary. As long as the API produces a _valid_ JSON, whose meaning is identical to the input, this should be enough _from the JSON-LD point of view_. It is of course perfectly fine to refer to the c14n spec in the sense of recommending to implementation to use this if it is important to compare and/or sign the resulting graph or any other form of the JSON-LD content, but that is not the same as casting it into concrete for the API.

In any case, unless the JCS spec is as stable as it can be, we would have a problem referring to it normatively. (We should consider [the relevant guidelines](https://www.w3.org/2013/09/normative-references#stability).) Also, normatively specing a JCS of our own, I believe, should be a no-no. Defining, essentially, a standard that is a generic JSON specification whose role goes way beyond JSON-LD when there is work happening elsewhere would create lots of turmoil. Copying and freezing  in our document an existing standard work that is happening elsewhere is also a (big!) no-no.

I believe we will have to find a way to refer to JCS if we want, but _not_ normatively.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-03-22-json-ld#section1-3).

- `RESOLVED:  Move forwards with a JSON native data type, with a warning that it cannot be canonicalized`
JSON datatype<br/>
Rob Sanderson: link: <a href="https://github.com/w3c/json-ld-syntax/issues/4">https://github.com/w3c/json-ld-syntax/issues/4</a><br/>
Rob Sanderson: PR: <a href="https://github.com/w3c/json-ld-api/pull/72">https://github.com/w3c/json-ld-api/pull/72</a><br/>
Rob Sanderson: we also have discussed the JSON datatype on github<br/>
… Gregg, you’ve been the most involved (as always)<br/>
… could you summarize?<br/>
Gregg Kellogg: the issue comes down to representation<br/>
… if you are going to describe both the lexical and value space<br/>
… somewhat like HTML<br/>
… the lexical space cannon be guaranteed<br/>
… the JSON literal quality is lost when its turned into a native representation<br/>
… you loose the original key ordering, key escaping, and lexical numerical representations<br/>
… so it seems we will need to canonicalize<br/>
… which has been referenced in the issue<br/>
… it’s sadly not as close to done as I’d hoped<br/>
… and we can’t count on it being final in time<br/>
… so, do we care if two implementations use the same canonicalization<br/>
… so we have done some things about do we use Integer or Doubles for numbers<br/>
… so when you’d turn the JSON literal into RDF (in the toRDF space), we do need to say something about that at least<br/>
… and the elimination of whitespace<br/>
… and the ordering of keys<br/>
… I think that can be done<br/>
… there’s a lot of detail in that, but we should be able to reference ECMAScript for this<br/>
… or we could do it ourselves<br/>
Rob Sanderson: last time we talked about the canonicalization issue<br/>
… we also talked about HTML being not easily canonicalizable<br/>
Gregg Kellogg: HTML is a little different<br/>
… they will preserve order, and whitespace<br/>
… so you do have the opportunity return to that result<br/>
Ivan Herman: well, attribute order and things are not covered<br/>
… this would be a problem if you were to attempt to sign an HTML document<br/>
Gregg Kellogg: if we weren’t in an era when signatures weren’t as important as they are now, then maybe we wouldn’t need to care about this so much<br/>
Rob Sanderson: so, is there a JSON-LD document that could include a JSON “native” data type that also needs to be signed<br/>
… so if the only use case is to import GeoJSON<br/>
… do we need to worry<br/>
Ivan Herman: I have spent time on this issue with others<br/>
… aside from the canonicalization problem<br/>
… if we do make a native JSON type, we will have to put it into some namespace–rdf: or jsonld:<br/>
Rob Sanderson: +1 to RDF namespace<br/>
Ivan Herman: if we do that, we’ll have to write the SWIG mailing list, to announce the new datatype, etc.<br/>
… we can do this as part of our document<br/>
… the other problem is<br/>
… I did put a reference in the issue for the rules we have to follow when we point to something normatively<br/>
… my first reading is that unfortunately, this JSON canonicalization specification cannot be referred to normatively<br/>
… the second problem is bringing our own canonicalization into our document<br/>
… if we do that, I can safely say the Director would say no to that<br/>
… so, we can’t just take an IETF spec and put it into a W3C spec<br/>
… all of these are admin problems<br/>
… But I am still not convinced that we need the canonicalization as a normative part of our spec<br/>
… we could say that someone else may do this and reference forthcoming work<br/>
… but when the issue is that we have a JSON portion we want to store in RDF<br/>
… we can state that the only expectation is that [the same processor will produce the same output]<br/>
… none of the arguments that I heard is that canonicalization needs to be normative<br/>
Pierre-Antoine Champin: <a href="http://tinyurl.com/y2gmzxf8">http://tinyurl.com/y2gmzxf8</a><br/>
Pierre-Antoine Champin: I was wondering about this example<br/>
… there’s an Integer in the non-canonical form<br/>
… would that be canonicalized or not?<br/>
Gregg Kellogg: yes, that would be canonicalized<br/>
… I don’t know any processors that would properly serialize that with a leading zero<br/>
… if you’re going to the internal representation<br/>
… it is the number 42<br/>
… some might do 42.0<br/>
… or 42E+0<br/>
… that would be fine, but I don’t think most JSON serializers would do that<br/>
Pierre-Antoine Champin: for the moment, we know how to sign this thing<br/>
Dave Longley: I think this falls into the same category as HTML<br/>
… it’s a string in the JSON; it’s not native HTML<br/>
… or a native number in the example’s case<br/>
… if we’re storing stuff in a string, then store it as a string<br/>
… but people want a native JSON object in their JSON<br/>
Pierre-Antoine Champin: but if you remove the leading 0 you don’t get the same signature<br/>
… so I’m assuming that the signature is dealing with the order or absence of order in the object when signed<br/>
… so if the object was a native JSON object, then it would already benefit<br/>
… and regardless we already have this problem with other string-expressed literals<br/>
Rob Sanderson: if you instead make it value 42.0<br/>
… since no one really serializes as 042<br/>
… whatever you change here will change the signature<br/>
… even though it will canonicalize as something different<br/>
Dave Longley: I disagree<br/>
Rob Sanderson: what do you disagree with?<br/>
Ivan Herman: I think in these examples, the current JSON-LD specification doesn’t say anything about what you put in strings<br/>
… we don’t suggest any sort of mini-canonicalization for things like this<br/>
… having built-in canonicalization for the native JSON representation<br/>
… would be a departure from what we’ve done previously<br/>
Dave Longley: my response to all that is that we have very consistent rules about moving non-string data into strings<br/>
… so we do have those sorts of specifications<br/>
… from a native JSON value into a string<br/>
… this same thing would exist for native JSON objects<br/>
… for things that come in via a string, those will stay as whatever that string is<br/>
… so strings have no issue<br/>
… so if you take pchampin’s example, and change it to a real number: 42<br/>
Gregg Kellogg: 42, 42.0, 42.0E0, 4.2E+1 are all the same number<br/>
Dave Longley: and if you put that in the playground, check the nquads tab, you’ll find the same number<br/>
Ivan Herman: yep I acknowledge that<br/>
Rob Sanderson: maybe then it’s the playground which is at fault<br/>
… I put in several examples, and the signature changes for all of these different 42’s as an integer<br/>
Dave Longley: you’re looking at the RSA signature, so you’ll see it change constantly<br/>
… because that injects random data<br/>
… what you need to look at is the N-Quads or normalized tabs<br/>
… the data there stays the same<br/>
Gregg Kellogg: this is in the data round tripping section<br/>
Gregg Kellogg: so, imo, if we create a datatype for JSON<br/>
… before there is a canonicalization for it<br/>
… then we’re in danger of doing things too early<br/>
… ultimately we need to deal with a canonicalized JSON<br/>
Gregg Kellogg: so the best thing we can do right now is nothing<br/>
… and defer this until there is a canonicalized form<br/>
… otherwise whitespace, object ordering, etc are all variable<br/>
… and the literals really won’t be worth doing any lexical representation is important<br/>
… better not to do anything until a canonicalization spec exists<br/>
Ivan Herman: my take would be milder<br/>
… the GeoJSON example doesn’t care about canonicalization<br/>
Rob Sanderson: +1 to ivan<br/>
Ivan Herman: with the canonicalization things differed<br/>
… and state that this feature is not recommended<br/>
… so we differ it, and if/when the canonicalization becomes standard or whatever, then we at that point suggest that that spec gets used<br/>
Rob Sanderson: it would be better to have a JSON datatype and state that later we’ll do canonicalization<br/>
Dave Longley: let’s provide rules for how to produce the JSON string that match the draft – but that you can do something else and be very clear it’s preferred that everyone do the same thing<br/>
Rob Sanderson: so we should start with JSON datatypes, and just suggest that you can’t sign these<br/>
Jeff Mixter: +1 to ivan and azaroth<br/>
Gregg Kellogg: if we don’t do canonicalization now, we don’t seem to be prevented from doing it later<br/>
… if we end up as a living spec, then we could do it that way<br/>
… and we could also suggest that for testing purposes it is always canonicalized<br/>
Rob Sanderson: a warning or a note?<br/>
Proposed resolution: Move forwards with a JSON native data type, with a warning that it cannot be canonicalized (Rob Sanderson)<br/>
Rob Sanderson: I’d suggest a warning<br/>
Gregg Kellogg: +1<br/>
Jeff Mixter: +1<br/>
Ivan Herman: +1<br/>
Rob Sanderson: +1<br/>
Simon Steyskal: +1<br/>
Tim Cole: +1<br/>
Dave Longley: +0<br/>
Benjamin Young: +0 still have concerns about eager misuse<br/>
David I. Lehn: +0.5<br/>
Jeff Mixter: I echo bigbluehat concerns but I also have very valid reasons to add JSON to RDF data.<br/>
Dave Longley: +1 to everything Benjamin is saying … but that we should really also have JSON literals … but they should also all be converted to the same strings in processors :)<br/>
David Newbury: +1<br/>
Resolution #3: Move forwards with a JSON native data type, with a warning that it cannot be canonicalized<br/>
Dave Longley: JSON literals can be an escape hatch but ONLY an escape hatch.<br/>
</details>

gkellogg:
I updated the spec to only mention JCS, but described a canonical lexical form for the purposes of round-tripping. It's close to JCS, but uses our established conventions for serializing numbers, rather than what ES6 specifies.

@iherman suggested not specifying anything, as I understood from the call, but I don't see how we can remain entirely silent on this. Please suggest an edit to soften the language, if advisable.

dlongley:
@davidlehn,

> It's close to JCS, but uses our established conventions for serializing numbers, rather than what ES6 specifies.

Can you look into how this would impact our JavaScript implementation? It sounds problematic. I think it's one thing to serialize numbers on their own when converting to `xsd:integer`, etc. but I doubt that, when serializing an entire JSON literal (that happens to contain numbers), we should deviate from ES6.

gkellogg:
You can look at the latest changes to the test files to see how differently numbers are treated. In general, I think the ES6/JCS serialization is more developer friendly, but it’s odd to have a difference.

The other main difference is that keys are sorted as UTF-16, not -8, which is a subtle difference.

I’m happy to go however people like, and if we say serialze numbers as in ES6, that’s fine with me.

dlongley:
I'd prefer to go with ES6/JCS ... otherwise we're likely to do something different from everyone else.

PR #74, opened 2019-03-23
compaction and expansion of indexes using a property

gkellogg:
Add API text and tests for compaction and expansion of indexes using a property.

For w3c/json-ld-syntax#145.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/74.html" title="Last updated on Mar 25, 2019, 5:09 PM UTC (70712fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/74/c8c199f...70712fc.html" title="Last updated on Mar 25, 2019, 5:09 PM UTC (70712fc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -926,6 +926,7 @@
       <span class="changed">an optional <a>context</a></span>,
       <span class="changed">an optional <dfn>nest value</dfn>,</span>
       <span class="changed">an optional <dfn>prefix flag</dfn>,</span>
+      <span class="changed">an optional <dfn>index mapping</dfn>,</span>
       <span class="changed">a <dfn>protected</dfn>, used to mark this as a protected term,</span>
       and an optional <dfn>container mapping</dfn>.
       A <a>term definition</a> can not only be used to map a <a>term</a>
@@ -1371,6 +1372,20 @@
               has been detected and processing is aborted.</li>
             <li>Set the <a>container mapping</a> of <var>definition</var> to
               <var>container</var>.</li>
+          </ol>
+        </li>
+        <li class="changed">If <var>value</var> contains the <a>member</a> <code>@index</code>:
+          <ol>
+            <li>If <a>processing mode</a> is <code>json-ld-1.0</code> or
+              <a>container mapping</a> does not include <code>@index</code>,
+              an <a data-link-for="JsonLdErrorCode">invalid term definition</a>
+              has been detected and processing is aborted.</li>
+            <li>Initialize <var>index</var> to the value associated with the
+              <code>@index</code> <a>member</a>, which MUST a <a>string</a> expanding to an <a>absolute IRI</a>.
+              Otherwise, an
+              <a data-link-for="JsonLdErrorCode">invalid term definition</a>
+              has been detected and processing is aborted.</li>
+            <li>Set the <a>index mapping</a> of <var>definition</var> to <var>index</var></li>
           </ol>
         </li>
         <li class="changed">If <var>value</var> contains the <a>member</a> <code>@context</code>:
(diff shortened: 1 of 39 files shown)


index.html
```diff
@@ -1974,36 +1991,41 @@ <h3>Algorithm</h3>
...
+                          ensuring that the value is represented using an <a>array</a>.</li>
+                        <li class="changed">If <var>container mapping</var> includes <code>@index</code>,
+                          <var>index key</var> is not <code>@index</code>,
+                          <var>item</var> does not have a <a>member</a> <code>@index</code>
+                          and <var>expanded index</var> is not <code>@none</code>,
+                          set <var>index property values</var> to the concatenation of
+                          <var>expanded index</var> with any existing values of
+                          <var>index key</var> in <var>item</var>.
+                          Add the key-value pair (<var>expanded index</var>-<var>index property values</var>) to <var>item</var>.
+                          If expanded value is a value object,
```
pchampin: shouldn't 'expanded value' be `<var>item</var>`?

index.html
```diff
@@ -2757,17 +2779,31 @@ <h3>Algorithm</h3>
...
                       Set <var>map key</var> to the value of <code>@language</code> in <var>expanded item</var>, if any.</li>
-                    <li>If <var>container</var> includes <code>@index</code> set <var>map key</var> to the value of <code>@index</code> in <var>expanded item</var>, if any,
+                    <li>Otherwise, if <var>container</var> includes <code>@index</code>
+                      and <var>index key</var> is <code>@index</code>,
+                      set <var>map key</var> to the value of <code>@index</code> in <var>expanded item</var>, if any,
                       and remove <var>container key</var> from <var>compacted item</var>.</li>
-                    <li class="changed">If <var>container</var> includes <code>@id</code>, set
+                    <li class="changed">Otherwise, if <var>container</var> includes <code>@index</code>
+                      and <var>index key</var> is not <code>@index</code>,
+                      Set <var>map key</var> to the first value of <var>container key</var> in <var>compacted item</var>, if any.
```
pchampin: shouldn't it be `<var>index key</var>` instead of `<var>container key</var>`?

index.html
```diff
@@ -2757,17 +2779,31 @@ <h3>Algorithm</h3>
...
+                      If there are remaining values in <var>compacted item</var>
```
pchampin: This sentence is really not clear for me. It seems to me that `compacted container` should be replaced by `container key` and that `compacted value` should be replaced by `compacted item`.
Indeed:
* `compacted item` is never defined, AFAICT
* `compacted value` is used in other parts of the algorithm, but does not seem to have a value in this branch

The same sentence appears later (line 2808), and should be fixed the same way, if I'm correct.

gkellogg:
> Appart from those I think this PR is good. It is consistent with [w3c/json-ld-syntax#145](https://github.com/w3c/json-ld-syntax/pull/145) . However, I seem to remember you suggesting a change, where instead of
> 
> ```json
> { "@container": "@index", "@index": "nickname" }
> ```
> 
> one would simply write
> 
> ```json
> { "@container": "nickname" }
> ```
> 
> Are you still willing to make that change (in which case the syntax and api PRs should be updated)?

It's an open question, which we need to resolve quickly. On the one hand, piggy-backing on `@index` may reduce the cognitive load on users, on the other is separating concepts which have somewhat different constraints. I'm 50/50 on making such a change. Maybe the best thing is to resolve as is, and raise a new issue to separate, but I'd rather not drag this out.

index.html
gkellogg: No, this is a bit tricky, as the index case is handled differently from other types of indexing earlier in the compaction algorithm. This is one of the things that make index maps different from other maps. If we were to re-write the compaction algorithm from scratch, we might do it differently.

index.html
gkellogg: __compacted item__ is defined in 9.7.5. (You can select the variable in HTML to find all the places it's used in the algorithm, which should start with its definition.)

It should be using __compacted item__ and __compacted key__ consistently. I'll fix.

pchampin: Sorry, I meant "`compacted container` is never defined"

Issue #76, opened 2019-04-08
More compact @prefix


rubensworks:
Currently, `@prefix` has boolean as range, and can be used as follows:

```json
{
  "@context": {
    "@version": 1.1,
    "compact-iris": { "@id": "http://example.com/compact-iris-", "@prefix": true }
  },
}
```

In most cases, `@prefix` will probably be used together with `@id`.
For this reason, it might be beneficial to **also** allow the following as equivalent representation:

```json
{
  "@context": {
    "@version": 1.1,
    "compact-iris": { "@prefix": "http://example.com/compact-iris-" }
  },
}
```

As far as I can see, this should not cause conflicts or major performance issues when processing.

BigBlueHat:
See PR #77 for more.

dlongley:
PROPOSAL: If `@prefix` is present in a term definition, and its value is a string, then the term may only be used as a prefix. Additionally, including `@id` in the same term definition is an error.

Note that this adds additional complexity to processing as we would now have a new class of term that has no `@id`. I'm not sure how complex it would be.

To be clear, this would be different from treating `@prefix: <string>` as just syntactic sugar for `@id: <string>, @prefix: true` as I believe was originally proposed in this issue. Instead, this syntax would be a way to define a prefix that cannot be used on its own as a term.

dlongley:
@pchampin -- Does the above sound like what we meant on the call?

rubensworks:
> PROPOSAL: If @prefix is present in a term definition, and its value is a string, then the term may only be used as a prefix. Additionally, including @id in the same term definition is an error.

So this would mean that if `@id` and `@prefix: true` would be used, that the term can be used as both a prefix and a property, right?

This would only be useful for some exotic use cases, but indeed important for backwards-compatibility with 1.0 behaviour. (We may want to clarify that this kind of usage is discouraged)

dlongley:
@rubensworks,

> So this would mean that if @id and @prefix: true would be used, that the term can be used as both a prefix and a property, right?

Yes.

> This would only be useful for some exotic use cases, but indeed important for backwards-compatibility with 1.0 behaviour. (We may want to clarify that this kind of usage is discouraged)

And yes. +1

pchampin:
@dlongley yes, that's how I understand the outcome of our discussion.
and +1 as well to @rubensworks comment above.

iherman:
[...]

PR #77, opened 2019-04-08
toRdf tests for @prefix

rubensworks:
This PR does two things:

1. It modifies toRdf-0088 so that the compact IRI term definition that ends with a non-gen-delim character is removed, as this is not allowed anymore unless `@prefix` is used. This is needed, as this is a breaking change in 1.0:
    > This represents a small change to the 1.0 algorithm to prevent IRIs that are not really intended to be used as prefixes from being used for creating compact IRIs.

2. toRdf-0130 is added as an extension of toRdf-0088, and makes use of this new `@prefix` feature.

### DIFF
--- a/tests/toRdf-manifest.jsonld
+++ b/tests/toRdf-manifest.jsonld
@@ -880,6 +880,14 @@
       "input": "toRdf/0129-in.jsonld",
       "expect": "toRdf/0129-out.nq"
     }, {
+      "@id": "#t0130",
+      "@type": ["jld:PositiveEvaluationTest", "jld:ToRDFTest"],
+      "name": "Compact IRIs and prefix",
+      "purpose": "1.1 extension of toRdf-0088",
+      "input": "toRdf/0130-in.jsonld",
+      "expect": "toRdf/0130-out.nq",
+      "option": {"specVersion": "json-ld-1.1", "processingMode": "json-ld-1.1"}
+    }, {
       "@id": "#th001",
       "@type": ["jld:PositiveEvaluationTest", "jld:ToRDFTest"],
       "name": "Transforms embedded JSON-LD script element",
--- a/tests/toRdf/0088-in.jsonld
+++ b/tests/toRdf/0088-in.jsonld
@@ -1,15 +1,14 @@
 {
   "@context": {
     "term": "http://example.com/terms-are-not-considered-in-id",
-    "compact-iris": "http://example.com/compact-iris-",
     "property": "http://example.com/property",
     "@vocab": "http://example.org/vocab-is-not-considered-for-id"
   },
   "@id": "term",
   "property": [
     {
-      "@id": "compact-iris:are-considered",
-      "property": "@id supports the following values: relative, absolute, and compact IRIs"
+      "@id": "http://example.com/absolute-iri",
+      "property": "@id supports the following values: relative and absolute IRIs"
     },
     {
       "@id": "../parent-node",
--- a/tests/toRdf/0088-in.jsonld
+++ b/tests/toRdf/0088-in.jsonld
@@ -1,14 +1,20 @@
 {
   "@context": {
     "term": "http://example.com/terms-are-not-considered-in-id",
+    "unsafe-compact-iris": "http://example.com/unsafe-compact-iris-",
+    "safe-compact-iris": "http://example.com/safe-compact-iris/",
     "property": "http://example.com/property",
     "@vocab": "http://example.org/vocab-is-not-considered-for-id"
   },
   "@id": "term",
   "property": [
     {
-      "@id": "http://example.com/absolute-iri",
-      "property": "@id supports the following values: relative and absolute IRIs"
+      "@id": "unsafe-compact-iris:are-considered",
+      "property": "@id supports the following values: relative, absolute, and compact IRIs ending with gen-delim character"
+    },
+    {
+      "@id": "safe-compact-iris:are-considered",
+      "property": "@id supports the following values: relative, absolute, and compact IRIs ending with gen-delim character"
     },
     {
       "@id": "../parent-node",
--- a/tests/toRdf/0088-out.nq
+++ b/tests/toRdf/0088-out.nq
@@ -1,4 +1,3 @@
 <http://example.com/absolute-iri> <http://example.com/property> "@id supports the following values: relative and absolute IRIs" .
 <https://w3c.github.io/json-ld-api/tests/parent-node> <http://example.com/property> "relative IRIs get resolved against the document's base IRI" .
-<https://w3c.github.io/json-ld-api/tests/toRdf/term> <http://example.com/property> <http://example.com/compact-iris-are-considered> .
 <https://w3c.github.io/json-ld-api/tests/toRdf/term> <http://example.com/property> <https://w3c.github.io/json-ld-api/tests/parent-node> .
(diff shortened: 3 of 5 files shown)


rubensworks:
I did not add any tests for compact IRIs with term definition not ending with a gen-delim for two reasons:

1. I am not sure if these cases should result in an error or ignoring the term (I assume the latter).
2. If an error should be thrown, no `NegativeEvaluationTest`s exist in the toRdf suite, so I suspect there may be a reason for this (is there?).

gkellogg:
Tests are failing because some files from json-ld-wg/common need to be copied over.

I’ll look at the changes in more detail when I get home in a day or so.

It is intended that Compact IRIs using a term not ending in a gen-delim are treated as if there were no term definition, so it will likely be treated as an absolute IRI. Generating an error might be heavy handed, a warning would be useful. We could consider ignoring the Compact IRI as well, like an undefined term.

rubensworks:
> It is intended that Compact IRIs using a term not ending in a gen-delim are treated as if there were no term definition, so it will likely be treated as an absolute IRI. Generating an error might be heavy handed, a warning would be useful. We could consider ignoring the Compact IRI as well, like an undefined term.

Updated the tests to reflect this.

gkellogg:
We'll need some spec-text to ignore such Compact IRIs. IRI Expansion step 5.4 will need some discussion to qualify the term for use, considering gen-delim and `@prefix` definitions. But, this might not be an acceptable change considering backwards compatibility. So, we may not be able to pull this PR.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-05-03-json-ld#section3-2).

- `RESOLVED:  dlongley and pchampin to craft alternate proposal to #76 with @prefix having 3 values`
More compact @prefix<br/>
Benjamin Young: link: <a href="https://github.com/w3c/json-ld-api/issues/76">https://github.com/w3c/json-ld-api/issues/76</a><br/>
Gregg Kellogg: related: <a href="https://github.com/w3c/json-ld-api/pull/77">https://github.com/w3c/json-ld-api/pull/77</a><br/>
Benjamin Young: proposal from rubenworks, to make it simpler to declare @prefix terms<br/>
Ruben Taelman: a similar thing is already possible with @reverse, so it should not be too hard to add in implementations<br/>
Dave Longley: is this an error?<br/>
… what do we do when @id and @prefix are both strings and they disagree? is that an error?<br/>
… my understanding is that the proposal is that "@prefix": `` is just syntactic magic for `"@id": ``, "@prefix": true` … not a replacement<br/>
Gregg Kellogg: it is with @reverse, so it should be an error, yes<br/>
… there is a point in the issue about terms which are not appropriate as prefixes<br/>
… there are two aspects in the PR: 1/ use @prefix where the value is an IRI, while the current practice is to use @id with an IRI, and @prefix is a boolean. Should we obsolete the current practice?<br/>
… 2/ what if the value of @prefix is a pname?<br/>
Ruben Taelman: [answers to gkellogg]<br/>
Gregg Kellogg: we must keep in mind the JSON 1.0 behavior<br/>
Ruben Taelman: I see this as an alternate representation<br/>
Ivan Herman: does the combination of @id and @prefix:false come up in practice?<br/>
Gregg Kellogg: there are reasons to use it. I don’t know if it is used in the wild?<br/>
Ivan Herman: that would be a reason to keep the current practice, and consider the new version as a shorthand<br/>
… and I wouldn’t be shocked if the new notation @prefix:IRI was roundtripped to @id+@prefix:true<br/>
Gregg Kellogg: we don’t roundtrip contexts anyway<br/>
Benjamin Young: there seems to be a consensus on considering the proposal as a shorthand for the current syntax<br/>
Benjamin Young: <a href="https://w3c.github.io/json-ld-syntax/#compact-iris">https://w3c.github.io/json-ld-syntax/#compact-iris</a> (for current text)<br/>
Gregg Kellogg: there are some implications; must be some working on the implications of combinations of @id and @prefix<br/>
… you probably would not use the prefix term as a property<br/>
Dave Longley: perhaps if @prefix is a string, then also including @id is an error … and then process @prefix as a string first and do the transformation to @id: `, @prefix: true`<br/>
Benjamin Young: <a href="https://w3c.github.io/json-ld-syntax/#example-32-using-explicit-prefix-declaration-to-create-compact-iris">https://w3c.github.io/json-ld-syntax/#example-32-using-explicit-prefix-declaration-to-create-compact-iris</a><br/>
Benjamin Young: @id+@prefix:true feels a lot like using @vocab<br/>
Gregg Kellogg: it works a little differently in the compactIRI algo<br/>
… we are going contrary the trend of moving compact IRIs away<br/>
… the reason it is there is that, in 1.0, many terms were used to build compact IRIs, but were not intended to do that<br/>
… eg: Sport := schema:Sport, then using Sport:Event<br/>
… @prefix was a way of capturing the intention of the creator of the term<br/>
Dave Longley: lots of unwanted CURIEs being greedily generated in 1.0<br/>
Benjamin Young: strange compact IRIs can be used maliciously, to misdirect users to an unexpected endpoint URL<br/>
Benjamin Young: <a href="https://github.com/w3c/json-ld-syntax/issues/155">https://github.com/w3c/json-ld-syntax/issues/155</a><br/>
Dave Longley: I like the syntax, makes it more obvious that the term is intended to be used as a prefix<br/>
… even though I don’t like CURIEs in general<br/>
… in JSON-LD<br/>
… It was a mistake in 1.0 to allow any term to be used as a prefix<br/>
Benjamin Young: curie spec <a href="https://www.w3.org/TR/curie/">https://www.w3.org/TR/curie/</a><br/>
Benjamin Young: I agree with that<br/>
Pierre-Antoine Champin: Proposal is not to make @prefix:false the default?<br/>
Dave Longley: No<br/>
Gregg Kellogg: we did change 1.0 behavior by limiting the terms that can be used as prefixes<br/>
… only IRIs ending with gen-delims can be used as prefixes<br/>
Dave Longley: instead of just “property”<br/>
Benjamin Young: gendelim <a href="https://tools.ietf.org/html/rfc3986#section-2.2">https://tools.ietf.org/html/rfc3986#section-2.2</a><br/>
Ivan Herman: what does @prefix:false means?<br/>
Ivan Herman: for the records: <a href="https://tinyurl.com/yxw7qv98">https://tinyurl.com/yxw7qv98</a> (example)<br/>
Gregg Kellogg: it means that we do not want this term to be used by compaction to generate a compact IRI<br/>
Gregg Kellogg: <a href="https://w3c.github.io/json-ld-syntax/#example-32-using-explicit-prefix-declaration-to-create-compact-iris">https://w3c.github.io/json-ld-syntax/#example-32-using-explicit-prefix-declaration-to-create-compact-iris</a><br/>
Gregg Kellogg: for expansion, it does not matter whether @prefix is true or false<br/>
Dave Longley: if you click on compact in the link above<br/>
… by copying the context to the right and changing the value of @prefix<br/>
… you can see the difference<br/>
… but you can’t make a term only a prefix, this is a 1.0 problem, we can’t go back<br/>
Gregg Kellogg: but we could define @prefix:IRI to mean exactly that<br/>
… but we still mean @prefix:false<br/>
Dave Longley: yes, but that would differ from what we said earlier. This is not syntactic sugar anymore<br/>
Pierre-Antoine Champin: another way would be add 3 values for @prefix<br/>
Rob Sanderson: +1 to PA<br/>
Pierre-Antoine Champin: @default would say you can use this prefix as a term<br/>
… false as you can not use this as prefix<br/>
… true, the proposed new meaning that gregg proposed about @prefix IRI<br/>
… this way we can keep backwards compatibility with 1.0<br/>
Proposed resolution: dlongley and pchampin to craft alternate proposal to #76 with @prefix having 3 values (Benjamin Young)<br/>
Gregg Kellogg: +1<br/>
Adam Soroka: +1<br/>
Benjamin Young: +1<br/>
Ruben Taelman: +1<br/>
Ivan Herman: +1<br/>
Dave Longley: +1 sure<br/>
David I. Lehn: +1<br/>
Dave Longley: the third value would be an IRI, and we’re back to what gkellogg just proposed<br/>
Rob Sanderson: +1<br/>
Resolution #2: dlongley and pchampin to craft alternate proposal to #76 with @prefix having 3 values<br/>
</details>

gkellogg:
We've deferred this feature of `@prefix` to a future version.

PR #78, opened 2019-04-13
Protected term errors

gkellogg:
Replaces #69.

- Changes simple naming text as needed.
- Adds a number of tests.
- Redefinition attempts will fail on protection violations.


gkellogg:
Pending any comments to the contrary, I'll plan to merge this tomorrow.

index.html
```diff
@@ -1211,9 +1226,11 @@ <h3>Algorithm</h3>
...
           <a data-link-for="JsonLdErrorCode">keyword redefinition</a>
           error has been detected and processing is aborted.</li>
-        <li class="changed">If the <var>active context</var> has an existing
-          <a>term definition</a> for <var>term</var> which is protected, processing is aborted;
-          processors SHOULD issue a warning that an attempt was made to redefine a protected term.</li>
+        <li class="changed">If the <var>from term</var> is <code>false</code>
+          and <var>active context</var> has an existing
+          <a>term definition</a> for <var>term</var> which is protected,
+          a <a data-link-for="JsonLdErrorCode">protected term redefinition</a> error has been detected,
+          and processing is aborted.</li>
```
dlongley: The one thing we might want to modify here is to allow for "redefinitions" that cause the *exact same term definition* to occur. There's really no reason disallow that from an authorship perspective -- and people might be surprised that they can't combine two contexts that share a common base context with protected terms. On the other hand, this would complicate implementation.

@davidlehn -- since you were in the code most recently, would this be a tough thing to support?

dlongley: Either way, I don't think that this should hold up the PR.

gkellogg: We could say that an attempt to redefine a term definition that is exactly equivalent to the existing definition is allowed, needing to define what we mean by equivalent, but it seems an unnecessary complication to me.

dlongley:
@pchampin,

It's passed as `true` here:

https://github.com/w3c/json-ld-api/pull/78/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR1402

dlongley:
@pchampin, Oh you said in the compaction algorithm ... I'll double check with @davidlehn to make sure we didn't need something there. I suspect we might want another test.

gkellogg:
You’re right, it’s unused in the Compaction algorithm. The test is used when expanding, but perhaps there’s an unconsidered case when compacting with what would have been an illegal context. This could use some tests. I’ll take another look.

gkellogg:
> Thanks, @gkellogg! Sorry we didn't get to adding these today, we were swamped with other work. We'll check them against our JS implementation as soon as we can. Just one note -- you added 4 tests (which all look correct to me) but only 3 made it into the manifest. Otherwise, this looks great.

Edit problem, added in again.

dlongley:
Thanks again, @gkellogg. Our JS implementation passes all these tests now. I think the create term definition algorithm needs a tweak though (hard to tell w/o preview mode) -- wherever it says to call the algorithm recursively it needs to pass the options (so if we're processing a property term scoped context that flag gets propagated during recursion).

gkellogg:
> Thanks again, @gkellogg. Our JS implementation passes all these tests now. I think the create term definition algorithm needs a tweak though (hard to tell w/o preview mode) -- wherever it says to call the algorithm recursively it needs to pass the options (so if we're processing a property term scoped context that flag gets propagated during recursion).

It doesn’t seem to be necessary to propagate the flame recursively from Create Term Definition. Do you see a case where this affects processing?

dlongley:
@gkellogg,

> It doesn’t seem to be necessary to propagate the flame recursively from Create Term Definition. Do you see a case where this affects processing?

Hmm ... I think you're right. Looking at the algorithm text it we only recurse when there's a prefix that needs to be defined and under those conditions there are no potential `@protected` interactions.

Ok, I think we're good to go here.

PR #80, opened 2019-04-18
Use "dcterms" in favor of "dc".

gkellogg:
For w3c/json-ld-syntax#159.

### DIFF
--- a/common/extract-examples.rb
+++ b/common/extract-examples.rb
@@ -24,6 +24,10 @@ end
 
 PREFIXES = {
   dc:     "http://purl.org/dc/terms/",
+  dct:    "http://purl.org/dc/terms/",
+  dcterms:"http://purl.org/dc/terms/",
+  dc11:   "http://purl.org/dc/elements/1.1/",
+  dce:    "http://purl.org/dc/elements/1.1/",
   cred:   "https://w3id.org/credentials#",
   ex:     "http://example.org/",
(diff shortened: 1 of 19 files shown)


common/extract-examples.rb
```diff
@@ -24,6 +24,10 @@
 
 PREFIXES = {
   dc:     "http://purl.org/dc/terms/",
```
iherman: Shouldn't 'dc' be changed?

tests/compact/0007-context.jsonld
```diff
@@ -1,6 +1,6 @@
 {
   "@context": {
-    "dc": "http://purl.org/dc/elements/1.1/",
+    "dcterms": "http://purl.org/dc/elements/1.1/",
```
iherman: That does not sound right: the url for dcterms shouldn't be `elements`

iherman: (the same comment applies to all the files below...)

gkellogg: I changed those in framing, but missed them here. Tests were using the prefix inconsistently.

common/extract-examples.rb
gkellogg: as mentioned on API, this is there because of backwards compatibility, as previously "dc" was used for terms, although we've changed them all to "dcterms" now.

PR #82, opened 2019-04-24
Add checks for terms including a colon

gkellogg:
which do not expand in a natural way.

For w3c/json-ld-syntax#155

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/82.html" title="Last updated on Apr 24, 2019, 5:45 PM UTC (cffd00e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/82/ef43954...cffd00e.html" title="Last updated on Apr 24, 2019, 5:45 PM UTC (cffd00e)">Diff</a>

### DIFF
--- a/tests/expand-manifest.jsonld
+++ b/tests/expand-manifest.jsonld
@@ -187,7 +187,8 @@
       "name": "Term definition with @id: @type",
       "purpose": "Expanding term mapping to @type uses @type syntax",
       "input": "expand/0026-in.jsonld",
-      "expect": "expand/0026-out.jsonld"
+      "expect": "expand/0026-out.jsonld",
+      "option": {"specVersion": "json-ld-1.0"}
     }, {
       "@id": "#t0027",
       "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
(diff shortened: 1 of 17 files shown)


tests/expand-manifest.jsonld
```diff
@@ -1225,6 +1227,22 @@
       "purpose": "Verifies that an exception is raised on expansion when processing an invalid context attempting to define @container on a keyword",
       "input": "expand/e042-in.jsonld",
       "expect": "keyword redefinition"
+    }, {
+      "@id": "#te043",
+      "@type": ["jld:NegativeEvaluationTest", "jld:ExpandTest"],
+      "name": "Term definition with @id: @type",
+      "purpose": "Expanding term mapping to @type uses @type syntax now illegal",
+      "input": "expand/e043-in.jsonld",
```
dlongley: The label `#te043` here doesn't match the input file name prefix `e043` (same for `e044`) -- which seems to break the naming convention, so I suspect this is a typo. Just checking here, was this intentional or a typo?

gkellogg: Tests all of an identifier such as `#txxxx` with corresponding files `expand/xxxx-in.jsonld` and `expand/xxx-out.jsonld`. In this case the `#te043` is consistent with the file named `e043-in` and follows the same pattern as that for `#te042`.

davidlehn: I think the convention has been the test ids have a `t` prefix, the file names don't.  Is that what you meant?

gkellogg: Yes, the ids all start with a "t", the filename's don't. This convention was established in the 1.0 days. I established the practice if replacing one or two leading zeros with letters to help separate out the new tests by broad feature categorization, which helps in filtering as well.

The `#texxx` tests typically are for errors, some of which previously came from a separate error manifest.

Issue #66, opened 2019-02-25
What if a @context URL response is HTML?


BigBlueHat:
See #53.

The HTML-based response has the value of potentially providing documentation for the contained context, but we lack a few things and/or need to clarify others.

See @gkellogg's suggestions at https://github.com/w3c/json-ld-api/issues/53#issuecomment-445401552

azaroth42:
Then it's an error.

> We dereference the remote context and replace context with the value of the `@context` member of the top-level object in the retrieved JSON-LD document. If there's no such member, an invalid remote context has been detected.

From https://www.w3.org/TR/json-ld11-api/#overview

gkellogg:
It’s currently an error, but we can update to handle this case.

azaroth42:
Only if there's a previous context that establishes a 1.1 processing mode, no?

gkellogg:
Maybe, there’s not really a danger of a 1.00 processor misinterpreting this, which is the purpose of the `@version` announcement, see we might allow this as 1.1 behavior. It’s really only when there’s a worry that different versions will process the same document differently that it matters.

gkellogg:
This requires some syntax changes as well, and we should allow frames to be in HTML too.

BigBlueHat:
While I like the opportunities this could open up, I remain a bit concerned that the JSON-LD API itself now requires HTML processing vs. having a separate process for extracting JSON-LD and then processing it.

Practically, it means that shipping fully conformant JSON-LD API code requires far more dependencies than it used to. I'd hoped that we could modularize it a bit more such that these things could be layered together when needed.

Would it make sense to move this HTML stuff to an "extraction" process, which is then fed to expansion/compaction/framing, etc?

gkellogg:
Other than for context processing, it is handled in the WebIDL, with a section on extracting JSON-LD from HTML. We could probably move more algorithmic bits into a separate section and remove some steps from the context area, but there are some subtle differences that might make that more complex.

Once we decided to deal with JSON-LD in HTML normatively, they dye was cast on the need to include HTML tools in a JSON-LD processor.

iherman:
> Once we decided to deal with JSON-LD in HTML normatively, they dye was cast on the need to include HTML tools in a JSON-LD processor.

Indeed.

BigBlueHat:
> Once we decided to deal with JSON-LD in HTML normatively, they dye was cast on the need to include HTML tools in a JSON-LD processor.

[...]

PR #83, opened 2019-04-25
Issue 66 html context

gkellogg:
Allow contexts to be HTML documents, with preference towards script elements of type `application/ld+json;profile=http://www.w3.org/ns/json-ld#context`.

For #66.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/83.html" title="Last updated on May 1, 2019, 1:27 PM UTC (ccc0ccf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/83/b0ba962...ccc0ccf.html" title="Last updated on May 1, 2019, 1:27 PM UTC (ccc0ccf)">Diff</a>


gkellogg:
Note that I didn't add compact tests, as the compact tests take the extracted context, and not a URL. All the important paths are checked here.

We'll want to add text to allow frame documents to be HTML too, for the same reasons of documentation.

rubensworks:
Is there a particular reason to only take the first context that was found in the HTML?
If I understand the PR correctly,
a processor  would only extract the context from the first script tag in the following:

```html
<html>
<head>
  <script id="context" type="application/ld+json;profile=http://www.w3.org/ns/json-ld#context">
    {
      "@context": {
        "term3": "http://example.com/term3",
        "term4": "http://example.com/term4",
        "term5": "http://example.com/term5"
      }
    }
  </script>
  <script id="context" type="application/ld+json;profile=http://www.w3.org/ns/json-ld#context">
    {
      "@context": {
        "term4": "http://example.com/term4",
        "term5": "http://example.com/term5"
      }
    }
  </script>
</head>
</html>
```

In situations where Web designers are generating their views from multiple templates,
and different contexts may be contained in different templates,
it may make sense to allow multiple contexts in HTML.

Semantically, this could be processed similarly as context arrays.

gkellogg:
It is certainly something to consider, but if we fall back to using the profile-less type, I don’t see us taking them all, just the first, so it might seem inconsistent. People should 👍 or 👎 your suggestion.

gkellogg:
I tried to implement this, and ran into problems. One problem is that several tests fail, which previously passed, as normative text requires that a remote context be a top-level object, not array.

From [Context Processing Algorithm step 3.2.4](https://w3c.github.io/json-ld-api/#algorithm):

> If the dereferenced document has no top-level dictionary with an `@context` member, an **invalid remote context** has been detected and processing is aborted; otherwise, set _context_ to the value of that member.

I don't think it's worth softening or confusing this text without a compelling use case.

dlongley:
I can see this approach being fraught with problems, mostly around dynamically generated script tags. Many (most?) people don't build static sites anymore -- which is why the search engines have adapted to run JS to do things like read JSON-LD on the page. Are we going to require the same of JSON-LD processors?

If not, we're relying on a pretty specific scenario under which this will work: A website that serves a static page or at least one where the first script tag with an `@context` is static. Is the assumption that the people using this feature will do this? Is this a safe/fair assumption?

It seems to me like this might be a feature that's too targeted for the Web of yesterday ... or one that requires much more than the (already heavy) HTML parsing lift to function in a way users expect.

gkellogg:
The stated purpose of using HTML files for context (and frame) is to be able to document those files, in which case the JSON-LD script elements can be statically generated (perhaps the HTML is dynamically generated from the JSON-LD).

Otherwise, if the JSON-LD is dynamically generated and inserted into a script tag, you may as well content-negotiate to generate the pure JSON-LD file for a direct response.

We certainly can't anticipate every new web technology that may come about, but we can recommend practices for working for JSON-LD in web pages, particularly when the documentatative aspect is primary.

gkellogg:
@BigBlueHat @dlongley @iherman I added a section on processor levels and added text around parts of the algorithm which manipulate HTML to use them. (Also added similar text to w3c/json-ld-syntax#167). See if this satisfies your concerns. If so, I can split out tests into a separate context.

rubensworks:
> Many (most?) people don't build static sites anymore

Static websites are still very much a thing. Many highly popular tools exist for this (Jekyll (used by GitHub pages), Gastby, Mkdocs, ...), so we should definitely consider this an active domain.

> A website that serves a static page or at least one where the first script tag with an @context is static. Is the assumption that the people using this feature will do this? Is this a safe/fair assumption?

In any case, I don't have a strong preference to allow multiple context, but I could imagine that people writing webpages expect this to be possible. For example, CSS allows multiple stylesheets to be referenced, where each next one builds upon the previous one. So logically, a developer may assume that JSON-LD (contexts) work in the same way.

That being said, defining JSON-LD contexts in HTML is a very specific use-case, so I don't think multiple contexts in HTML are a must-have feature, it just would be nice-to-have for developers.

iherman:
> @BigBlueHat @dlongley @iherman I added a section on processor levels and added text around parts of the algorithm which manipulate HTML to use them. (Also added similar text to w3c/json-ld-syntax#167). See if this satisfies your concerns. If so, I can split out tests into a separate context.

I am basically fine with this. The minor thing that bothers me is that the definitions for pure processor & Co. are in a non-normative sections, whereas the reference to these terms are all in normative sections. Maybe these terms should become normative?

index.html
```diff
@@ -5067,6 +5123,7 @@ <h3>The <dfn>JsonLdProcessor</dfn> Interface</h3>
             <code>application/ld+json</code>,
             nor any other media type using a
             <code>+json</code> suffix as defined in [[RFC6839]],
+            <span class="changed">or if the processor is a <a>pure JSON Processor</a></span>,
```
pchampin: The whole sentence is a bit complex to parse... But it seems to me that this last part ("or if the processor...") wrongly implies that a pure JSON Processor will reject any media-type.
More precisely, I read the whole sentence as IF (media-type neither X,Y...) OR (pure JSON processor) THEN reject.

I suggest rewriting it as:
* If content-type is text/html and the processor is a pure JSON Processor, reject.
* Otherwise, if the content-type is neither application/ld+json..., reject.

pchampin:
+1 to @iherman about making "pure JSON processor" and "full processor" normative. However, "event-based JSON processor" should probably not be, because: 
* we decided during the last call *not* to fully specifiy it,
* the current text states that only pure JSON processors may reject HTML documents, so it implicitly requires event-based processor to be able to parse HTML. Not all streaming implementations might want this constraint.

pchampin:
Thinking more about it, I think that mixing the "pure JSON vs. full" distinction with "event-based" is confusing... especially if two out of three are to be defined normatively. "Event based" is not a "level", it is somewhat orthogonal to the others.

I propose to 
* make the "processor levels" section normative (can we mark as normative a subsection of the non-normative Section 1?), possibly by moving it in Section 3: Conformance;
* move the description of "event-based provessor" somewhere else – although I don't have a clear idea where...

gkellogg:
I can certainly make the section normative, which will require moving it someplace else. Whether or not the term “event-based JSON Processor” is, itself normative, it’s not referenced from any other normative statements, but the text can elaborate that the term is non-normative, or I can add a non-normative subsection to describe it.

index.html
gkellogg: Yes, I was being lazy in not trying to rewrite the text more, but it was indeed, incorrect.

Issue #85, opened 2019-05-02
Update documentLoader to return parsed document


gkellogg:
The documentation for [`documentLoader`](https://w3c.github.io/json-ld-api/#dom-jsonldoptions-documentloader) documentation is a bit lax, but it seems to provide for retrieving processed documents.

In the API, the `documentLoader` option is of type [`DocumentLoaderCallback`](https://w3c.github.io/json-ld-api/#dom-loaddocumentcallback), which is a `Promise<USVString> (USVString url);`. Adding an optional `options` parameter can allow to provide more information, such as for use in processing contexts, or when `extractAllScripts` is specified. But, it indicates that the promise returns a `USVString`.

We also define a [`RemoteDocument`](https://w3c.github.io/json-ld-api/#dom-remotedocument) type, which is really what the `documentLoader` promise should result in. `RemoteDocument` has a [`document`](https://w3c.github.io/json-ld-api/#dom-remotedocument-document) attribute of `any`, and is stated to either return the raw payload or parsed document. It would be useful for us to limit this to the parsed document, which would allow us to encapsulate more of the HTML processing within this definition (see https://github.com/w3c/json-ld-syntax/pull/167#issuecomment-488710786). But, this might be considered breaking the 1.0 API contract.

The Context Processing Algorithm doesn't explicitly describe using the `documentLoader` promise to load remote contexts, although it is implicit in the description of the `documentLoader`. This is somewhat complicated by the attempt to keep all WebIDL information, such as promises, out of the algorithms themselves, which remains a challenging aspect.

With suitable wording, I propose that we move all of the HTML-specific processing rules into the definition of [`RemoteDocument`](https://w3c.github.io/json-ld-api/#dom-remotedocument) and suitably parameterize calls to `documentLoader` (including within the Context Processing Algorithm) to handle the various cases, and result in the internal JSON representation. We can handle the 1.0 contract, which allows an external information to simply return the raw document, by wrapping the callback in code which detects this, and provides it's own processing of the results.

gkellogg:
PR #87 reviewed by @dlongley. Closing.

PR #87, opened 2019-05-03
Rewrite LoadDocumentCallback

gkellogg:
to include most HTML-specific and profile-specific processing, and use in *expand* and *Context Processing*.

For #85.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/87.html" title="Last updated on May 6, 2019, 9:20 PM UTC (27f650f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/87/cbf6f4f...27f650f.html" title="Last updated on May 6, 2019, 9:20 PM UTC (27f650f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5613,7 +5613,7 @@
               <p>If no such element is found,
                 or the located element is not a <a>JSON-LD script element</a>,
                 the <var>promise</var> is rejected with a <a>JsonLdError</a> whose code is set to <a data-link-for="JsonLdErrorCode">loading document failed</a>
-                and processing is terminated.</p></li>
+                and processing is terminated.</p>
             </li>
             <li>Otherwise, if the <a data-link-for="LoadDocumentOptions">profile</a>
               option is specified,
@@ -5661,7 +5661,6 @@
           nor any other media type using a
           <code>+json</code> suffix as defined in [[RFC6839]],
           reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
-        </li>
         <li>Create a new <a>RemoteDocument</a> <var>remote document</var> using <var>document</var>,
           the returned <a>Content-Type</a> (without parameters) as <a data-link-for="RemoteDocument">contentType</a>,
           any returned <code>profile</code> parameter,
--- a/index.html
+++ b/index.html
@@ -5110,8 +5110,8 @@
           <li>If the passed <a data-lt="jsonldprocessor-expand-input">input</a>
             is a <a>string</a> representing the <a>IRI</a> of a remote document, await and dereference it as <var>remote document</var>
             using <a>LoadDocumentCallback</a>, passing <a data-lt="jsonldprocessor-expand-input">input</a>
-            for <a data-link-for="LoadDocumentCallback">url</a>,
-            the <a data-link-for="JsonldOptions">extractAllScripts</a> option for <a data-link-for="LoadDocumentOptions">extractAllScripts</a>.
+            for <a data-link-for="LoadDocumentCallback">url</a>
+            and the <a data-link-for="JsonldOptions">extractAllScripts</a> option for <a data-link-for="LoadDocumentOptions">extractAllScripts</a>.
           </li>
           <li class="changed">If necessary, transform <a data-link-for="RemoteDocument">document</a>
             from <var>remote document</var> into the <a>internal representation</a>.
--- a/index.html
+++ b/index.html
@@ -5078,7 +5078,7 @@
             and if passed, the <a data-link-for="JsonldOptions">compactArrays</a>
             <span class="changed">and <a data-link-for="JsonldOptions">ordered</a></span>
             flags in <a data-lt="jsonldprocessor-compact-options">options</a>.</li>
-          <li>Fulfill the <var>promise</var> passing <var>compacted output</var>
+          <li>Resolve the <var>promise</var> with <var>compacted output</var>
             <span class="changed">transforming <var>compacted output</var> from the
               <a>internal representation</a> to a JSON serialization</span>.</li>
         </ol>
@@ -5110,8 +5110,8 @@
           <li>If the passed <a data-lt="jsonldprocessor-expand-input">input</a>
             is a <a>string</a> representing the <a>IRI</a> of a remote document, await and dereference it as <var>remote document</var>
             using <a>LoadDocumentCallback</a>, passing <a data-lt="jsonldprocessor-expand-input">input</a>
-            for <a data-link-for="LoadDocumentCallback">url</a>
-            and the <a data-link-for="JsonldOptions">extractAllScripts</a> option for <a data-link-for="LoadDocumentOptions">extractAllScripts</a>.
+            for <a data-link-for="LoadDocumentCallback">url</a>,
+            the <a data-link-for="JsonldOptions">extractAllScripts</a> option for <a data-link-for="LoadDocumentOptions">extractAllScripts</a>.
           </li>
           <li class="changed">If necessary, transform <a data-link-for="RemoteDocument">document</a>
             from <var>remote document</var> into the <a>internal representation</a>.
@@ -5137,7 +5137,7 @@
             and if passed, the <span class="changed"><a data-link-for="JsonldOptions">frameExpansion</a></span>
             <span class="changed">and <a data-link-for="JsonldOptions">ordered</a></span>
             flags in <a data-lt="jsonldprocessor-compact-options">options</a>.</li>
-          <li>Fulfill the <var>promise</var> passing <var>expanded output</var>
+          <li>Resolve the <var>promise</var> with <var>expanded output</var>
             <span class="changed">transforming <var>expanded output</var> from the
               <a>internal representation</a> to a JSON serialization</span>.</li>
         </ol>
@@ -5183,7 +5183,7 @@
             and if passed, the <a data-link-for="JsonldOptions">compactArrays</a>
             <span class="changed">and <a data-link-for="JsonldOptions">ordered</a></span>
             flags in <a data-lt="jsonldprocessor-compact-options">options</a>.</li>
-          <li>Fulfill the <var>promise</var> passing <var>flattened output</var>
+          <li>Resolve the <var>promise</var> with <var>flattened output</var>
             <span class="changed">transforming <var>flattened output</var> from the
               <a>internal representation</a> to a JSON serialization</span>.</li>
         </ol>
@@ -5265,7 +5265,7 @@
               as is the support for <a>generalized RDF Datasets</a>
               and thus the <a data-link-for="JsonLdOptions">produceGeneralizedRdf</a> option may be also be removed.</div>
             </li>
-          <li>Fulfill the <var>promise</var> passing <var>dataset</var>.</li>
+          <li>Resolve the <var>promise</var> with <var>dataset</var>.</li>
         </ol>
 
         <dl class="parameters">
@@ -5538,8 +5538,8 @@
 
       <p>The <dfn>LoadDocumentCallback</dfn> defines a callback that custom document loaders
         have to implement to be used to retrieve remote documents and contexts.
-        Upon successful completion, the callback returns a <a>Promise</a> containing a <a>RemoteDocument</a>.
-        On failure, the Promise is rejected with an appropriate error <a data-link-for="JsonLdError">code</a>.</p>
+        The callback returns a <a>Promise</a> resolving to a <a>RemoteDocument</a>.
+        On failure, the <a>Promise</a> is rejected with an appropriate error <a data-link-for="JsonLdError">code</a>.</p>
 
       <pre class="idl" data-transform="unComment"><!--
         callback LoadDocumentCallback = Promise&lt;RemoteDocument> (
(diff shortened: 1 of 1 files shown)


gkellogg:
Per our meeting, I've updated the document loader section to consolidate HTML and profile processing. I'll give this until EOD Monday before pulling and releasing a new WD, pending feedback. As discussed today, we won't close this issue pending further review, but don't want to hold back a WD.

index.html
```diff
@@ -5135,88 +5112,32 @@ <h3>The <dfn>JsonLdProcessor</dfn> Interface</h3>
...
-                <code>application/json</code>,
-                <code>application/ld+json</code>,
-                nor any other media type using a
-                <code>+json</code> suffix as defined in [[RFC6839]],
-                reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
-            </ol>
+            is a <a>string</a> representing the <a>IRI</a> of a remote document, await and dereference it as <var>remote document</var>
+            using <a>LoadDocumentCallback</a>, passing <a data-lt="jsonldprocessor-expand-input">input</a>
+            for <a data-link-for="LoadDocumentCallback">url</a>,
+            the <a data-link-for="JsonldOptions">extractAllScripts</a> option for <a data-link-for="LoadDocumentOptions">extractAllScripts</a>.
```
```suggestion
            and the <a data-link-for="JsonldOptions">extractAllScripts</a> option for <a data-link-for="LoadDocumentOptions">extractAllScripts</a>.
```

index.html
```diff
@@ -5135,88 +5112,32 @@ <h3>The <dfn>JsonLdProcessor</dfn> Interface</h3>
...
-              <li>Otherwise, if the retrieved document's <a>Content-Type</a> is neither
-                <code>application/json</code>,
-                <code>application/ld+json</code>,
-                nor any other media type using a
-                <code>+json</code> suffix as defined in [[RFC6839]],
-                reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
-            </ol>
+            is a <a>string</a> representing the <a>IRI</a> of a remote document, await and dereference it as <var>remote document</var>
+            using <a>LoadDocumentCallback</a>, passing <a data-lt="jsonldprocessor-expand-input">input</a>
+            for <a data-link-for="LoadDocumentCallback">url</a>,
```
```suggestion
            for <a data-link-for="LoadDocumentCallback">url</a>
```

gkellogg:
> Organizationally, this is a massive improvement, IMO -- a much better separation of concerns. Thanks, @gkellogg! I will only be able to offer more detailed feedback (if any) when we attempt implementation. There do seem to be more profile requirements than before (from the last time I reviewed) and I don't know what impact there will be on existing systems (hopefully none!).

The profile requirement is that systems that don't process HTML, which includes all 1.0 processors, as they will reject the mime type, should raise the same error they would have under 1.0. 1.1 _full processors_ MUST handle the HTML content, and include `text/html` in the Accept header (or equivalent). We don't seem to have every said in 1.0 that the Accept header should include `application/ld+json` or `application/json` but it seems implicit.

The net is that to be a _full processor_ you'll need to support HTML and a callback implementation must conform to the specified behavior. We do have tests to make sure that it handles the requests properly, but not for processor type, nor do we presently have a way to test the compliance of custom loaders, but I'm not sure how we would.

There is some cleanup of the test needed to separate out the HTML behavior into a different manifest.

index.html
```diff
@@ -5131,92 +5108,36 @@ <h3>The <dfn>JsonLdProcessor</dfn> Interface</h3>
...
+            the <a data-link-for="JsonldOptions">extractAllScripts</a> option for <a data-link-for="LoadDocumentOptions">extractAllScripts</a>.
           </li>
+          <li class="changed">If necessary, transform <a data-link-for="RemoteDocument">document</a>
```
pchampin: The meaning of "if necessary" is not entirely clear to me... I'm guessing this means "if 'document' is a string rather than a dictionary or array"?

Note by the way that, according to the definition, a string is a valid 'internal representation' (indeed, a JSON document may encode a single string). So this "if necessary" carries a lot of implicit...

index.html
```diff
@@ -5612,25 +5533,174 @@ <h3>Remote Document and Context Retrieval</h3>
...
+        <li>If the retrieved resource's <a>Content-Type</a> is <code>application/json</code>
+          or any media type with a <code>+json</code> suffix as defined in [[RFC6839]]
+          except <code>application/ld+json</code>,
+          and the response has an HTTP Link Header [[RFC8288]] using the <code>http://www.w3.org/ns/json-ld#context</code> link relation,
+          set <var>contextUrl</var> to the associated <code>href</code>.
+          <p>If multiple HTTP Link Headers using the <code>http://www.w3.org/ns/json-ld#context</code> link relation are found,
+            the <var>promise</var> is rejected with a <a>JsonLdError</a> whose code is set to <a data-link-for="JsonLdErrorCode">multiple context link headers</a>
+            and processing is terminated.</p>
+          <p>Processors MAY transform <var>document</var> to the <a>internal representation</a>
+            using the <a href="#extract-script-content">Extract Script Content algorithm</a>.</p>
```
pchampin: This is a little confusing, as the 'Extract Script Content Algorithm' is defined to handle the content of HTML script elements, which is not what we have here... Furthermore, that algorithm may raise an 'invalid script element' error, which would not be appropriate here.

I suggest rephrasing this in the spirit of line 5116 (with the clarification that I have suggested in my previous comment).

index.html
```diff
@@ -5612,25 +5533,174 @@ <h3>Remote Document and Context Retrieval</h3>
...
+              of <code>application/ld+json</code> along with the value of the
+              <a data-link-for="LoadDocumentOptions">profile</a> option, if found.</li>
+            <li>If <var>source</var> is undefined and the <a data-link-for="LoadDocumentOptions">extractAllScripts</a> option is not present, or <code>false</code>,
+              set <var>source</var> to the <a data-cite="DOM#dom-node-textcontent">textContent</a>
+              of the first <a>JSON-LD script element</a> in <var>document</var>.
+              <p>If no such element is found,
+                or the located element is not a <a>JSON-LD script element</a>,
+                the <var>promise</var> is rejected with a <a>JsonLdError</a> whose code is set to <a data-link-for="JsonLdErrorCode">loading document failed</a>
+                and processing is terminated.</p></li>
+            <li>If <var>source</var> is defined and the,
```
pchampin: something must be missing between "and the" and the trailing comma...

index.html
```diff
@@ -5612,25 +5533,174 @@ <h3>Remote Document and Context Retrieval</h3>
...
+              set <var>document</var> to the result of the
+              <a href="#extract-script-content">Extract Script Content algorithm</a>,
+              using <var>source</var>, rejecting <var>promise</var>
+              with a <a>JsonLdError</a> whose code set from the result, if an error is detected
+              and processing is terminated.
+            </li>
+            <li>Otherwise, if <var>source</var> is undefined and the <a data-link-for="LoadDocumentOptions">extractAllScripts</a> option is not present, or <code>false</code>,
+              the <var>promise</var> is rejected with a <a>JsonLdError</a> whose code is set to <a data-link-for="JsonLdErrorCode">loading document failed</a>
+              and processing is terminated.</li>              
+            <li>Otherwise, if <var>source</var> is undefined and the <a data-link-for="LoadDocumentOptions">extractAllScripts</a> option is <code>true</code>,
```
pchampin: If I understand correctly, this "otherwise" is the final "else". The condition ("source is undefined and the extractAllScripts option is true") is necessarily true.

To makes this clearer, I suggest rephrasing: "Otherwise, source is undefined and the extractAllScripts option is true. Set document to..."

index.html
```diff
@@ -5612,25 +5533,174 @@ <h3>Remote Document and Context Retrieval</h3>
...
+                  using <var>source</var>, rejecting <var>promise</var>
+                  with a <a>JsonLdError</a> whose code set from the result, if an error is detected
+                  and processing is terminated.</li>
+                <li>If <var>script content</var> is an <a>array</a>, merge it to the end of <var>document</var>.</li>
+                <li>Otherwise, append <var>script content</var> to <var>document</var>.</li>
+              </ol>
+            </li>
+          </ol>
+        </li>
+        <li>Otherwise, if the retrieved document's <a>Content-Type</a> is neither
```
pchampin: Same remark as above, this is an "else", not an "else if".
Suggested change (1/3)
```suggestion
        <li>Otherwise, the retrieved document's <a>Content-Type</a> is neither
          <code>text/html</code>,
```

index.html
```diff
@@ -6014,6 +6084,8 @@ <h2>Changes since JSON-LD Community Group Final Report</h2>
       of type <code>application/ld+json;profile=http://www.w3.org/ns/json-ld#context</code>,
       or <code>application/ld+json</code> is used as the context for further processing.
       This allows a mechanism for documenting the content of a context using HTML.</li>
+    <li>Consolodate <a>RemoteDocument</a> processing into the <a>LoadDocumentCallback</a>
```
```suggestion
    <li>Consolidate <a>RemoteDocument</a> processing into the <a>LoadDocumentCallback</a>
```

index.html
pchampin: "If necessary" is not very clear to me. I'm guessing it means "if 'document' is a neither a dictionary not an array". This should be made explicit, all the more than a plain string is listed as a valid "internal representation" (which makes sense as the JSON syntax is able to represent a single string).

index.html
pchampin: This is confusing: this algorithm is explicitly provided for processing HTML scripts, which is not what we have here. Furthermore, it may raise an 'invalid script element' error, which would be inappropriate here.

I suggest rephrasing this in the spirit of line 5116 (modulo the improvement that I suggested).

index.html
pchampin: If I understand correctly, this "otherwise" is an "else", but it reads like an "else if", which might confuse the reader.
I suggest rephtrasing it (1/2)
```suggestion
            <li>Otherwise,<var>source</var> is undefined and the <a data-link-for="LoadDocumentOptions">extractAllScripts</a> option is <code>true</code>.
```

index.html
```diff
@@ -5612,25 +5533,174 @@ <h3>Remote Document and Context Retrieval</h3>
...
+                <li>Otherwise, append <var>script content</var> to <var>document</var>.</li>
+              </ol>
+            </li>
+          </ol>
+        </li>
+        <li>Otherwise, if the retrieved document's <a>Content-Type</a> is neither
+          <code>application/json</code>,
+          <code>application/ld+json</code>,
+          nor any other media type using a
+          <code>+json</code> suffix as defined in [[RFC6839]],
```
pchampin: (cont 2/3)
```suggestion
          <code>+json</code> suffix as defined in [[RFC6839]].
```

index.html
```diff
@@ -5612,25 +5533,174 @@ <h3>Remote Document and Context Retrieval</h3>
...
+          reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
```
pchampin: (cont. 3/3)
```suggestion
          Reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
```

index.html
```diff
@@ -5612,25 +5533,174 @@ <h3>Remote Document and Context Retrieval</h3>
...
+              <a href="#extract-script-content">Extract Script Content algorithm</a>,
+              using <var>source</var>, rejecting <var>promise</var>
+              with a <a>JsonLdError</a> whose code set from the result, if an error is detected
+              and processing is terminated.
+            </li>
+            <li>Otherwise, if <var>source</var> is undefined and the <a data-link-for="LoadDocumentOptions">extractAllScripts</a> option is not present, or <code>false</code>,
+              the <var>promise</var> is rejected with a <a>JsonLdError</a> whose code is set to <a data-link-for="JsonLdErrorCode">loading document failed</a>
+              and processing is terminated.</li>              
+            <li>Otherwise, if <var>source</var> is undefined and the <a data-link-for="LoadDocumentOptions">extractAllScripts</a> option is <code>true</code>,
+              set <var>document</var> to a new empty <a>array</a>.
```
pchampin: (cont 2/2)
```suggestion
              Set <var>document</var> to a new empty <a>array</a>.
```

index.html
pchampin: Something is missing between "and the" and the trailing comma.

index.html
gkellogg: I'll change, but even though a string is a valid JSON document, it's not a valid JSON-LD document, so the intention should be clear.

index.html
gkellogg: I'll just remove the bit about the Extract Script Content, as we don't really need to show how it should be transformed.

gkellogg: At this point, it's just the HTTP body, so it will always be in a string form. The contract allows an user-supplied implementation to do the transformation.

index.html
gkellogg: This should simply remove the "and the".

index.html
gkellogg: I added some deeper lists to clarify.

gkellogg:
Thanks @pchampin, the algorithm had an odd branch logic because of some fall-through behavior. With some extra structure based on your suggestions, I think it reads more clearly.

PR #89, opened 2019-05-08
Update spec and add tests for limiting `@type`-scoped terms to their typed object.

dlongley:
Related to https://github.com/w3c/json-ld-syntax/issues/174.

I've created this PR for jsonld.js with the necessary algorithm changes.

https://github.com/digitalbazaar/jsonld.js/pull/312

What do you think @gkellogg?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/89.html" title="Last updated on May 26, 2019, 11:25 PM UTC (c8a7e0a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/89/c71d460...c8a7e0a.html" title="Last updated on May 26, 2019, 11:25 PM UTC (c8a7e0a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -974,14 +974,13 @@
     <p>The <a>active context</a> contains the active
       <a>term definitions</a> which specify how
       keys and values have to be interpreted as well as the current <a>base IRI</a>,
-      the <a>vocabulary mapping</a> and the <a>default language</a>. Each
-      <a>term definition</a> consists of:</p>
+      the <a>vocabulary mapping</a>, the <a>default language</a>,
+      <span class="changed">and an optional <dfn>previous context</dfn>,
+        used when a type-scoped <a>context</a> is defined</span>.
+      Each <a>term definition</a> consists of:</p>
     <ul>
       <li>an <dfn>IRI mapping</dfn>,</li>
       <li>a boolean flag <dfn>reverse property</dfn>,</li>
-      <li class="changed">a boolean flag <dfn>from type</dfn>,</li>
-      <li class="changed">a <dfn>previous definition</dfn>
-        used to record any previous <a>term definition</a>,</li>
       <li>an optional <dfn>type mapping</dfn>, or <dfn>language mapping</dfn>,</li>
       <li class="changed">an optional <a>context</a>,</li>
       <li class="changed">an optional <dfn>nest value</dfn>,</li>
@@ -1079,6 +1078,10 @@
         <li>If <var>local context</var> is not an <a>array</a>,
           set it to an <a>array</a> containing only
           <var>local context</var>.</li>
+        <li class="changed">
+          If <var>from type</var> is <code>true</code>, and <var>result</var>
+          does not have a <a>previous context</a>, set <a>previous context</a>
+          in <var>result</var> to <var>active context</var>.</li>
         <li>
           For each item <var>context</var> in <var>local context</var>:
           <ol>
@@ -1089,8 +1092,10 @@
                   an <a data-link-for="JsonLdErrorCode">invalid context nullification</a>
                   has been detected and processing is aborted.</li>
                 <li>Otherwise, set <var>result</var> to a
-                  newly-initialized <var>active context</var> and continue with the
-                  next <var>context</var>.
+                  newly-initialized <var>active context</var>,
+                  <span class="changed">setting <a>previous context</a> in <var>result</var>
+                    to the previous value of <var>result</var> if <var>from type</var> is <code>true</code></span>.
+                  Continue with the next <var>context</var>.
                   <span class="note">In JSON-LD 1.0, the <a>base IRI</a> was given
                     a default value here; this is now described conditionally
                     in <a href="#the-application-programming-interface" class="sectionRef"></a>.</span></li>
@@ -1275,8 +1280,7 @@
           <var>protected</var> which defaults to <code>false</code>,
           <var>from property</var>, defaulting to <code>false</code>,
           which is used to allow changes to protected terms,
-          and <var>from type</var>, defaulting to <code>false</code>,
-          which is used to mark <a>term definitions</a> associated with a type-scoped <a>context</a>.</span>.
+          and <var>from type</var>, defaulting to <code>false</code>.</span>.
         </p>
       <ol>
         <li>If <var>defined</var> contains the <a>member</a> <var>term</var> and the associated
@@ -1314,8 +1318,6 @@
           <code>@id</code>-<code>null</code>, set the
           <a>term definition</a> in <var>active context</var> to
           <code>null</code>,
-          <span class="changed">along with <a>from type</a> from <var>from type</var>
-            and <a>previous definition</a> from <var>previous definition</var>,</span>
           and set the value associated with <var>defined</var>'s
           <a>member</a> <var>term</var> to <code>true</code>, and return.</li>
         <li>Otherwise, if <var>value</var> is a <a>string</a>, convert it
@@ -1326,9 +1328,7 @@
           <a data-link-for="JsonLdErrorCode">invalid term definition</a>
           error has been detected and processing is aborted.
           <span class="changed">Set <var>simple term</var> to <code>false</code></span>.</li>
-        <li>Create a new <a>term definition</a>, <var>definition</var>,
-          including <a>from type</a> from <var>from type</var>
-            and <a>previous definition</a> from <var>previous definition</var>.</li>
+        <li>Create a new <a>term definition</a>, <var>definition</var>.</li>
         <li class="changed">If the <code>@protected</code> member in <var>value</var>
           is <code>true</code>, or there is no <code>@protected</code> member in <var>value</var>
           and the <var>protected</var> parameter is <code>true</code>,
@@ -1559,34 +1559,6 @@
     </section>
   </section> <!-- end of Term Creation -->
 
-  <section class="changed algorithm"><h2>Revert Type Scoped Terms</h2>
-    <p>This algorithm returns a new <a>context</a> based on the <a>active context</a>
-      where any <a>term definitions</a> which were created as part of a type-scoped <a>context</a>
-      are replaced with any previous term definitions, or removed.</p>
-
-    <p>The algorithm takes a single required parameter <var>active context</var></p>
-    <ol>
-      <li>If <var>active context</var> contains no <a>term definitions</a>
-        where <var>from type</var> is <code>true</code>, return <var>active context</var>.</li>
-      <li>Initialize <var>result</var> to the result of cloning <var>active context</var>.</li>
-      <li>For each <a>member</a> <a>term definition</a> with key <var>term</var> and value <var>definition</var> in <var>result</var>:
-        <ol>
-          <li>If <var>definition</var> has <var>from type</var> with a value of <code>true</code>:
-            <ol>
-              <li>Set <var>previous definition</var> to any <a>previous definition</a>
-                in <var>definition</var>.</li>
-              <li>If <var>previous definition</var> is undefined,
-                remove <var>term</var> and <var>definition</var> from <var>result</var>.</li>
-              <li>Otherwise, update the <a>member</a> value for <var>term</var> in <var>result</var>
-                with <var>previous definition</var>.</li>
-            </ol>
-          </li>
-        </ol>
-      </li>
-      <li>Return <var>result</var>.</li>
-    </ol>
-  </section>
-
   <section>
     <h2>IRI Expansion</h2>
 
@@ -1808,10 +1780,16 @@
         <li>If <var>element</var> is <code>null</code>, return <code>null</code>.</li>
         <li class="changed">If <var>active property</var> is <code>@default</code>,
           set the <a data-link-for="JsonLdOptions">frameExpansion</a> flag to <code>false</code>.</li>
+        <li class="changed">If <var>active property</var> has a <a>term definition</a> in <var>active context</var>
+          has a <a>local context</a>, set <var>property scoped context</var> to the that <a>local context</a>.</li>
         <li>If <var>element</var> is a <a>scalar</a>,
           <ol>
             <li>If <var>active property</var> is <code>null</code> or <code>@graph</code>,
               drop the free-floating <a>scalar</a> by returning <code>null</code>.</li>
+            <li class="changed">If <var>property scoped context</var> is defined,
+              set <var>active context</var> to the result of the
+              <a href="#context-processing-algorithm">Context Processing algorithm</a>,
+              passing <var>active context</var> and <var>property scoped context</var> as <var>local context</var>.</li>
             <li>Return the result of the
               <a href="#value-expansion">Value Expansion algorithm</a>, passing the
               <var>active context</var>, <var>active property</var>, and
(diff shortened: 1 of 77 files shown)


gkellogg:
@dlongley if you don't mind, I'll change the title on this issue and add algorithm steps to handle it, so that it is complete.

As you note, we'll need to update syntax a bit to allow for this.

dlongley:
> if you don't mind, I'll change the title on this issue and add algorithm steps to handle it, so that it is complete.

Sure and thanks, @gkellogg!

gkellogg:
@dlongley see if you like the spec text associated with the feature.

index.html
```diff
@@ -975,16 +975,23 @@ <h2>Context Processing Algorithm</h2>
...
+      <li>a boolean flag <dfn>reverse property</dfn>,</li>
+      <li class="changed">a boolean flag <dfn>from type</dfn>,</li>
+      <li class="changed">a <dfn>previous definition</dfn>
+        used to record any previous <a>term definition</a>,</li>
+      <li>an optional <dfn>type mapping</dfn>, or <dfn>language mapping</dfn>,</li>
+      <li class="changed">an optional <a>context</a>,</li>
+      <li class="changed">an optional <dfn>nest value</dfn>,</li>
+      <li class="changed">an optional <dfn>prefix flag</dfn>,</li>
+      <li class="changed">an optional <dfn>index mapping</dfn>,</li>
+      <li class="changed">a <dfn>protected</dfn>, used to mark this as a protected term,</li>
```
dlongley: We might want to call `protected` a `protected flag` here instead.

index.html
```diff
@@ -2068,15 +2121,21 @@ <h3>Algorithm</h3>
...
-                      to <var>term context</var>.</li>
+                    <li class="changed">If <var>container mapping</var> includes <code>@type</code>:
+                      <ol>
+                        <li>Set <var>map context</var> to the result of calling the
+                          <a href="#revert-type-scoped-terms">Revert Type Scoped Terms</a> algorithm
+                          passing <var>term context</var> as <var>active context</var>.</li>
+                        <li>If <var>index</var>'s <a>term definition</a> in
+                          <var>map context</var> has a <a>local context</a>, update
+                          <var>map context</var> to the result of the
+                          <a href="#context-processing-algorithm">Context Processingalgorithm</a>,
```
```suggestion
                          <a href="#context-processing-algorithm">Context Processing algorithm</a>,
```

dlongley:
@gkellogg,

On the last call, we discussed what might happen if someone nullified the context via a type-scoped context. This resulted in me changing my implementation to handle this case. I think the implementation is actually simpler:

https://github.com/digitalbazaar/jsonld.js/pull/316/commits/1c00ec3ec829f79201cc4ae053c2887fab4d462d

In short, whenever a type-scoped context is being processed, we save the previous active context. Then, when we call `revertTypeScopedContext` (instead of `revertTypeScopedTerms`), which restores the context to whatever it was prior to processing the type-scoped context. This will also make sure to revert any changes to `@base`, `@vocab`, etc. that were made just for the typed object.

To handle property-scoped contexts that come up within a type-scoped context, we process them against the stored previous context (updating it), rather than against the current active context, as they occur "within" and apply "according to" that previous context, not the type-scoped one.

At first, I was thinking that we would want the property-scoped contexts to be processed against the type-scoped context, but in practice, this is already confusing. For example, if you set `@protected` or `@version` on a context, then any scoped contexts do not "automatically" have this flag set, rather, they must be set on their own. Further confusion can arise when mixing multiple types that use multiple type-scoped contexts -- nested scoped contexts should not make assumptions about how/which existing prefix definitions will manifest. Therefore, I think that scoped contexts should be considered "self-contained", specifically declaring their own prefixes, versions, protected flags, etc. (any context-wide flags they use within their own context definition). We can mention this in the best practices doc.

With the other change that the group approved on the last call (allow redefinitions of protected terms so long as the definitions match, modulo the protected flag itself), redefining protected prefixes to the same thing within scoped contexts is not a problem.

We still need to add tests for context nullification, `@base`, and `@vocab` changes -- and I think we should add one for a type-scoped context that has multiple property-scoped contexts within it to ensure we're covered.

dlongley:
@gkellogg,

JFYI -- We've also discovered that type-scoped nullification has some other weird properties, such that it wipes out the `@type` definition for the object that brought that scoped context in to begin with! So if a type-scoped context was brought in via the type `Foo` and that context does nullification, then the output will treat `Foo` like it doesn't exist (it will treat it as a relative IRI). A similar problem could occur even without nullification, where the type-scoped context redefines its own type to be something else!

It seems like that's quite unexpected and potentially even evil -- and this problem is totally orthogonal to the rest of these other patches (it's a separate bug that's been around that no one has detected yet). The good news is that, with the above change I mentioned, we have access to the `previousContext` ... which means we can interpret `@type` values within the object against the previous context, which is probably the expected behavior. We're working on a patch and tests for that right now.

gkellogg:
Basically, this seems that we call `revertTypeScopedContext` on entry to compact, before handling `@context` or a property-scoped context, where before it was done after previously. On expand, we use any saved context for creating the property-scoped contexts.

For dealing with the context-clearing issue, I'm not sure I follow your processing. On expansion, the first thing I do is revert to the previous context, which would be `null` in this case. Are you saying there's special processing for `@type`, which continues to use the context before reverting? Does it do this only if the active context is `null`? Or, is there something else going on.

gkellogg:
Also, the minutes said something about wildcarding properties, but I didn't understand what this might refer to.

gkellogg:
Also, Expand tpr23-25 and tpr27 relate to changing a protected term definition that is equivalent, which should be in a different PR.

davidlehn:
Protected term redefinition tests moved here: https://github.com/w3c/json-ld-api/pull/92

dlongley:
@gkellogg,

> On expansion, the first thing I do is revert to the previous context, which would be null in this case.

I'll have more to say a bit later, but the "previous context" should be whatever the context was before the type-scoped context was processed. That means that if a type-scoped context happens to be an array of contexts, the "previous context" does not change as you process each one in the array as they all apply as "type-scoped" contexts. Rather, it is only set once. Perhaps we need a better name than "previous context". Your comment about the "previous context" being `null` makes me think your implementation is updating it as it goes through the type-scoped array of contexts, which it should not be doing.

Our implementation has also added some extra processing to ensure that values and subject references that occur within a typed object are compacted using the appropriate type-scoped context, rather than erroneously reverting. There's a test added to cover that case.

dlongley:
@gkellogg,

So, as mentioned above, we set the previous context once at the start of context processing, prior to processing any contexts in an array of contexts. Then, when expanding `@type` values, we use that previous context if it's set:

https://github.com/digitalbazaar/jsonld.js/pull/316/commits/5e4e2e2710dc542aa3558e52efafad907440228e

That ensures that the `@type` values are consistent, even when using type-scoped contexts that might clear or change the definition of the terms used.

During compaction, we ensure we keep a reference to the current active context to find all of the type-scoped contexts as we process them to produce a new active context. Then we do something analogous to expansion where we use the previous context, if present, to compact any values for `@type`:

https://github.com/digitalbazaar/jsonld.js/pull/316/commits/42392d5511629ef2adf71dec25b7fff3a8124196

Lastly, also during expansion, we make sure that values and subject references are processed against the type-scoped context, as they aren't really "nested" values:

https://github.com/digitalbazaar/jsonld.js/pull/316/commits/2ee232ae2682851260df9c4fd37bfc1db4adffab

With these changes, we pass the tests we've submitted here and under #92 and #93.

dlongley:
@gkellogg,

JFYI, we're still working through a bug on value expansion, will let you know when we've gotten past that.

dlongley:
@gkellogg, in order to expand values properly, we needed to pass in the type-scoped context and use it if present. We've added tests that expand and compact values of typed-object properties.

dlongley:
@gkellogg,

Another change for you to be aware of:

Since we ensure that when we combine property-scoped contexts with type-scoped contexts that we processed the property-scoped contexts against the previous context (as opposed to the type-scoped ones that will be reverted), we also need to ensure that when we expand or compact IRIs we do the same. This patch does that:

https://github.com/digitalbazaar/jsonld.js/commit/074527ebb55ef374fd3a7f68fdd907d5c2c342eb

We've added tests for this as well.

gkellogg:
Hopefully we're not making the algorithms full of special cases to handle this. In spite of the unexpected results, the previous was reasonably simple.

I'll look at this more tomorrow.

dlongley:
@gkellogg,

> Hopefully we're not making the algorithms full of special cases to handle this.

I don't think we've got too much of that (any more than usual for other features), but I also suspect we could unify and simplify a bit. Right now I've just got to make sure it works and the features are composable, which I believe we're in good shape for now. I have to ship our implementation to get the VCWG `@context` updated (which composes all these new JSON-LD 1.1 features together) and its related test suite running.

> In spite of the unexpected results, the previous was reasonably simple.

Yes, but of course a simple algorithm that yields unexpected results is its own set of problems :). Also, some of the changes we've made had to do with things that weren't tested previously and did not work as is (e.g., nullification/changing type terms). I do think that changes made (which, at the end of the day weren't too numerous) can be boiled down to some organizing principles that could be potentially simplified in algorithms/implementations.

Essentially, when processing contexts:
1. Type-scoped contexts must track previous contexts
2. Property-scoped contexts that are composed with type-scoped contexts must be applied to that previous context

When expanding/compacting:
1. We must revert type-scoped contexts when traversing out of the typed-object into another node object
2. We must use the aforementioned previous context when the active context was produced from a property-scoped context in conjunction with a type-scoped context

tests/compact/c015-in.jsonld
```diff
@@ -3,8 +3,14 @@
...
     "@id": "http://example/typed-base#typed-id",
     "@type": ["http://example/Type"],
-    "http://example/nested": [{
-      "@id": "http://example/base-base#nested-id"
+    "http://example/subjectReference": [{
+      "@id": "http://example/typed-base#subject-reference-id"
+    }],
+    "http://example/nestedNode": [{
+      "@id": "http://example/base-base#nested-id",
+      "http://example/foo": [{
```
gkellogg: Why should the nestedNode not use the base defined for `Type`? Nesting is/should be in-scope of the object containing the type; the purpose of nesting is to allow structure without affecting semantics, so values of a nested node should be treated as if they are part of the containing object on which `Type` is set.

tests/expand/c018-in.jsonld
```diff
@@ -0,0 +1,26 @@
...
+          "prop": "http://example/foo-prop"
+        }
+      ]
+    }
+  },
+  "@type": ["Foo", "Bar"],
+  "prop": "foo",
+  "nested": {
+    "prop": "will-not-exist"
+  }
```
gkellogg: Same comment here, the nested `prop` should be in scope of `Bar`.

> Why should the nestedNode not use the base defined for `Type`? Nesting is/should be in-scope of the object containing the type; the purpose of nesting is to allow structure without affecting semantics, so values of a nested node should be treated as if they are part of the containing object on which `Type` is set.

gkellogg:
@dlongley: I've updated the algorithms.

I do not get your results for compaction c025: https://github.com/w3c/json-ld-api/blob/c0a1531dda6fdce3e7bc9862310103978c507566/tests/compact-manifest.jsonld#L1134-L1141

I do not get the same compacted result because the term definition for "nested" includes `"@type": "@id"`. This prevents term selection from selecting this term, because IRI Compaction ends up setting _type/language_ to `@language`, which doesn't get entries created in the inverse context when `@type: @id`. It doesn't seem to be what the test is testing (so we might just remove that from the term definition), but it points to a some other difference in the term selection algorithms.

gkellogg:
As I mentioned above, the minutes said something about wildcarding properties, but I didn't understand what this might refer to. Otherwise, with an exception of c025, I think work is complete here, and per the WG discussion, we can merge it when you're happy with the spec text.

dlongley:
@gkellogg,

> @dlongley: I've updated the algorithms.

Thanks!

> I do not get your results for compaction c025:
> I do not get the same compacted result because the term definition for "nested" includes "@type": "@id". 

The value for `nested` is a node object, more specifically, a node object in a graph container, so it should compact using that term. Does your implementation compact to the `nested` term if you remove all the type scoping and just use a flat context? (I believe it should, and therefore should behave the same way when using type-scoped contexts):

```js
{
  "@context": {
    "@version": 1.1,
    "type": "@type",
    "Outer": {
      "@id": "ex:Outer"
    },
    "nested": {
      "@id": "ex:nested",
      "@type": "@id",
      "@container": "@graph"
    },
    "Inner": {
      "@id": "ex:Inner"
    }
  }
}
```

> As I mentioned above, the minutes said something about wildcarding properties, but I didn't understand what this might refer to.

I think we probably need to open a new issue for that. Essentially, now you can use property-scoped term definitions within a type-scoped term definition to cause terms to be defined beyond the type-scoped object with fine grained composability and without clashes w/`@protected` terms. This is great and what we want, but @azaroth42 has some cases where the same property-scoped context would be used across many properties in the same type-scoped context. Therefore, there's a desire to find a way to do this with some shorthand/sugar ("wildcard") that allows for pulling in a context that will define all of those properties without having to tediously declare all of those properties with the same `@context`.

This is related to:

> 2b. There is a shortcut suggestion to allow type scoped contexts to have a method to say that all properties of the instance should be treated as if they were property-scoped to the same context. This seems to be the right fix.

https://github.com/IIIF/api/issues/1846

> Otherwise, with an exception of c025, I think work is complete here, and per the WG discussion, we can merge it when you're happy with the spec text.

Sounds good. I looked over the text earlier and it looked good, but perhaps something is missing given c025 (which I think has correct output as is). I'll take a closer look when I can.

dlongley:
@gkellogg,

Yeah, I don't see one more change to the IRI compaction/expansion algorithms in the preview that is necessary. When expanding or compacting an IRI, if the context is from a property-term-scoped context that has been combined with a type-scoped context, we use the previous context to do the expansion/compaction. My implementation sets a flag on the active context when the context is from a property-term-scoped context and then uses the presence of a previous context to check for this condition:

https://github.com/digitalbazaar/jsonld.js/commit/074527ebb55ef374fd3a7f68fdd907d5c2c342eb

I suspect that will resolve the difference with test `c025`.

gkellogg:
I’ll check my implementation again, but I think I’ve hit everything that’s in the spec.

The c025 difference is due to the term selection algorithm, the proper context is loaded when calling Compact IRI. I’ll detail the inverse context and what’s in the call to term select. It has type/language set to language, which is empty in the inverse context, so the term doesn’t get selected. Remove `@type: @id` from the term definition, and it is selected.

dlongley:
@gkellogg,

> I’ll check my implementation again, but I think I’ve hit everything that’s in the spec.

But I think the spec is missing the change that handles IRI expansion/compaction when you've got a property-term-scoped context inside a type-scoped context (see: https://github.com/w3c/json-ld-api/pull/89#issuecomment-495843441). In this case, context processing updates the previous context, which means that IRI expansion/compaction must use that context. All that's necessary is to check for that case and then assign the active context to the previous context before continuing on with those algorithms. That will cause the proper inverse context to be used, etc.

gkellogg:
The issue doesn't seem related to selecting a context in compact/expand IRI, as I do end up with the correct term definition when compacting; I'm not sure why the code you cite might be needed, as I'm producing all the appropriate results otherwise.

What's going on when trying to do the compact IRI on `ex:nested` is more the issue. The inverse context entry I produce looks like the following:

```json
"ex:nested": {
  "@graph": {
    "@language": {},
    "@type": {"@id": "nested"},
    "@any": {"@none": "nested"}
  }
}
```

This seems correct according to the algorithm in the API.

Because `nested` has a type mapping, step 3.10 of the [Inverse Context algorithm](https://www.w3.org/TR/json-ld11-api/#algorithm-5) creates the `"@id": "nested"}`, but `@language` remains empty.

In compact IRI, I have the following arguments to the [Term Selection algorithm](https://www.w3.org/TR/json-ld11-api/#algorithm-7):

`container mapping = ["@graph", "@graph@set", "@set", "@graph@index", "@graph@index@set", "@graph@id", "@graph@id@set", "@index", "@index@set", "@none", "@index", "@index@set"]`

`type/language = "@language"`

`preferred values = ["@null", "@none", "@any"]`

It's because `type/language` is `@language`, that it doesn't find the entry in the inverse context. `type/language` is initialized to `@language` in the beginning of the algorithm, and never gets set to anything else. This is why removing `@type: @id` from the term definition allows it to work, because the inverse context will have entries for both `@type` and `@language`. Your implementation must either create the inverse context differently, or have different logic to set `type/language`.

As for the code to expand/compact IRI to use the previous context, can you point me to a test which should fail? It must be doing the equivalent logic elsewhere for the results to be the same, aside form c025.

dlongley:
@gkellogg,

This is the value I have that is passed into the Compact IRI Algorithm when choosing the term for `ex:nested`:

```js
{
  "@graph": [
    {
      "@type": [
        "ex:Inner"
      ],
      "ex:foo": [
        {
          "@value": "bar"
        }
      ]
    }
  ]
}
```

Which results in `type/language = "@type"` and preferred values = ["@id"]`.

Do you have a different value?

dlongley:
@gkellogg,

I'm looking at the IRI Compaction algorithm in the spec and it looks like it has changed in a breaking way from 1.0. I'm trying to tease out the differences now.

dlongley:
@gkellogg,

Yes, there's a new conditional fork for graph objects (step 2.8) in the 1.1 algorithm. Before that conditional was added, graph objects would fall through to the last step (step 2.9) and be detected as a non-value, which would cause steps 2.9.2 and 2.9.3 to run. These two steps are never run now for graph objects, so they need to be copied (or otherwise rearranged to run) into the end of 2.8. Otherwise the type/language is set to `@language` (as you're seeing in your implementation) for graph objects, which is clearly incorrect.

dlongley:
@gkellogg,

> The issue doesn't seem related to selecting a context in compact/expand IRI, as I do end up with the correct term definition when compacting; I'm not sure why the code you cite might be needed, as I'm producing all the appropriate results otherwise.

Yes, I don't think it's needed. I've looked more closely at your changes to the compaction algorithm and it applies a simplification my implementation doesn't have. It removes the early (step 1) processing of property scoped contexts and moves it down to a new step 6 after any type-scoped context has been reverted. I'll likely update my implementation to do this when I get the chance.

So it sounds like the only issue is the above missing step that will set `type/language` to `@type` properly for `@graph`. If you agree and that change fixes your test, I'll approve for merging.

gkellogg:
Good catch, yes it struck me that the graph path wasn’t examining more properties, but it was changed so long ago. I’ll update my implementation and the spec.

gkellogg:
Okay, I think that did it. See if you think the new steps in IRI Compaction are correct. Then, I think we're good to merge.

dlongley:
@gkellogg,

I think we're good here too. The only thing that looks a bit odd to me is that `@set` gets added to `containers` three separate times after that latest commit. I'm pretty sure that the way it's used in the Term Selection algorithm would make this unnecessary or erroneous. It should probably only be added at the very end, as it's the least specific choice? I defer to you on that as I still haven't updated my implementation to the latest version of that algorithm.

Everything else looks good. Thanks!

gkellogg:
The last `@set` was errant, and the one before, as you note, was redundant.

Issue #90, opened 2019-05-13
Tweak `@protected` algorithm to allow reuse of existing contexts w/o conflicts


dlongley:
A PR (https://github.com/digitalbazaar/jsonld.js/pull/316) in jsonld.js would prevent errors from being thrown when overriding protected terms if the override would result in the same definition as before. The only caveat to this is that if the new definition does not include `@protected`, this flag is carried forward to maintain the protection status. Otherwise, the definitions must match precisely, including any scoped contexts.

This tweak makes it much easier to reuse existing contexts that do not have any conflicting definitions in them with new contexts that use `@protected`. Without it, any existing `@context` that defines common things like aliases for `@id` or `@type` would be incompatible with contexts that protect those aliases.

The implementation change involved moving the `@protected` check to the end of the Create Term Definition algorithm and running a definition comparison check. All tests continue to pass with this change.

This change would still be a potentially valid interpretation of the current JSON-LD 1.1 syntax spec's text on `@protected`:

> A protected term definition is a term definition with a member `@protected` set to true. It generally prevents further contexts from overriding this term definition, either through a new definition of the same term, or through clearing the context with "@context": null.

As there is no "new definition of the same term", but rather, the same definition of the same term. The "same definition" is determined by comparing those things that a "term definition consists of" per the API spec. Again, the only caveat would be carrying forward `@protected` mode status even if not specified in the new definition.

So, if we want to do this, we may only need the above changes to the JSON-LD API spec, though we might want to add a clarifying sentence or two to the syntax spec to indicate that redefinitions of the same term with the same definition won't raise an error but that `@protected` status will remain unless the term is redefined without it when protection is not in effect.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-05-17-json-ld#section3-2).

- `RESOLVED:  Assuming the update to the algorithm is not controversial, protected definitions of terms do not raise an error when the term is redefined in exactly the same way, with the potential exception of not being protected`
<details><summary><i class="differentiate">View the transcript</i></summary>
3.2. Tweak @protected algorithm to allow reuse of existing contexts w/o conflicts<br/>
Dave Longley: <a href="https://github.com/w3c/json-ld-api/issues/90">https://github.com/w3c/json-ld-api/issues/90</a><br/>
Dave Longley: so this one is about reusing existing contexts<br/>
[...]

PR #92, opened 2019-05-21
Add protected term redefinition tests.

davidlehn:
A few of these tests look a bit crazy.  Sorry. ;-)  Suggested improvements would be welcome.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/92.html" title="Last updated on Jun 12, 2019, 5:21 PM UTC (71612ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/92/849e153...71612ae.html" title="Last updated on Jun 12, 2019, 5:21 PM UTC (71612ae)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1309,15 +1309,9 @@
           <a>term definition</a> for <var>term</var> in <var>active context</var>.</li>
         <li>Otherwise, remove any <var>previous definition</var> from
           <var>active context</var>.</li>
-        <li>If <var>value</var> is <code>null</code> or <var>value</var>
-          is a <a class="changed">dictionary</a> containing the key-value pair
-          <code>@id</code>-<code>null</code>, set the
-          <a>term definition</a> in <var>active context</var> to
-          <code>null</code>,
-          and set the value associated with <var>defined</var>'s
-          <a>member</a> <var>term</var> to <code>true</code>, and return.
-          <div class="ednote">Should probably store as a dictionary with <code>@id</code>-<code>null</code>
-            and retain <code>@protected</code> so we can protect against redefinitions.</div></li>
+        <li class="changed">If <var>value</var> is <code>null</code>,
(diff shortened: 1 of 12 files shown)


dlongley:
@gkellogg,

This is related to https://github.com/w3c/json-ld-api/issues/90.

dlongley:
@davidlehn, can you resolve the conflict?

davidlehn:
@dlongley Fixed.

dlongley:
@davidlehn, thanks!

gkellogg:
@dlongley note the comment in the last commit: "Note that the algorithm sets terms to `null` in context if they're defined as `null` or `{"@id": null}` and doesn't retain the protected status, which it probably should."

If you agree, I'll update that part too.

dlongley:
@gkellogg,

> @dlongley note the comment in the last commit: "Note that the algorithm sets terms to null in context if they're defined as null or {"@id": null} and doesn't retain the protected status, which it probably should."
> 
> If you agree, I'll update that part too.

It does seem like it should to retain protected status, i.e. what we're saying is that we'd allow for terms to be defined as null *and* protect that null definition, right? I know we have some other issue about it perhaps doing that for prefixes for security purposes.

We'd have to implement it to make sure it doesn't do anything strange or break any existing tests. We would also need new tests to check it. Have you happened to have implemented it and run it against the existing tests?

gkellogg:
My implementation does create an empty `term definition` for either `null` or `{"@id": null}`, although it doesn't presently set `@protected` on it.

We should update step 5.4 in [IRI Expansion](https://w3c.github.io/json-ld-api/#iri-expansion) to not use a term as a Compact IRI prefix if it's _IRI mapping_ is `null`.

PR #103, opened 2019-06-03
Add toRdf tests for merging rootless base paths

fsteeg:
This came up in an issue with jsonld-java:
https://github.com/jsonld-java/jsonld-java/issues/232

See second bullet point in:
https://tools.ietf.org/html/rfc3986#section-5.2.3

### DIFF
--- a/tests/toRdf-manifest.jsonld
+++ b/tests/toRdf-manifest.jsonld
@@ -896,6 +896,13 @@
       "input": "toRdf/0131-in.jsonld",
       "expect": "toRdf/0131-out.nq"
     }, {
+      "@id": "#t0132",
+      "@type": ["jld:PositiveEvaluationTest", "jld:ToRDFTest"],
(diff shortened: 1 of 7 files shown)


iherman:
Marked as non substantive for IPR from ash-nazg.

fsteeg:
Marked as non substantive for IPR from ash-nazg.

davidlehn:
There is a similar jsonld.js issue. There are likely more tricky test cases and test variations that could be written to cover these issues.
https://github.com/digitalbazaar/jsonld.js/issues/305.

fsteeg:
Added a test case for a base URI with a trailing slash, which (according to my understanding of the second bullet point in https://tools.ietf.org/html/rfc3986#section-5.2.3) results in preserving the full base URI (the scenario in https://github.com/digitalbazaar/jsonld.js/issues/305).

PR #110, opened 2019-06-15
Detect an attempt to compact an IRI which results in an absolute IRI

### LINKED ISSUE #104: compaction changes the meaning of absolute IRIs in some cases
## The problem

Some absolute IRIs are undistinguishable from compact IRIs (e.g. `mailto:a@b.c` or `tag:champin.net,2019:xyz`). This is in fact the case for any IRI that has no double-slash after the first colon.

Now consider the following data (DATA1):
```json
{
  "tag:champin.net,2019:prop": "hello world"
}
```
Here, the property is interpreted (by expansion or RDF-serialization) as an absolute IRI, since there is no `tag` prefix to interpret it as a compact IRI.

Now consider the following context (CTX1):
```json
{"@context": {
    "tag": "http://example.org/ns/tag/"
}}
```

If we compact DATA1 with CTX1, we get DATA2:
```json
{
  "@context": {
    "tag": "http://example.org/ns/tag/"
  },
  "tag:champin.net,2019:prop": "hello world"
}
```
([try it in the playground](https://json-ld.org/playground/#startTab=tab-compacted&json-ld=%7B%22tag%3Achampin.net%2C2019%3Aprop%22%3A%22hello%20world%22%7D&context=%7B%22%40context%22%3A%7B%22tag%22%3A%22http%3A%2F%2Fexample.org%2Fns%2Ftag%2F%22%7D%7D)).

This is wrong because in DATA2, the property is now interpreted (by expansion or RDF-serialization) as `http://example.org/ns/tag/champin.net,2019:prop` as [shown in the playground](https://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40context%22%3A%7B%22tag%22%3A%22http%3A%2F%2Fexample.org%2Fns%2Ftag%2F%22%7D%2C%22tag%3Achampin.net%2C2019%3Aprop%22%3A%22hello%20world%22%7D&context=%7B%22%40context%22%3A%7B%22tag%22%3A%22http%3A%2F%2Fexample.org%2Fns%2Ftag%2F%22%7D%7D).

The compaction algorithm should detect there will be a problem when 1) it would leave an absolute IRI unmodified AND 2) the active context contains a term definition matching that IRI's scheme – call this CONDITION1.

## Proposed solution 1

If CONDITION1 above is met, an error or a warning (although I would prefer an error) is raised. The rationale is that it is impossible to produce a JSON-LD document preserving this absolute IRI with the given context.

Probably this should only happen when the context has `"@version": 1.1` set, to not break backward compatibility -- even though this was a bug in JSON-LD 1.0 as well...

## Proposed solution 2

If CONDITION1 above is met, another context is included in the output data, which contains a fresh prefix resolving to the IRI's scheme. That fresh prefix is used to transform the absolute IRI:
```json
{
  "@context": [
    {
      "_abs_tag": "tag:"
    },
    {
      "tag": "http://example.org/ns/tag/"
    }
  ],
  "_abs_tag:champin.net,2019:prop": "hello world"
}
```
But is it acceptable for the compaction algorithm to output a different context than the one given as input? Also, it might be not trivial to coin a clash-less prefix.

## Proposed solution 3

Another option would be to have a microsyntax to force a string to be interpreted as an absolute IRI. This microsyntax would be used by the compaction algorithm whenever CONDITION1 above is detected. E.g.::
```json
{
  "@context": {
    "tag": "http://example.org/ns/tag/"
  },
  "<tag:champin.net,2019:prop>": "hello world"
}
```

It could also be used by data authors to address the issue raised by w3c/json-ld-syntax/issues/177.

Proposed microsyntaxes:

* put the IRI between pointy brackets, as is done in Turtle: `"<tag:champin.net,2019:prop>"`
* prefix the IRI with a new "keyword-like" prefix: `"@abs:tag:champin.net,2019:prop"`

gkellogg:
…where the scheme matches a term in the active context.

Fixes #104.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/110.html" title="Last updated on Jun 17, 2019, 8:29 PM UTC (d2302d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/110/1466f97...d2302d9.html" title="Last updated on Jun 17, 2019, 8:29 PM UTC (d2302d9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3547,7 +3547,7 @@
           not confused with a <a>compact IRI</a>,
           if the <a data-cite="RFC3986#section-3.1">IRI scheme</a> of <var>var</var>
           matches any term in <a>active context</a> with <a>prefix flag</a> set to <code>true</code>,
-          and <var>var</var> has no <a datacite="https://tools.ietf.org/html/rfc3986#section-3.2">IRI authority</a> (beginning with double-forward-slash (<code>//</code>),
+          and <var>var</var> has no <a data-cite="RFC3986#section-3.1">IRI authority</a> (beginning with double-forward-slash (<code>//</code>),
           an <a data-link-for="JsonLdErrorCode">IRI confused with prefix</a> error has been detected,
           and processing is aborted.</li>
         <li>If <var>vocab</var> is <code>false</code>,
@@ -6032,7 +6032,7 @@
         <dt><dfn>invalid vocab mapping</dfn></dt>
         <dd>An invalid <a>vocabulary mapping</a> has been detected,
           i.e., it is neither an <a>IRI</a> nor <code>null</code>.</dd>
-        <dt class="changed">IRI confused with prefix</dt>
+        <dt class="changed"><dfn>IRI confused with prefix</dfn></dt>
         <dd>When compacting an <a>absolute IRI</a> would result in an <a>absolute IRI</a>
           which could be confused with a compact IRI (because its <a data-cite="RFC3986#section-3.1">IRI scheme</a> matches a <a>term definition</a> and it has no <a data-cite="RFC3986#section-3.1">IRI authority</a>).</dd>
         <dt><dfn>keyword redefinition</dfn></dt>
--- a/index.html
+++ b/index.html
@@ -3293,6 +3293,11 @@
         <a>base IRI</a>. Finally, if the <a>IRI</a> or
         <a>keyword</a> still could not be compacted, it is returned
         as is.</p>
+
+      <p class="changed">In the case were this algorithm would return the input <a>IRI</a> as is,
+        and that <a>IRI</a> can be mistaken for a <a>compact IRI</a> in the <a>active context</a>,
+        this algorithm will raise an error,
+        because it has no way to return an unambiguous representation of the original <a>IRI</a>.</p>
     </section>
 
     <section class="algorithm">
--- a/index.html
+++ b/index.html
@@ -3546,7 +3546,7 @@
         <li class="changed">To ensure that the <a>absolute IRI</a> <var>var</var> is
           not confused with a <a>compact IRI</a>,
           if the <a data-cite="RFC3986#section-3.1">IRI scheme</a> of <var>var</var>
-          matches any term in <a>active context</a>,
+          matches any term in <a>active context</a> with <a>prefix flag</a> set to <code>true</code>,
           and <var>var</var> has no <a datacite="https://tools.ietf.org/html/rfc3986#section-3.2">IRI authority</a> (beginning with double-forward-slash (<code>//</code>),
           an <a data-link-for="JsonLdErrorCode">IRI confused with prefix</a> error has been detected,
           and processing is aborted.</li>
--- a/index.html
+++ b/index.html
@@ -3547,7 +3547,7 @@
           not confused with a <a>compact IRI</a>,
           if the <a data-cite="RFC3986#section-3.1">IRI scheme</a> of <var>var</var>
           matches any term in <a>active context</a>,
-          an <a data-link-for="JsonLdErrorCode">iri confused with prefix</a> error has been detected,
+          an <a data-link-for="JsonLdErrorCode">IRI confused with prefix</a> error has been detected,
           and processing is aborted.</li>
         <li>If <var>vocab</var> is <code>false</code>,
           transform <var>var</var> to a <a>relative IRI</a> using
@@ -5923,7 +5923,7 @@
             "invalid value object",
             "invalid value object value",
             "invalid vocab mapping",
-            "iri confused with prefix",
+            "IRI confused with prefix",
             "keyword redefinition",
             "loading document failed",
             "loading remote context failed",
@@ -6031,7 +6031,7 @@
         <dt><dfn>invalid vocab mapping</dfn></dt>
         <dd>An invalid <a>vocabulary mapping</a> has been detected,
           i.e., it is neither an <a>IRI</a> nor <code>null</code>.</dd>
-        <dt class="changed">iri confused with prefix</dt>
+        <dt class="changed">IRI confused with prefix</dt>
         <dd>When compacting an <a>absolute IRI</a> would result in an <a>absolute IRI</a>
           with an <a data-cite="RFC3986#section-3.1">IRI scheme</a> could be confused with a term.</dd>
         <dt><dfn>keyword redefinition</dfn></dt>
--- a/index.html
+++ b/index.html
@@ -3516,7 +3516,7 @@
           <a data-lt="active context">active context's</a> <a>vocabulary mapping</a>.
           Try to create a <a>compact IRI</a>, starting by initializing
           <var>compact IRI</var> to <code>null</code>. This variable will be used to
-          tore the created <a>compact IRI</a>, if any.</li>
+          store the created <a>compact IRI</a>, if any.</li>
         <li>For each key <a>term</a> and value <a>term definition</a> in
           the <var>active context</var>:
           <ol>
@@ -3543,6 +3543,12 @@
           </ol>
         </li>
         <li>If <var>compact IRI</var> is not <code>null</code>, return <var>compact IRI</var>.</li>
+        <li class="changed">To ensure that the <a>absolute IRI</a> <var>var</var> is
+          not confused with a <a>compact IRI</a>,
+          if the <a data-cite="RFC3986#section-3.1">IRI scheme</a> of <var>var</var>
+          matches any term in <a>active context</a>,
+          an <a data-link-for="JsonLdErrorCode">iri confused with prefix</a> error has been detected,
+          and processing is aborted.</li>
         <li>If <var>vocab</var> is <code>false</code>,
           transform <var>var</var> to a <a>relative IRI</a> using
           the <span class="changed"><a>base IRI</a> from <var>active context</var>, if it exists</span>.</li>
@@ -5917,6 +5923,7 @@
             "invalid value object",
             "invalid value object value",
             "invalid vocab mapping",
+            "iri confused with prefix",
             "keyword redefinition",
             "loading document failed",
             "loading remote context failed",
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -5917,6 +5923,7 @@ <h4>JsonLdErrorCode</h4>
             "invalid value object",
             "invalid value object value",
             "invalid vocab mapping",
+            "iri confused with prefix",
```
davidlehn: Match test and capitalization in other codes.
```suggestion
            "IRI confused with prefix",
```

index.html
```diff
@@ -3543,6 +3543,12 @@ <h3>Algorithm</h3>
           </ol>
         </li>
         <li>If <var>compact IRI</var> is not <code>null</code>, return <var>compact IRI</var>.</li>
+        <li class="changed">To ensure that the <a>absolute IRI</a> <var>var</var> is
+          not confused with a <a>compact IRI</a>,
+          if the <a data-cite="RFC3986#section-3.1">IRI scheme</a> of <var>var</var>
+          matches any term in <a>active context</a>,
+          an <a data-link-for="JsonLdErrorCode">iri confused with prefix</a> error has been detected,
```
davidlehn: Match test and capitalization in other codes.
```suggestion
          an <a data-link-for="JsonLdErrorCode">IRI confused with prefix</a> error has been detected,
```

index.html
```diff
@@ -3543,6 +3543,12 @@ <h3>Algorithm</h3>
           </ol>
         </li>
         <li>If <var>compact IRI</var> is not <code>null</code>, return <var>compact IRI</var>.</li>
+        <li class="changed">To ensure that the <a>absolute IRI</a> <var>var</var> is
+          not confused with a <a>compact IRI</a>,
+          if the <a data-cite="RFC3986#section-3.1">IRI scheme</a> of <var>var</var>
+          matches any term in <a>active context</a>,
```
```suggestion
          matches any term in <a>active context</a>,
          and <var>var</var> has no <a datacite="https://tools.ietf.org/html/rfc3986#section-3.2">IRI authority</a> (beginning with double-forward-slash (<code>//</code>),
```

index.html
```diff
@@ -6024,6 +6031,9 @@ <h4>JsonLdErrorCode</h4>
         <dt><dfn>invalid vocab mapping</dfn></dt>
         <dd>An invalid <a>vocabulary mapping</a> has been detected,
           i.e., it is neither an <a>IRI</a> nor <code>null</code>.</dd>
+        <dt class="changed">iri confused with prefix</dt>
+        <dd>When compacting an <a>absolute IRI</a> would result in an <a>absolute IRI</a>
+          with an <a data-cite="RFC3986#section-3.1">IRI scheme</a> could be confused with a term.</dd>
```
```suggestion
          which could be confused with a compact IRI (because its <a data-cite="RFC3986#section-3.1">IRI scheme</a> matches a <a>term definition</a> and it has no <a data-cite="RFC3986#section-3.1">IRI authority</a>).</dd>
```
* "which" was missing from the original sentence
* this phrasing is in line with my proposed change above

index.html
```diff
@@ -6190,6 +6200,9 @@ <h2>Changes since JSON-LD Community Group Final Report</h2>
       This allows a mechanism for documenting the content of a context using HTML.</li>
     <li>Consolidate <a>RemoteDocument</a> processing into the <a>LoadDocumentCallback</a>
       including variations on HTML processing.</li>
+    <li>The <a href="#iri-compaction">IRI compaction algorithm</a> may generate an error if the result is an
+      <a>absolute IRI</a> where the scheme part could be confused with a term in the
```
```suggestion
      <a>absolute IRI</a> which could be confused with a compact IRI in the
```
* since the condition is actually a little more complex

index.html
gkellogg: Alternatively, we might remove the double-slash caveat, which was just one way of addressing this issue, and just use the scheme/term distinction. Otherwise, it tacitly allows the use of terms which are also IRI schemes.

pchampin: I strongly disagree. The double-slash "caveat" ensures that *only a subset* of absolute IRIs can potentially be confused with compact IRIs. I think we agreed, during the last meeting, that this ambiguity between absolute IRIs and compact IRIs, was unfortunate. Removing the double-slash rule would increase this ambiguity. And doing so, we increase the number of cases were compaction would fail with `iri confused with prefix`...

> Otherwise, it tacitly allows the use of terms which are also IRI schemes.

which might be absolutely valid ([example]). The problem is using them as prefixes...

[example]: https://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40context%22%3A%7B%22%40vocab%22%3A%22http%3A%2F%2Fexample.org%2FserviceOntology%23%22%2C%22protocol%22%3A%7B%22%40type%22%3A%22%40vocab%22%7D%2C%22http%22%3A%22http%3A%2F%2Fdbpedia.org%2Fresource%2FHypertext_Transfer_Protocol%22%2C%22https%22%3A%22http%3A%2F%2Fdbpedia.org%2Fresource%2FHTTPS%22%2C%22ftp%22%3A%22http%3A%2F%2Fdbpedia.org%2Fresource%2FFile_Transfer_Protocol%22%7D%2C%22%40type%22%3A%22Service%22%2C%22address%22%3A%22123.45.67.89%22%2C%22protocol%22%3A%22http%22%7D

index.html
gkellogg: Did through separate commit, as there was a third.

index.html
gkellogg: Did through separate commit, as there was a third.

gkellogg:
> * We don't need to raise an error when the absolute IRI has an 'authority' (i.e. the "suffix" starts with `//`) because the IRI expansion algorithm will _never_ consider those as compact IRIs. I proposed changes to exclude these absolute IRIs from the error case.

I accepted this, but it may be that it's simpler to always just error when an absolute IRI has a scheme patching a prefixable term. The more specific test may imply that it's otherwise okay to do this, which is odd.

> * The references to "IRI scheme" are pointing to [RFC3986](https://tools.ietf.org/html/rfc3986) "URIs" rather than [RFC3987](https://tools.ietf.org/html/rfc3987) "IRIs". I did the same when adding references to "IRI authority". I am assuming that this is deliberate, since RFC3987 does not describe in detail the structure of IRIs, it relies on RFC3986 for that, but I preferred to check.

Yes, same reasoning. RFC3987 does not define these parts as it is based on RFC3986, which does, and thus the direct reference to the definition in RFC3986.

iherman:
This may be a change of far-reaching editorial consequences, so we may want to ignore it, but I thought it is worth raising: I was made to realize, lately, that the [WhatWG URL spec](https://url.spec.whatwg.org/) encompasses URI-s, URN-s, IRI-s, etc. See, for example, [my forensics](https://github.com/w3c/wpub/pull/456#issuecomment-500350365), and also some [additional references](https://github.com/w3c/wpub/pull/456#issuecomment-500383632). The comment that made me realize my own, faulty misconception comes from [this comment](https://github.com/w3c/wpub/pull/456#issuecomment-499382473).

As I said, it may be too late by now, but maybe it is worth considering changing the terminology and thereby getting rid of the issue of which rfc to refer to...

pchampin:
@gkellogg In general I'm all in favour of simplicity. But in this case, the additional complexity serves a valid purpose, IMO: limiting the ambiguity btw absolute IRIs and compact IRIs.

@iherman I think this question deserves an issue of its own...

index.html
```diff
@@ -3543,6 +3543,13 @@ <h3>Algorithm</h3>
           </ol>
         </li>
         <li>If <var>compact IRI</var> is not <code>null</code>, return <var>compact IRI</var>.</li>
+        <li class="changed">To ensure that the <a>absolute IRI</a> <var>var</var> is
+          not confused with a <a>compact IRI</a>,
+          if the <a data-cite="RFC3986#section-3.1">IRI scheme</a> of <var>var</var>
+          matches any term in <a>active context</a>,
```
```suggestion
          matches any term in <a>active context</a> with <a>prefix flag</a> set to <code>true</code>,
```

index.html
```diff
@@ -6024,6 +6032,9 @@ <h4>JsonLdErrorCode</h4>
         <dt><dfn>invalid vocab mapping</dfn></dt>
         <dd>An invalid <a>vocabulary mapping</a> has been detected,
           i.e., it is neither an <a>IRI</a> nor <code>null</code>.</dd>
+        <dt class="changed">IRI confused with prefix</dt>
+        <dd>When compacting an <a>absolute IRI</a> would result in an <a>absolute IRI</a>
+          which could be confused with a compact IRI (because its <a data-cite="RFC3986#section-3.1">IRI scheme</a> matches a <a>term definition</a> and it has no <a data-cite="RFC3986#section-3.1">IRI authority</a>).</dd>
```
pchampin: This is normative text, so should we *again* specify that the prefix flag must be set to `true` (which would be very verbose)?

gkellogg: I'm not sure it's necessary here, as we describe elsewhere in the algorithm that the _prefix flag_ must be `true`. This is just saying that an error is generated because the IRI scheme matches a term; the fact that the _prefix flag_ is `true` is implicit.

gkellogg:
> This may be a change of far-reaching editorial consequences, so we may want to ignore it, but I thought it is worth raising: I was made to realize, lately, that the [WhatWG URL spec](https://url.spec.whatwg.org/) encompasses URI-s, URN-s, IRI-s, etc. See, for example, [my forensics](https://github.com/w3c/wpub/pull/456#issuecomment-500350365), and also some [additional references](https://github.com/w3c/wpub/pull/456#issuecomment-500383632). The comment that made me realize my own, faulty misconception comes from [this comment](https://github.com/w3c/wpub/pull/456#issuecomment-499382473).
> 
> As I said, it may be too late by now, but maybe it is worth considering changing the terminology and thereby getting rid of the issue of which rfc to refer to...

We should definitely consider doing this; the references can be updated with some amount of work, but better to get it done right now. Agree that this should be a different issue.

index.html
pchampin: Fine with me.

pchampin:
This is looking good. Just a thought: shouldn't the "6.3.1.Overview" section also mention the change? For example, an additional paragraph saying:

> In the case were this algorithm would return the input IRI as is, and that IRI can be mistaken for a compact IRI in the active context, this algorithm will raise an error, because it has no way to return an unambiguous representation of the original IRI.

gkellogg:
> This is looking good. Just a thought: shouldn't the "6.3.1.Overview" section also mention the change? For example, an additional paragraph saying:
> 
> > In the case were this algorithm would return the input IRI as is, and that IRI can be mistaken for a compact IRI in the active context, this algorithm will raise an error, because it has no way to return an unambiguous representation of the original IRI.

Yes, that's reasonable, I'll add that paragraph.

PR #112, opened 2019-06-24
Updates to add @propagate and @import

gkellogg:
* Text and tests for `@propagate` and `@source`

For w3c/json-ld-syntax#174.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/112.html" title="Last updated on Jul 12, 2019, 9:10 PM UTC (08c9296)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/112/43c75e6...08c9296.html" title="Last updated on Jul 12, 2019, 9:10 PM UTC (08c9296)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1003,7 +1003,7 @@
       keys and values have to be interpreted as well as the current <a>base IRI</a>,
       the <a>vocabulary mapping</a>, the <a>default language</a>,
       <span class="changed">and an optional <dfn>previous context</dfn>,
-        used when a type-scoped <a>context</a> is defined</span>.
+        used when a non-propagated <a>context</a> is defined</span>.
       Each <a>term definition</a> consists of:</p>
     <ul>
       <li>an <dfn>IRI mapping</dfn>,</li>
@@ -1069,6 +1069,11 @@
         they affect how the other <a>entries</a> are processed. Please note that <code>@base</code> is
         ignored when processing remote contexts.</p>
 
+      <p class="changed">If <a>context</a> is not to be propagated,
+        a reference to the <var>previous context</var> is retained so that
+        it may be rolled back when a new <a>node object</a> is entered.
+        By default, all contexts are propagated, other than type-scoped contexts.</p>
+
       <p>Then, for every other <a>entry</a> in <a>local context</a>, we update
         the <a>term definition</a> in <var>result</var>. Since
         <a>term definitions</a> in a <a>local context</a>
@@ -1096,19 +1101,23 @@
         <span class="changed">, <var>from property</var>, defaulting to <code>false</code>,
           which is used to allow changes to protected terms,
           and <var>from type</var>, defaulting to <code>false</code>,
-          which is used to mark <a>term definitions</a> associated with a type-scoped <a>context</a>.</span>.
+          which is used to set the default for <var>propagated</var>
+          to mark <a>term definitions</a> associated with non-propagated <a>contexts</a>.</span>.
       </p>
 
       <ol>
         <li>Initialize <var>result</var> to the result of cloning
           <var>active context</var>.</li>
+        <li class="changed">If <var>local context</var> is an object containing the member <code>@propagate</code>,
+          its value MUST be <a>boolean</a> <code>true</code> or <code>false</code>,
+          set <var>propagate</var> to that value.</li>
+        <li class="changed">Otherwise, set <var>propagate</var> to <var>from type</var>.</li>
+        <li class="changed">If <var>propagate</var> is <code>false</code>, and <var>result</var>
+          does not have a <a>previous context</a>, set <a>previous context</a>
+          in <var>result</var> to <var>active context</var>.</li>
         <li>If <var>local context</var> is not an <a>array</a>,
           set it to an <a>array</a> containing only
           <var>local context</var>.</li>
-        <li class="changed">
-          If <var>from type</var> is <code>true</code>, and <var>result</var>
-          does not have a <a>previous context</a>, set <a>previous context</a>
-          in <var>result</var> to <var>active context</var>.</li>
         <li>
           For each item <var>context</var> in <var>local context</var>:
           <ol>
@@ -1242,12 +1251,24 @@
                   error has been detected and processing is aborted.</li>
               </ol>
             </li>
+            <li>If <var>context</var> has an <code>@propagate</code> <a>entry</a>
+              and its value is not <a>boolean</a> <code>true</code> or <code>false</code>,
+              or <a>processing mode</a> is <code>json-ld-1.0</code>,
+              an <a data-link-for="JsonLdErrorCode">invalid local context</a>
+              error has been detected and processing is aborted.
+              <span class="note"><var>previous context</var> was determined before.</span></li>
             <li>Create a <a class="changed">map</a> <var>defined</var> to use to keep
               track of whether or not a <a>term</a> has already been defined
               or currently being defined during recursion.</li>
             <li>For each <var>key</var>-<var>value</var> pair in <var>context</var> where
-              <var>key</var> is not <code>@base</code>, <code>@vocab</code>,
-              <code>@language</code>, <code>@protected</code>, or <code>@version</code>, invoke the
+              <var>key</var> is not
+              <code>@base</code>,
+              <code>@vocab</code>,
+              <code>@language</code>,
+              <code class="changed">@propagate</code>,
+              <code class="changed">@protected</code>, or
+              <code>@version</code>,
+              invoke the
               <a href="#create-term-definition">Create Term Definition algorithm</a>,
               passing <var>result</var> for <var>active context</var>,
               <var>context</var> for <var>local context</var>, <var>key</var>,
@@ -1791,7 +1812,7 @@
           the <a data-link-for="JsonLdOptions">ordered</a> flag, used to order
           <a>map entry</a> keys lexicographically, where noted,
           and the <var>from map</var> flag, used to control reverting
-          previous <a>term definitions</a> in the <a>active context</a> associated with a type-scoped <a>context</a>.</span>
+          previous <a>term definitions</a> in the <a>active context</a> associated with non-propagated <a>contexts</a>.</span>
         To begin, the <var>active property</var> is set to <code>null</code>,
         and <var>element</var> is set to the <a>JSON-LD input</a>.
         <span class="changed">If not passed, the both flags are set to <code>false</code></span>.</p>
@@ -1866,7 +1887,7 @@
         </li>
         <li>Otherwise <var>element</var> is a <a class="changed">map</a>.</li>
         <li class="changed">If <var>active context</var> has a <a>previous context</a>,
-          the <var>active context</var> is type-scoped.
+          the <var>active context</var> is not propagated.
           If <var>from map</var> is undefined or <code>false</code>,
           and <var>element</var> does not contain a <a>entry</a> expanding to <code>@value</code>,
           and <var>element</var> does not consist of a single entry expanding to <code>@id</code>,
@@ -2570,7 +2591,7 @@
         </li>
         <li>Otherwise <var>element</var> is a <a class="changed">map</a>.</li>
         <li class="changed">If <var>active context</var> has a <a>previous context</a>,
-          the <var>active context</var> is type-scoped.
+          the <var>active context</var> is not propagated.
           If <var>element</var> does not contain an <code>@value</code> <a>entry</a>,
           and <var>element</var> does not consist of a single <code>@id</code> entry,
           set <var>active context</var> to <a>previous context</a> from <var>active context</var>,
@@ -6215,6 +6236,9 @@
     <li>The <a href="#iri-compaction">IRI compaction algorithm</a> may generate an error if the result is an
       <a>absolute IRI</a> which could be confused with a compact IRI in the
       <a>active context</a>.</li>
+    <li>By default, all contexts are propagated when traversing <a>node objects</a>, other than
+      type-scoped contexts. This can be controlled using the <code>@preserve</code>
+      <a>member</a> in a <a>local context</a>.</li>
   </ul>
 </section>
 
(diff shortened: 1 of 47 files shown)


index.html
```diff
@@ -6216,6 +6237,9 @@ <h2>Changes since JSON-LD Community Group Final Report</h2>
     <li>The <a href="#iri-compaction">IRI compaction algorithm</a> may generate an error if the result is an
       <a>absolute IRI</a> which could be confused with a compact IRI in the
       <a>active context</a>.</li>
+    <li>By default, all contexts are propagated when traversing <a>node objects</a>, other than
+      type-scoped contexts. This can be controlled using the <code>@preserve</code>
```
```suggestion
      type-scoped contexts. This can be controlled using the <code>@propagate</code>
```

gkellogg: Perhaps an indication of difficulty, but I find myself often typing `@preserve` when I mean `@propogate`.

gkellogg:
Added support for `@source` and expansion tests. (Could do compaction tests, but I there is no change in that execution path.

Note that this version allows `@source` _in addition_ to other elements of a context. It could be updated to restrict other elements within `@context` with some extra steps, but I think there's value in being able to reference a context and potentially modify some of the terms as well. Processing considers `@source` immediately after `@version` (which was moved up before `@base`).

iherman:
Per @pchampin 

> My initial idea was that {"@source": "some:iri"} was somehow equivalent to "some:iri" (in the context of a @context member), that would allow for additional metadata about that IRI. Nothing more.

That was also my view...

gkellogg:
> The proposal "works", but I am still a bit concern about the overhead that it adds to the standard as a whole. Granted, it does not make the algorithms much more complex. But the ramifications of `@source` do worry me (as expressed in [w3c/json-ld-syntax#174 (comment)](https://github.com/w3c/json-ld-syntax/issues/174#issuecomment-505117274)).
> 
> For example, why couldn't source accept an _array_ of strings?
> 
> Also, the original idea of introducing `@source` was to allow, in the future, for adding some metadata in context _references_, like:
> 
> ```
> { "@context": { "@source": "...", "@sri": "123456abcdef" }, ... }
> ```
> 
> but what would the `@sri` property mean then if `@source` was co-existing with term definitions? E.g.:

I don’t think this creates a problem, the `@sri` applies to the intermediate result of processing `@source`.

> ```
> { "@context": { "@source": "...", "@sri": "123456abcdef", "foo": "http://example.org/foo" }, ... }
> ```
>
> My initial idea was that `{"@source": "some:iri"}` was somehow equivalent to `"some:iri"` (in the context of a `@context` member), that would allow for additional metadata about that IRI. Nothing more.
> 
> In the current proposal, `@source` is a general purpose _import_ mechanism, and I would then agree with @dlongley's proposal to call it `@import`. Except that I'm not sure we need it in the first place.

I thought this was a good starting point to avoid complicating the algorithm. We can always decide to add more restrictions if `@source` is present, such as restricting other members of the local context from including members other than `@version` and `@propagate`.

dlongley:
So if `@source` allows for modifying the incoming context, I think it should be called `@import`. It seems like this better matches Rob's use case for "constructing a context from another one with modifications".

Also, I suspect that this feature is really restricted to pulling in an existing JSON-LD 1.0 context for modification and use with JSON-LD 1.1. I don't see how it would be that useful to import a JSON-LD 1.1 context that might have scoped-contexts scattered throughout it -- whereby modifications like setting `@propagate` wouldn't deeply apply (or it would be too difficult to target them without redefining entire terms).

To me, it seems like this feature is really just about reusability, for "importing an existing JSON-LD 1.0 context and applying some JSON-LD 1.1 changes". Changes that might be useful include `@propagate` and `@protected` ... and any other keyword we have that globally modifies term definitions in a context. Perhaps we should consider it largely for that purpose, call it `@import`, and leave the `@sri` and `@source` stuff for another day.

pchampin:
@gkellogg wrote:
> I don’t think this creates a problem, the `@sri` applies to the intermediate result of processing `@source`

That would be quite misleading if the context containing `@sri` contains its own term definitions...

@dlongley 
> To me, it seems like this feature is really just about reusability, for "importing an existing JSON-LD 1.0 context and applying some JSON-LD 1.1 changes". (...) Perhaps we should consider it largely for that purpose, call it `@import`, and leave the `@sri` and `@source` stuff for another day.

Even if the purpose is to "upgrade" a 1.0 context, I don't see the point of adding additional term definitions in it. Seing `{"@source": "some:iri"}` as an extensible version of `"some:iri"` meets this requirement.

But I don't even see a requirement for "upgrading" or modifying a 1.0 context. @azaroth42's concerns was that, with the way type-scoped contexts now work, 1.0 contexts would give unexpected results (compared to the expectation by their authors that they always propagate). By deciding that `@propagate` defaults to `true` in any context with no `"@version"` member, I think we solve this. No need to be able to alter any context "from the outside".

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-06-28-json-ld#section4-1).

- No actions or resolutions
4.1. Continuing discussions from last week around “propagates”<br/>
Benjamin Young: <a href="https://github.com/w3c/json-ld-syntax/issues/174#issuecomment-504513373">https://github.com/w3c/json-ld-syntax/issues/174#issuecomment-504513373</a><br/>
Gregg Kellogg: There’s also a PR <a href="https://github.com/w3c/json-ld-api/pull/112">https://github.com/w3c/json-ld-api/pull/112</a><br/>
Gregg Kellogg: based on Rob’s proposal, but instead of “<code>@src”,</code> uses “<code>@source”…</code><br/>
… <code>@propagates</code> defaults to True except for type scoped contexts, can be overrriden in either case<br/>
Benjamin Young: focus on <code>@propagates</code> for now<br/>
Dave Longley: <a href="https://github.com/w3c/json-ld-syntax/issues/174">https://github.com/w3c/json-ld-syntax/issues/174</a><br/>
Dave Longley: propagate makes sense to me, but there other considerations in type scoped contexts…<br/>
… I didn’t check whether gkellogg’s implementation addresses these.<br/>
… previous contexts can now be any context, including type scoped, where changes can occur underneath<br/>
… have to make sure that <code>@type</code> gets evaluated using previous contexts<br/>
… correct keyword sb <code>@import</code> instead of <code>@source</code><br/>
… feature makes a lot of sense to bring ld 1.0 contexts into the 1.1 fold without having to rewrite<br/>
… may not make a lot of sense to import 1.1 context, so maybe we should focus on <code>@import</code> 1.0 context<br/>
Gregg Kellogg: May need more tests. Checks in compaction and expansion … if type scoped context is overridden to<br/>
… <code>@source</code> vs. <code>@import</code> - separate discussion. Should discuss SRI types as well<br/>
Dave Longley: It would be unexpected to evaluate <code>@type</code> against type scoped context – it would break round-tripping…<br/>
… expectation is that typed value will always be evaluated against previous context.<br/>
Gregg Kellogg: can dave represent concern in issue or PR?<br/>
Dave Longley: <a href="https://github.com/w3c/json-ld-syntax/issues/174#issuecomment-506747857">https://github.com/w3c/json-ld-syntax/issues/174#issuecomment-506747857</a><br/>
Gregg Kellogg: if you try to round-trip example above, it would behave as expected. If, however, we were to process <code>@type</code> using<br/>
… type scoped context, it would destroy its own type, which would be unexpected.<br/>
Ivan Herman: example is drastic, but even if you have a type definition in the scoped context, how does it relate to the type in the enclosing context?<br/>
Gregg Kellogg: Prior to PR, worked the way that was expected. The way to update w/ <code>@propagates,</code> would be to add <code>@propagates</code> to second embedded context….<br/>
… but question is what is controlling propagation. We need to flesh this out to understand what adding <code>@propagate</code> true to second context<br/>
… need to preserve processing chain independent of propagation<br/>
Benjamin Young: … appears to be consensus developing around PR<br/>
Ivan Herman: I am worried about syntax specification wrt. <code>@propagates</code> that is understandable to user. Would like to see PR that makes this clear spec-wise<br/>
… before I would accept API PR, I would like to seen syntax PR w/ examples<br/>
Dave Longley: <code>@import</code> speaks to what we can say in the spec, wrt. using <code>@propagate</code> for pulling LD 1.0 and protecting it<br/>
Ivan Herman: we need to see the whole story<br/>
Gregg Kellogg: JSON-LD 1.0 evolved by thinking of feature, implement and then describe syntax. Approach of syntax and then implementation is difficult. Would prefer to meet in the middle…<br/>
… would like to use sample implementation and examples to see whether this is the direction we want to to, followed by syntax spec.<br/>
Dave Longley: i’ve also thought about JSON-LD as … “here’s a feature JSON devs want/use to describe their JSON … how do we implement something to express the semantics in there properly?”<br/>
Gregg Kellogg: implementation allows us to decide what we prefer before syntax document. Lets not put this on hold<br/>
Dave Longley: +1 that both sides are important … we need to be able to describe the syntax simply enough and be able to do things in the implementation to demonstrate it’s even possible<br/>
Ivan Herman: You can get situations where awareness of implementation provides clarity, but if you aren’t familiar with the details it may not make a good story to the end user<br/>
… would like a clear story defined in document before we do the whole thing.<br/>
Dave Longley: +1<br/>
Gregg Kellogg: if you look at a grammar such as turtle or sparql, if you don’t take parsing issues into account, you’ve done a disservice. Advocate both ends<br/>
Dave Longley: +1<br/>
Ivan Herman: Need PR for syntax<br/>
Gregg Kellogg: need to agree on <code>@imports</code> vs. <code>@source</code> semantics before we do this in a syntax document. API helps us consider that<br/>
Dave Longley: If we do <code>@import,</code> can we add <code>@source</code> in the future? Would you just put both tags?<br/>
Gregg Kellogg: Could be done either way - integrity (SRI) becomes a modifier to source URL. In the presence of <code>@sri,</code> that value is extracted and passed to algorithm for evaluating ressults…<br/>
… would not import an array of things, so maybe <code>@source</code> makes this clear.<br/>
Benjamin Young: Is this more than bike shedding? Two different modes as represented by <code>@source</code> vs. <code>@import.</code> We should focus on semantics, not names.<br/>
… two terms represent two semantic categories w/ different behaviors.<br/>
Dave Longley: my view on the semantic difference: do you “update” a context you bring in (<code>@import)</code> or are you just making meta data assertions on it (<code>@source)</code> … not everyone will agree, maybe <code>@source</code> can do both.<br/>
Benjamin Young: importing a 1.0 context w/ a small 1.1 wrapper sounds “dreamy”…<br/>
Gregg Kellogg: agree w/ dlongley’s summary – the diffference between pulling a context in vs. referencing it. <code>@import</code> semantics that allow potential modification makes more sense to me<br/>
Dave Longley: using array is “process these contexts in this order”, while <code>@imports</code> allows re-use and modification of existing contexts<br/>
Benjamin Young: {"<code>@context":</code> {"<code>@import":</code> "<a href="http://...anno.json",">http://...anno.json",</a> "name": "<a href="https://schema.org/name"}}">https://schema.org/name"}}</a> &lt;– not a thing? guessing we should clarify the new limitations on <code>@context</code> in general…<br/>
Benjamin Young: this substantively changes what is in <code>@context</code><br/>
Dave Longley: ivan brought up issue w/ <code>@protected</code> where people wanted to override schema.org context elements. If terms had been protected, override would fail…<br/>
… <code>@import</code> would allow changes before it gets defined.<br/>
Harold Solbrig: <code>@bigbluehat:</code> how does <code>@import</code> jive with verifiable credentials, etc.<br/>
Dave Longley: yes - this should not run afowl of the rules, as it would allow tweaking. What you can do is add on to array and pull in existing contexts and make them compatible with core contexts defined in specs<br/>
Ivan Herman: We need this story. I would like to see it written down and spelled out.<br/>
Dave Longley: Can do this, but can’t commit to timing<br/>
Dave Longley: “update your context before it is processed … as if the term definitions were always that way”<br/>
Benjamin Young: schema.org may change on us in the future, but maybe text –> iri change would make a good example. Does not mean that google will understand what you’ve done…<br/>
Ivan Herman: schema.org may not be a good idea. foaf?<br/>
Gregg Kellogg: I can create an example of modifying a term. <code>@protected</code> may require more work – another reason that <code>@imports</code> works better vs. <code>@source</code><br/>
… <code>@source</code> and (possibly) <code>@propagates</code> w/ nothing else (except version) allowed?<br/>
… if you pull and modify a context, you are <code>@importing</code> it but <code>@source</code> wouldn’t allow mods.<br/>
… but question is whether SRI could apply to imports or …<br/>
Ivan Herman: My understanding is that SRI refers to the context I identify w/ a URL, whether used in import or source isn’t a big problem<br/>
… rob’s original proposal seemed to be simpler – we don’t know whether he would support this or not.<br/>
Gregg Kellogg: will work on syntax document and changes to PR<br/>
Dave Longley: I’m thinking this shouldn’t be used for embedded contexts, in either <code>@source</code> or <code>@imports</code> situation<br/>
Benjamin Young: {"<code>@context":</code> {"<code>@import":</code> "<a href="http://...anno.json",">http://...anno.json",</a> "name": "<a href="https://schema.org/name"}}">https://schema.org/name"}}</a> <– not a thing?<br/>
Benjamin Young: The above should not be allowed?<br/>
… Leave github issues as they are…<br/>
</details>

gkellogg:
@dlongley the text still uses uses the previous context for expanding and compacting values of `@type`, regardless of propagation.

Could you suggest a test or two using `@propagate` combinations on property-scoped and type-scoped contexts that would ensure we're handling this properly?

rubensworks:
The new expansion tests also look relevant for the toRdf tests, should they be added in this PR as well?

gkellogg:
@rubensworks  Nominally, all expansion tests are useful for toRdf, but we haven't continued to add them all. The spec says that toRdf depends on expansion, so there is something implicit in this, but we have definitely done this before, so if there are tests you'd like to see in toRdf, I'm fine with that.

rubensworks:
@gkellogg Some processors may not strictly follow the specified algorithm (such as my own parser), and therefore may not depend on, or support explicit expansion. So in this case explicit toRdf test would definitely be good to have.
If that's ok with you, I'd be happy go over all the current expansion tests, and port those for toRdf when applicable.

gkellogg:
@rubensworks that would be great, maybe wait until the `@import`/`@propagate` PR is complete.

PR #114, opened 2019-06-29
Update test README

### LINKED ISSUE #113: Tests page no longer links clearly to tests
A friend and potential implementor sent me a message asking, where are the json-ld tests?  Indeed, I went looking for the old page but couldn't find it.  Instead I found https://w3c.github.io/json-ld-api/tests/ but I couldn't find the actual tests on there anymore.

The answer seems to be https://github.com/w3c/json-ld-api/tree/master/tests so I suggest linking there from that page explicitly.

gkellogg:
…  with specific reference to top-level manifest and individual algorithm manifests.

Fixes #113.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -9,23 +9,23 @@ More information and an RDFS definition of the test vocabulary can be found at [
 
 # Design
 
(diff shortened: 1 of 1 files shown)


gkellogg:
cc/  @cwebber.

PR #115, opened 2019-07-02
Respec issues

gkellogg:
Hopefully, this will quiet ReSpec errors and publish all definitions for export.

cc/ marcoscaceres

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/115.html" title="Last updated on Jul 3, 2019, 7:16 PM UTC (44b57bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/115/cf59ac4...44b57bd.html" title="Last updated on Jul 3, 2019, 7:16 PM UTC (44b57bd)">Diff</a>


common/common.js
```diff
@@ -37,6 +37,20 @@ function restrictReferences(utils, content) {
   return (base.innerHTML);
 }
 
+// Mark definitions to be exported
+function exportReferences(utils, content) {
+  const base = document.createElement("div:not([data-cite])");
+  base.innerHTML = content;
+
+  const defns = base.querySelectorAll("dfn");
+  for (const item of defns) {
```
marcoscaceres: I suggest not  doing this... ReSpec  automatically exports the matching WebIDL definitions for you, so you don't need to worry about them. 

For the  rest, exports should  be very deliberate: only export things that are actually referenced by  other specs.

marcoscaceres:
@gkellogg, just shipped `data-no-xref` in respec@24.25.0. Hopefully it helps!

common/common.js
gkellogg: Then, I need something for the other specs to not export the from common files, or does this happen because the first spec to get into the database prevents the others from being externalized?

gkellogg:
@pchampin if you’re good with it, we need to publish something based on these changes to get ReSpec to handle definitions from this, and other specs, properly.

common/common.js
marcoscaceres: So, ideally, the other specs won’t need to include the common definitions anymore as the definitions will be linked to the main spec (unless you really want them, in which case the imported duplicates need to be marked with “data-no-export” to avoid redefinitions).

gkellogg: We do want the definitions in the other specs, so I’ll need to add data-cite to the other spec versions to reference them from the API doc which defines them, I suppose.

marcoscaceres: ok, that's fine. My personal preference is not to do that (i.e., just link, don't duplicate things across specs - like other web platform specs that build on each other) - but if duplicating definitions across specs is  convention in the JSON-LD community, that's ok. 

The core thing is that we have only one source of truth for the definitions, and to make sure we only export what is really needed by other specs.

gkellogg: Re-reading, your first comment, ReSpec automatically exports WebIDL definitions, but this function is used in for local term bibliographies for terms shared between the different specs, so I think it should remain so that they get exported.

The other specs will use `data-cite` to reference, so the sections will look the same, but they won't actually make the definitions and references will redirect to the API document.

I thought this was the point of adding `xref` to the config, so that the namespace for definitions was made common, and I wouldn't need to do these things manually, but I think I don't really understand the point of xref and why some definitions would go into your database and others wouldn't.

In my mind, the `xref` config would give a set of common specifications so that definitions created in any would use one of them (the first, I suppose) as the base for the definition and ReSpec would take care of anything else. I don't really care that definitions go into a database, just that URLs that are generated are unique.

gkellogg:
@azaroth42 @pchampin @iherman @BigBlueHat My plan is to merge this PR and publish as a new WD, which will allow exported definitions to be used by the syntax and framing docs and not create duplicate definitions. It will also fix some long-standing ReSpec errors.

azaroth42:
:+1:

PR #125, opened 2019-07-22
Term updates

gkellogg:
for w3c/json-ld-syntax#185.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/125.html" title="Last updated on Jul 22, 2019, 5:13 PM UTC (780f00b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/125/91311ff...780f00b.html" title="Last updated on Jul 22, 2019, 5:13 PM UTC (780f00b)">Diff</a>

### DIFF
--- a/common/algorithm-terms.html
+++ b/common/algorithm-terms.html
@@ -1,48 +1,48 @@
 <dl class="termlist" data-sort>
-  <dt><dfn>active graph</dfn></dt><dd>
+  <dt><dfn data-cite="JSON-LD11-API#dfn-active-graph">active graph</dfn></dt><dd>
     The name of the currently active graph that the processor should use when processing.</dd>
-  <dt><dfn>active object</dfn></dt><dd>
+  <dt><dfn data-cite="JSON-LD11-API#dfn-active-object">active object</dfn></dt><dd>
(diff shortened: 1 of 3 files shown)


pchampin:
Things are much clearer with this changes, thanks.

However, the definitions of some of the "borrowed" terms include some JSON-LD specific information. So for those, I am not sure that linking *directly* to the original definition is such a good idea after all...

What about something like:

    array (see _original definition_)

      In the JSON serialization...

and have the text in the specification still point to the "local" definition?

Here is the list of terms in whose definition I have spotted some JSON-LD specifc informations:

- array
- JSON object
- null
- number
- relative IRI
- base IRI
- blank node
- literal

gkellogg:
@marcoscaceres objected to our defining definitions for things which were not truly JSON-LD terms, and _array_ is not core to JSON-LD. The closest concept was _list_, but that _is_ a core JSON-LD (or at least RDF) concept which is in conflict. The best solution seemed to be to create an _array_ definition which is an alias for _list_.

I believe your other suggestions are handled in #130.

PR #129, opened 2019-08-07
verify that compact IRIs and terms as `@vocab` expand properly

pchampin:
Following my comment in w3c/json-ld-syntax#184, @gkellogg updated all specs to make it clear that compact IRIs and terms are allowed as `@vocab`.

However, the playground and distiller do not seem to support it at the moment, and I could not find any specifc test for this in the test suite.

Please check that the proposed tests correspond to the intended behaviour. Also, I'm not sure how compaction should work in this situation, but probably a symetrical compaction test should be added?


pchampin:
Ok, thanks for the clarification, that makes more sense. I still think the tests have some value (if only to document a working case of @vocab using a compact IRI). Plus, I notice that the playground does not give the expected result, so... :-)

PR #133, opened 2019-08-12
Add support for the `alternate` link relation

gkellogg:
… with type `application/ld+json` to be used as a redirect if the retrieved document is not JSON.

For w3c/json-ld-syntax#204.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/133.html" title="Last updated on Aug 14, 2019, 4:25 PM UTC (caecf69)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/133/a8094d9...caecf69.html" title="Last updated on Aug 14, 2019, 4:25 PM UTC (caecf69)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6379,7 +6379,7 @@
       a context. When this is set, vocabulary-relative IRIs, such as the
       <a>entries</a> of <a>node objects</a>, are expanded or compacted relative
       to the <a>base IRI</a> and the <a>vocabulary mapping</a> using string concatenation.</li>
-    <li>In the <a>LoadDocumentCallback</a>, if the retrieved content is not `application/ld+json`
+    <li>In the <a>LoadDocumentCallback</a>, if the retrieved content is not any JSON media type
       and there is a link header with `rel=alternate` and `type=application/ld+json`, redirect
       to that content.</li>
   </ul>
--- a/index.html
+++ b/index.html
@@ -5824,7 +5824,7 @@
         <li>Set <var>documentUrl</var> to the location of the retrieved resource
           considering redirections (exclusive of HTTP status <code>303</code> "See Other" redirects
           as discussed in [[?cooluris]]).</li>
-        <li>If the retrieved resource's <a>Content-Type</a> is not <code>application/ld+json</code>
+        <li>If the retrieved resource's <a>Content-Type</a> is not <code>application/json</code>
           nor any media type with a <code>+json</code> suffix as defined in [[RFC6839]],
           and the response has an HTTP Link Header [[RFC8288]] using the <code>alternate</code> link relation
           with type `application/ld+json`,
--- a/index.html
+++ b/index.html
@@ -5809,7 +5809,7 @@
       <ol class="changed">
         <li>Create a new {{Promise}} <var>promise</var> and return it.
           The following steps are then executed asynchronously.</li>
-        <li>Set <var>document</var> to the body retrieved from
+        <li id="LoadDocumentCallback-step-2">Set <var>document</var> to the body retrieved from
           the resource identified by <a data-link-for="LoadDocumentCallback">url</a>,
           or by otherwise locating a resource associated with <a data-link-for="LoadDocumentCallback">url</a>.
           When requesting remote documents the request MUST prefer <a>Content-Type</a> <code>application/ld+json</code>
@@ -5824,6 +5824,13 @@
         <li>Set <var>documentUrl</var> to the location of the retrieved resource
           considering redirections (exclusive of HTTP status <code>303</code> "See Other" redirects
           as discussed in [[?cooluris]]).</li>
+        <li>If the retrieved resource's <a>Content-Type</a> is not <code>application/ld+json</code>
+          nor any media type with a <code>+json</code> suffix as defined in [[RFC6839]],
+          and the response has an HTTP Link Header [[RFC8259]] using the <code>alternate</code> link relation
+          with type `application/ld+json`,
+          set <var>url</var> to the associated <code>href</code> relative to the previous <var>url</var>
+          and restart the algorithm from <a href="#LoadDocumentCallback-step-2">step 2</a>,
+          ensuring that <var>documentUrl</var> is set to the original <var>url</var>.</li>
         <li>If the retrieved resource's <a>Content-Type</a> is <code>application/json</code>
           or any media type with a <code>+json</code> suffix as defined in [[RFC6839]]
           except <code>application/ld+json</code>,
@@ -6372,6 +6379,9 @@
       a context. When this is set, vocabulary-relative IRIs, such as the
       <a>entries</a> of <a>node objects</a>, are expanded or compacted relative
       to the <a>base IRI</a> and the <a>vocabulary mapping</a> using string concatenation.</li>
+    <li>In the <a>LoadDocumentCallback</a>, if the retrieved content is not `application/ld+json`
+      and there is a link header with `rel=alternate` and `type=application/ld+json`, redirect
+      to that content.</li>
   </ul>
 </section>
 
--- a/index.html
+++ b/index.html
@@ -5826,7 +5826,7 @@
           as discussed in [[?cooluris]]).</li>
         <li>If the retrieved resource's <a>Content-Type</a> is not <code>application/ld+json</code>
           nor any media type with a <code>+json</code> suffix as defined in [[RFC6839]],
-          and the response has an HTTP Link Header [[RFC8259]] using the <code>alternate</code> link relation
+          and the response has an HTTP Link Header [[RFC8288]] using the <code>alternate</code> link relation
           with type `application/ld+json`,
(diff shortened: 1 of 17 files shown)


gkellogg:
Note that I wrote this so that the `documentUrl` remains that of the original document, not that that is redirected to, but I could be convinced otherwise.

index.html
```diff
@@ -5824,6 +5824,13 @@ <h3>LoadDocumentCallback</h3>
         <li>Set <var>documentUrl</var> to the location of the retrieved resource
           considering redirections (exclusive of HTTP status <code>303</code> "See Other" redirects
           as discussed in [[?cooluris]]).</li>
+        <li>If the retrieved resource's <a>Content-Type</a> is not <code>application/ld+json</code>
+          nor any media type with a <code>+json</code> suffix as defined in [[RFC6839]],
+          and the response has an HTTP Link Header [[RFC8259]] using the <code>alternate</code> link relation
```
pchampin: RFC8259 describe the JSON format, not the Link Header... Note that this error is present in other parts of the document (I imagine you just copy-pasted it from there).

index.html
```diff
@@ -6372,6 +6379,9 @@ <h2>Changes since JSON-LD Community Group Final Report</h2>
       a context. When this is set, vocabulary-relative IRIs, such as the
       <a>entries</a> of <a>node objects</a>, are expanded or compacted relative
       to the <a>base IRI</a> and the <a>vocabulary mapping</a> using string concatenation.</li>
+    <li>In the <a>LoadDocumentCallback</a>, if the retrieved content is not `application/ld+json`
+      and there is a link header with `rel=alternate` and `type=application/ld+json`, redirect
+      to that content.</li>
```
pchampin: This probably needs to be discussed (it is not just editorial).

This would create a different behaviour when the document is accessed directly, than when it is accessed through the "link header indirection". So I am not a big fan of the idea.

I would rather encourage the author of the JSON-LD document to explicitly set the base to the URL o the non-JSON-LD document, if that is the behaviour they intend.

index.html
gkellogg: Thanks, I’ll fix that. Also, this should limit application/json, not just ld+json and other +json types.

index.html
gkellogg: The PR is definitely prospective, as the issue is still marked to be discussed, but I wanted to get it out there and see the consequences. The link header shouldn’t affect any json format, but would affect html content. What document base should be applied can be debated, but using the original document has precedent.

dlongley: Hmm, I think I disagree with @pchampin ... and that we want to treat this like we do with redirects, which I believe uses the original document. The argument for doing what @pchampin suggests would be that there's no way to know ahead of time what the final document will be, so you couldn't set the base to it, whereas you could set the base to the original document if you wanted it to stick.

But, processors could have a flag to switch to this other behavior. It seems to me that the sensible default would be the original document. Especially given the way that we expect this feature to be used ... which is that the linked document is really a hidden implementation detail. If you wanted to use the linked document directly you would have specified it instead of some, e.g., HTML document.

index.html
BigBlueHat: Web Linking is [RFC8288](https://tools.ietf.org/html/rfc8288) (confusingly close looking 😉).

index.html
gkellogg: 👎 to introducing another flag, though. If the document should have some specific base different from the originally referenced resource, it can use `@base`.

My thought in using the original resource base, rather than the referenced document, is essentially the same reasoning that lead to the semantics of HTTP status 303 for range-14 considerations.

BigBlueHat: If this use of `alternate` (which comes from [HTML primarily](https://html.spec.whatwg.org/multipage/links.html#rel-alternate)) is meant to supplant using content negotiation, then the expectation would be (afaict) that the original resource URL would (semantically) be the correct base URL--which I think agrees with what @gkellogg said above.

Once the `alternate` is retrieved it should experientially be *like* getting a content negotiated response with a `Content-Location` set to the `alternate` and having `Location` still be set to the original Request-URI--which is how content negotiation plays out.

pchampin: I agree that the intention is similar to content negotiation, and this advocates for keeping the base IRI of the original document.

However, it is not, as @dlongley puts it, a *hidden* implementation detail (as conneg would be) because the client is given the URL of the actual JSON-LD content. In practice, it is much more similar to a 303 redirect. And I am not aware of any precedent where, when following a 303 redirect or `rel=alternate`, the client is supposed to keep the base IRI of the source document... Any pointer?

gkellogg: This is the same mechanism used, e.g. by DBPedia to distinguish a resource from the page about the resource.  In thier case http://dbpedia.org/resource/Richard_III_of_England redirects to http://dbpedia.org/page/Richard_III_of_England so that statements can be made about the resource, rather than the page which describes the resource. (Fragment identifiers are much nicer).

See [Wikipedia article on HTTPRange-14](https://en.wikipedia.org/wiki/HTTPRange-14) and [Jeni's discussion for the TAG](http://www.jenitennison.com/2012/05/11/using-punning-to-answer-httprange-14.html).

I don't know of anything specific for `rel="alternate"` that relates to this, although @BigBlueHat brought something up on the call. IMO, it is reasonable for a format, such as JSON-LD, to ascribe such semantics.

pchampin: I understand the parallel with HTTPRange-14 . But I see nowhere in those documents an indication that relative IRIs in the target document should be resolved against the source document's IRI...

pchampin: To make my point clearer: https://perso.liris.cnrs.fr/pierre-antoine.champin/2019/redirect/ shows that browsers do not keep the source URL of a 303 redirect to interpret relative URLs in the target document (or at least, mine does not).

BigBlueHat: What I mentioned on the call Friday was that the only definition of `rel="alternate"` (that I'm aware of) is in https://html.spec.whatwg.org/multipage/links.html#rel-alternate

The summary definition is "Gives alternate representations of the current document." but the follow-on expansion on it includes "The meaning of this keyword depends on the values of the other attributes."

The examples in that section include [alternate CSS styles](https://html.spec.whatwg.org/multipage/links.html#rel-alternate:rel-alternate-3) (i.e. `rel="alternate stylesheet"`), Atom/RSS alternate representations (i.e. `rel="alternate" type=""`), and using it for language [content] negotiation (i.e. `rel="alternate" hreflang=""` +/- `type=""`).

So, using `alternate` here to accomplish a pre-flight content negotiation (of sorts) is not without its precedence conceptually. However, none of that directly answers the base URI situation...as in the closest experiential case (language negotiation) the browser would navigate to the new resource based on language preferences--thereby acting more like a `30*` redirect than like content negotiation.

BigBlueHat: Additionally, if we stick with `alternate`, we may need to require other link parameters (or type parameters) to clarify that this is a representation of the current document *as a context* vs. a representation of the current document *expressed in JSON-LD* (i.e. as a data document).

We could use the existing context profile URL (which is used as a `rel` in [Interpreting JSON as JSON-LD](https://w3c.github.io/json-ld-syntax/#interpreting-json-as-json-ld)). The experience would be similar to referencing a data document as the context of another data document.

gkellogg:
> the current document _as a context_ vs. a representation of the current document expressed in JSON-LD (i.e. as a data document).

In fact, this is not simply for contexts, but for any non-JSON content where we want to use a JSON-LD representation.

I didn’t think that anything other than the `rel` and `type` fields would be necessary.

BigBlueHat:
> In fact, this is not simply for contexts, but for any non-JSON content where we want to use a JSON-LD representation.

Gotcha, so if I shipped a data document describing `http://bigbluehat.com/` which also included an inline context and then referenced that URL as the `@context` value in my other data, JSON-LD processors would know the difference in usage based on...context? 😜 Makes sense, but maybe we should explain that, so it's clear we're not overloading the meaning of `alternate` in some way.

gkellogg:
A context is just JSON-LD including `@context` which has meaning because of where it’s dereferenced from. The document loader is used to retrieve resourses, both as primary use (expand, compact, etc), for frames, and contexts.

The document loader will follow a link, if the retrieved resource is not JSON, for any of these cases. The use for context retrieval is incidental.

index.html
gkellogg: > I understand the parallel with HTTPRange-14 . But I see nowhere in those documents an indication that relative IRIs in the target document should be resolved against the source document's IRI...

There's a discussion in [Cool URIs](https://www.w3.org/TR/cooluris/#r303uri):

> ... Each of the RDF documents would contain statements about the appropriate resource, using the original URI, e.g. http://www.example.com/id/alice, to identify the described resource.

pchampin: Yes, but they do not say that a relative URI could or should be used for that...

gkellogg: At this point, I think you need to raise a new issue for discussion, because we seem to have quite different views on this, and discussing in a merged PR isn't a good way to be sure we resolve it.

PR #135, opened 2019-08-19
Encapsulate HTML processing

gkellogg:
Move HTML bits into a separate section, with callouts into the documentLoader algorithm to contain HTML-related processing. Also, removes processor levels describing this as a processor supporting "HTML script extraction".

For w3c/json-ld-syntax#213.

### DIFF
--- a/index.html
+++ b/index.html
@@ -940,52 +940,6 @@
       </tr>
     </tbody>
   </table>
-
-  <section>
-    <h2>Processor Levels</h2>
-
-    <!--
-      **************** IMPORTANT WARNING ****************
-      This section is duplicated (with a few adaptations)
-      in other JSON-LD specifications. Since these
-      definitions are normative, it is important to
-      reflect any change in the other documents.
-      ***************************************************
-    -->
-
-    <p>JSON-LD mostly uses the JSON syntax [[RFC8259]] along with
-      various micro-syntaxes based on XML Schema datatypes [[XMLSCHEMA11-2]].
-      However, it has become increasingly common to include JSON within
-      a <a data-cite="HTML/scripting.html#the-script-element">script element</a>
-      within an HTML document [[HTML]],
-      as described in <a href="#html-content-algorithms" class="sectionRef"></a>.
-      As not all processors operate in an environment which can include HTML,
-      this specification describes various categories of JSON-LD processors.</p>
-
-    <p>A <dfn data-cite="JSON-LD11#pure-json-processor">pure JSON Processor</dfn> only requires the use of a
-      JSON processor and is restricted to processing documents retrieved
-      with a JSON content type (e.g., <code>application/ld+json</code> or other JSON type).</p>
-
-    <p>A <dfn data-cite="JSON-LD11#full-processor">full Processor</dfn> is capable of processing JSON-LD embedded in HTML,
-      in addition to the capabilities of a <a>pure JSON Processor</a>.</p>
-
-    <section class="informative">
-      <h3>Additional Processor Levels</h3>
-
-      <p>In addition to the normatively defined processor levels, an additional processor
-        level is defined for reference.</p>
-
-      <p>A <dfn data-cite="JSON-LD11#event-based-json-processor">event-based JSON Processor</dfn> processes a stream of characters
-        expecting an event after each syntactic element is encountered.
-        Such processors are sensitive to the order of the members of <a>JSON objects</a>,
-        which can have a performance impact if the members of <a>JSON objects</a> are encountered in an unexpected order.
-        An <a>event-based JSON Processor</a> may process JSON-LD embedded in HTML.</p>
-
-      <p class="note">An <a>event-based JSON Processor</a>
-        may be sensitive to processing certain keywords in order, including
-        <code>@context</code>, <code>@id</code>, and <code>@type</code>.</p>
-    </section>
-  </section>
 </section> <!-- end of Conformance section -->
 
 <section>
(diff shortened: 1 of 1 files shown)


BigBlueHat:
> a feature that the WG considers as a "should have" in the final spec, but not sure whether that specific issue will pass the CR phase. By explicitly marking a feature to be at risk, the feature can be removed from the final spec if there aren't enough implementation, without the requirement to re-publish a new CR and delaying the Proposed Rec phase.

I think we need to keep this marked as "at risk" for exactly these reasons. The syntax spec [suggests avoiding using `<base>`](https://pr-preview.s3.amazonaws.com/w3c/json-ld-syntax/214/f1f4efe...d1250fa.html#example-140-using-the-document-base-url-to-establish-the-default-base-iri) due to it's dynamic nature. Current extractors (SEO bots and Structured Data Testing Tool) don't currently use `<base>` in relationship to data blocks and the SEO community generally recommend against relative URLs in general.

Given that this is, then, a new feature addition (beyond how JSON-LD in HTML currently works), and that supporting it may introduce additional requirements (DOM and/or JS processing) on JSON-LD API implementers, leaving it marked as "at risk" seems prudent until we know from implementers how they plan to handle this proposed feature.

BigBlueHat:
@gkellogg would love to see a "data block" mention added with reference to the [HTML spec which defines data blocks](https://html.spec.whatwg.org/multipage/scripting.html#data-block).

iherman:
@BigBlueHat maybe. But the note in the document says:

> (FEATURE AT RISK) ISSUE
> The use of the Document Base URL from [HTML] for setting the base IRI of the enclosed JSON-LD is an experimental feature, which may be changed in a future version of this specification.

The issue is not whether it will be changed in a future version or not; the issue is whether it will be kept in the current version or not. I.e., the note's text is suggesting something different. We have to decide what we want to say here...

BigBlueHat:
@iherman maybe I'm reading that note differently...or don't understand the distinction between "current" and "future" versions.

If it's "at risk" when we go to CR and still lacks implementation, then we can take it out easily (without having to go through more process). If we remove the "at risk" label now, then it's a heavier lift (i.e. triggers more process) if we find it's unimplemented or causes other issues/concerns.

I'd prefer we keep the "at risk" status until we have more implementation evidence.

iherman:
If we want to do what you want to do:-), then the text should say something like

> (FEATURE AT RISK) ISSUE
> The use of the Document Base URL from [HTML] for setting the base IRI of the enclosed JSON-LD is an experimental feature. If there are not enough implementation experience during the Candidate Recommendation phase, this feature may be removed from the final specification.

Of course, future version may do what they want, but the user should _not_ count on this feature in this version of the specification.

The current text may be understood as saying: "There will be a standard feature in the present recommendation, but that feature may change in future revisions". This is different...

BigBlueHat:
@iherman that makes sense. Guess @gkellogg should share thoughts since he wrote the original text. 😃

gkellogg:
We’ll just leave it as an issue an remove the “atrisk” part. That said, we should figure out how to close out such open issues in the specs.

index.html
```diff
@@ -6003,6 +5861,113 @@ <h3>RemoteDocument</h3>
...
+            </li>
+          </ol>
+        </li>
+      </ol>
+    </section>
+
+    <section id="extract-script-content" class="algorithm">
+      <h3>Extract Script Content Algorithm</h3>
+
+      <p>The algorithm extracts the text content a
```
```suggestion
      <p>The algorithm extracts the text content of a
```

index.html
```diff
@@ -6003,6 +5861,113 @@ <h3>RemoteDocument</h3>
     </section>
   </section> <!-- end of Remote Document and Context Retrieval -->
 
+  <section class="changed">
+    <h2>HTML Content Algorithms</h2>
+    <p class="note">This section describes features available
```
```suggestion
    <p class="note">This section describes optional features available
```
This is redundant with the "MAY" below, but just to be on the safe side?

index.html
```diff
@@ -6003,6 +5861,113 @@ <h3>RemoteDocument</h3>
...
+      <p>This sections describe an extension to the algorithm specified
+        in <a>LoadDocumentCallback</a> to support extracting JSON-LD from HTML.</p>
+
+      <p><a href="#LoadDocumentCallback-step-2">Step 2</a> is updated to add the following: A processor supporting <a>HTML script extraction</a> MUST include <code>text/html</code> at any preference level,
+        unless <a data-link-for="LoadDocumentOptions">requestProfile</a> is `http://www.w3.org/ns/json-ld#context`.</p>
+
+      <p>After <a href="#LoadDocumentCallback-step-5">step 5</a>, add the following processing step:
+        Otherwise, if the retrieved resource's <a>Content-Type</a> is <code>text/html</code>:</p>
+      <ol>
+        <li>If the processor does not support <a>HTML script extraction</a>
```
pchampin: It is a bit strange to have this test. As I understand this section, the algorithm should be so updated IF the document loader supports HTML script extraction. So the test would be trivially true...

gkellogg:
Regarding the (off-topic) atrisk section, in https://github.com/w3c/json-ld-api/pull/136, I simply removed it. It is discussed on the syntax doc.

index.html
gkellogg: Yes, that step is not necessary.

gkellogg:
> @gkellogg I'd love to see a "data block" mention added with reference to the [HTML spec which defines data blocks](https://html.spec.whatwg.org/multipage/scripting.html#data-block).

Would this go better in https://github.com/w3c/json-ld-syntax/pull/214? Is it more an API or a Syntax concern? Maybe suggest some wording.

tests/manifest.jsonld
```diff
@@ -10,6 +10,7 @@
     "flatten-manifest.jsonld",
     "fromRdf-manifest.jsonld",
     "remote-doc-manifest.jsonld",
-    "toRdf-manifest.jsonld"
+    "toRdf-manifest.jsonld",
+    "html-manifest.jsonld"
```
rubensworks: This new manifest should probably also be mentioned in the _Design_ section of`tests/README.md`.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-08-23-json-ld#section4-1).

- `RESOLVED:  gkellogg to merge #135 and #214 after reviewers have approved and close the relevant issues`
ncapsulate HTML processing<br/>
Rob Sanderson: See Syntax #214<br/>
Rob Sanderson: See API #135<br/>
Rob Sanderson: Discussion from last week has resulted in some PRs.<br/>
Ivan Herman: Gregg not here this week.<br/>
Pierre-Antoine Champin: dlongley: PRs are moving in a direction I would agree with<br/>
Rob Sanderson: I would agree as well, pushing things into the document loader as discussed last week.<br/>
… I guess the issue to discuss is – is there anyone who is not comfortable yet otherwise we should accept those PRs.<br/>
Pierre-Antoine Champin: scribeassist: pchampin<br/>
Rob Sanderson: Any objections to the approach?<br/>
Ivan Herman: I read through the documents and we have done the work.<br/>
Pierre-Antoine Champin: dlongley: I would like to wait for other reviews before minerging (including mine)<br/>
Proposed resolution: gkellogg to merge #135 and #214 after reviewers have approved (Rob Sanderson)<br/>
Proposed resolution: gkellogg to merge #135 and #214 after reviewers have approved and close the relevant issues (Rob Sanderson)<br/>
Dave Longley: +1<br/>
Ruben Taelman: +1<br/>
Rob Sanderson: +1<br/>
Benjamin Young: +1<br/>
Ivan Herman: +1<br/>
David I. Lehn: +1<br/>
Resolution #2: gkellogg to merge #135 and #214 after reviewers have approved and close the relevant issues<br/>
</details>

PR #137, opened 2019-08-21
atrisk

gkellogg:
* Remove atrisk on JSON Canonicallization.
* Remove atrisk div on the use of HTML base. This was resolved in w3c/json-ld-syntax#134 (comment) which is reflected in the syntax document.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4628,7 +4628,7 @@
           convert <var>value</var> to the <a>canonical lexical form</a>
           using the result of transforming the <a>internal representation</a> of <var>value</var>
           to JSON and set <var>datatype</var> to <code>rdf:JSON</code>.
-          <div class="issue atrisk">The JSON Canonicalization Scheme [[?JCS]]
+          <div class="issue">The JSON Canonicalization Scheme [[?JCS]]
             is an emerging standard for JSON canonicalization
             not yet ready to be referenced.
             When a JSON canonicalization standard becomes available,
@@ -5127,7 +5127,7 @@
         which SHOULD be serialized as <code>\\</code> and <code>\"</code> respectively.</li>
     </ul>
 
-    <p class="issue atrisk changed">The JSON Canonicalization Scheme [[?JCS]]
+    <p class="issue changed">The JSON Canonicalization Scheme [[?JCS]]
       is an emerging standard for JSON canonicalization
       not yet ready to be referenced.
       When a JSON canonicalization standard becomes available,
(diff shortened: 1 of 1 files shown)


iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-08-23-json-ld#section4-3).

- `RESOLVED:  approve PR #137`
Removing “at risk” declarations on inline issues and &lt;base> handling<br/>
Benjamin Young: See API #137<br/>
Benjamin Young: The at-riskness has been removed where it didn’t match the w3c semantics of “at risk” … and saying these are open issues instead.<br/>
… By removing at-risk for these other features, these will be in CR without question.<br/>
… My only hesitation is around the base tag one – all the vagueness around base calculation is a little dodgy and it’s a new feature. We don’t yet have multiple implementers of this.<br/>
… The use of base as we have it now does map to RFC3986. That it goes through the mechanics of any wrapping documents which is the HTML base tag at this point, but it does presume that you are operating within HTML and not pulling things out and passing it off to another processor.<br/>
… So I think this is a fairly significant shift but it is only setting the base default.<br/>
Adam Soroka: I’m comfortable with this change. To follow on – it’s oddly similar to what we just discussed, how much context does a processor have to consider about where the content came from.<br/>
… I don’t think we can say that no processor is in a vacuum but at the same time we don’t want people to have to drag the whole Web after them.<br/>
… And we want to do something reasonable to meet the use cases, etc. I think you’re talking about a piece of data to resolve URIs and we’re hedging/moving in that direction and you have to know a little more than what’s directly being processed in that moment.<br/>
Ivan Herman: The usage of the base, and how to use it is an issue we’ve discussed and we’ve closed it.<br/>
… The only question is whether it creates implementation issues – I don’t think it does because in practice, if you use the HTML parser, getting base is a piece of cake. Now we have a separation with document loader doing HTML or not, and that’s all clear.<br/>
Pierre-Antoine Champin: dlongley: Ivan covered most of what I was about to say.<br/>
Pierre-Antoine Champin: … If you have an HTML parser, it is easy to get the base at the time you are doing the processing.<br/>
Benjamin Young: The main reason I wanted to leave this open was to get feedback from Dave – and that works for me.<br/>
Adam Soroka: I said a real question about conformance – in a situation where we’re headed, with at least two flavors of document loaders. Do we need to see two different implementations of that more powerful document loader?<br/>
Ivan Herman: Yes.<br/>
Benjamin Young: A quick note to highlight about “when the processing happens” that becomes a really important aspect for authors. Base can change, JSON-LD can be injected, so on – it’s a timing question. All bets are off on a lot of Web apps on what you’re actually encoding based on timing.<br/>
Proposed resolution: approve PR #137 (Benjamin Young)<br/>
Benjamin Young: +1<br/>
Dave Longley: +1<br/>
Ivan Herman: +1<br/>
Adam Soroka: +1<br/>
David I. Lehn: +1<br/>
Ruben Taelman: +1<br/>
Benjamin Young: +1<br/>
Resolution #4: approve PR #137<br/>
</details>

PR #152, opened 2019-09-18
Mark tests that use relative `@vocab` as JSON-LD 1.1 tests

rubensworks:
This indicates that 1.0 processors should not run these tests, as relative vocabs are a 1.1 feature.


gkellogg:
If these were not caught, it may mean that we don’t have checks for relative `@vocab` when the processing mode is 1.0.

rubensworks:
So in the case of toRdf, you mean we should have 1.0 tests that explicitly test that relative `@vocab`'s result in omitted triples?
If so, do you want me to add those to this PR?

gkellogg:
Sure, that would be great. We should probably avoid using the processinMode option, if possible, and rely on `@version`. There may need to be some spec text about validating the value of `@vocab`, but not sure we can add something new when in 1.0 mode.

rubensworks:
I've added two tests for both expand and toRdf (115 is based on 92, 116 is based on 110).

> We should probably avoid using the processinMode option, if possible, and rely on `@version`.

I'm not sure what you meant with relying on `@version`. AFAIK, `@version` is only allowed to contain `1.1` as value, but not `1.0`.
On the new tests, I've now set `specVersion` to `json-ld-1.0`, without setting `processingMode`.

> There may need to be some spec text about validating the value of `@vocab`, but not sure we can add something new when in 1.0 mode.

Should this also be included in this PR?
If so, could you point me to the appropriate place in the document where this could be added? (https://w3c.github.io/json-ld-api/#iri-expansion ?)

-----

For reference, the JSON-LD playground ([example](https://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40context%22%3A%7B%22%40base%22%3A%22http%3A%2F%2Fexample.com%2Fsome%2Fdeep%2Fdirectory%2Fand%2Ffile%2F%22%2C%22%40vocab%22%3A%22%2Frelative%22%7D%2C%22%40id%22%3A%22relativePropertyIris%22%2C%22link%22%3A%22link%22%2C%22%23fragment-works%22%3A%22%23fragment-works%22%2C%22%3Fquery%3Dworks%22%3A%22%3Fquery%3Dworks%22%2C%22.%2F%22%3A%22.%2F%22%2C%22..%2F%22%3A%22..%2F%22%2C%22..%2Fparent%22%3A%22..%2Fparent%22%2C%22..%2F..%2Fparent-parent-eq-root%22%3A%22..%2F..%2Fparent-parent-eq-root%22%2C%22..%2F..%2F..%2F..%2F..%2Fstill-root%22%3A%22..%2F..%2F..%2F..%2F..%2Fstill-root%22%2C%22..%2F..%2F.%2F..%2F.%2F..%2F..%2Ftoo-many-dots%22%3A%22..%2F..%2F.%2F..%2F.%2F..%2F..%2Ftoo-many-dots%22%2C%22%2Fabsolute%22%3A%22%2Fabsolute%22%2C%22%2F%2Fexample.org%2Fscheme-relative%22%3A%22%2F%2Fexample.org%2Fscheme-relative%22%7D)) does not drop properties that use relative `@vocab` as required by the new tests. So I'm wondering now if dropping those is the proper solution, and if an error shouldn't be throw like jsonld.js does?

davidlehn:
@rubensworks Note that jsonld.js, and therefore the playground, is not yet passing some of the relative vocab tests.  Gregg's distiller is more up-to-date with those tests.

rubensworks:
@davidlehn Ah, that's good to know.
But in any case, the new tests (115,116) in this PR are for JSON-LD 1.0, so the behaviour for them should (in theory) be consistent across implementations. @gkellogg probably has a better view on the intended behaviour than me though.

gkellogg:
Need to look at bot 1.0 and 1.1 behavior of `@vocab` and get a better understanding of how relative IRIs are handled.

gkellogg:
1.0 syntax indeed requires that an IRI value of `@vocab` be an absolute IRI, but the API processing doesn't seem to do anything special to check for this. It's reasonable to have 1.0 tests to check for it, but I'm not sure what to do about failing tests. It could affect a processor in 1.0 mode, but we need to consider the effect of changing from explicit processingMode through `@vocab` to feature detection logic. (Of course, there's always explicit processingMode set to `json-ld-1.0` in the API).

gkellogg:
I'm going to say that a relative vocabulary IRI would be a detected feature triggering 1.1 mode, and if set to explicit 1.0 mode through the API, that the processor should detect the attempt to set a relative IRI when an absolute IRI is required and fail.

Prior to going to the feature-detection route, which involves more things, a processor SHOULD detect an error unless 1.1 mode is set explicitly (`@version` or API).

tests/expand-manifest.jsonld
```diff
@@ -848,6 +852,22 @@
       "input": "expand/0114-in.jsonld",
       "expect": "expand/0114-out.jsonld",
       "option": {"specVersion": "json-ld-1.1"}
+    }, {
+      "@id": "#t0115",
+      "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
+      "name": "Verifies that relative IRIs as properties with @vocab: '' in 1.0 are dropped",
+      "purpose": "Pathological relative property IRIs in 1.0",
```
gkellogg: In 1.0 mode, this should be an error, not simply dropped/ignored, as the syntax says they MUST be absolute IRIs. It should be a NegativeEvaluationTest and generate an "invalid vocab mapping" error.

tests/expand-manifest.jsonld
```diff
@@ -848,6 +852,22 @@
...
+      "input": "expand/0115-in.jsonld",
+      "expect": "expand/0115-out.jsonld",
+      "option": {"specVersion": "json-ld-1.0"}
+    }, {
+      "@id": "#t0116",
+      "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
+      "name": "Verifies that relative IRIs as properties with relative @vocab in 1.0 are dropped",
+      "purpose": "Pathological relative property IRIs in 1.0",
```
gkellogg: Error

tests/expand-manifest.jsonld
```diff
@@ -818,21 +819,24 @@
...
+      "option": {"processingMode": "json-ld-1.1", "specVersion": "json-ld-1.1"}
     }, {
       "@id": "#t0111",
       "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
       "name": "Various relative IRIs as properties with with relative @vocab itself relative to an existing vocabulary base",
       "purpose": "Pathological relative property IRIs",
       "input": "expand/0111-in.jsonld",
-      "expect": "expand/0111-out.jsonld"
+      "expect": "expand/0111-out.jsonld",
+      "option": {"processingMode": "json-ld-1.1", "specVersion": "json-ld-1.1"}
```
gkellogg: Why set processingMode in API rather than just use `@version`?

rubensworks:
Ok, so tests 115 and 116 should become `jld:NegativeEvaluationTest` then.
I don't immediately see an appropriate error code for this though, `"invalid IRI mapping"` could work, but may not be specific enough.
Shall I include the addition of something like `"invalid @vocab" value"` for the spec document in this PR?

tests/expand-manifest.jsonld
```diff
@@ -818,21 +819,24 @@
...
     }, {
       "@id": "#t0112",
       "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
       "name": "Various relative IRIs as properties with with relative @vocab relative to another relative vocabulary base",
       "purpose": "Pathological relative property IRIs",
       "input": "expand/0112-in.jsonld",
-      "expect": "expand/0112-out.jsonld"
+      "expect": "expand/0112-out.jsonld",
+      "option": {"processingMode": "json-ld-1.1", "specVersion": "json-ld-1.1"}
```
gkellogg: Why set processingMode in API rather than just use `@version`?

gkellogg:
> Ok, so tests 115 and 116 should become `jld:NegativeEvaluationTest` then.
> I don't immediately see an appropriate error code for this though, `"invalid IRI mapping"` could work, but may not be specific enough.
> Shall I include the addition of something like `"invalid @vocab" value"` for the spec document in this PR?

As indicate in other comments, "invalid vocab mapping" seems appropriate.

rubensworks:
I've updated the PR based on the comments.

Now that 115 and 116 are `jld:NegativeEvaluationTest`'s, I've also minimized their input.

PR #163, opened 2019-10-04
Fix IDL markup issues.

davidlehn:
This respec dfn support seems difficult to use.  Looking for comments if this patch is doing the right thing.
- The "iterable" properties didn't have a automatic `<code>` around them when using the dfn stuff.  Not sure why?  Other properties worked.  Added that by hand and made the RdfGraph version linkable.
- That `graphName` word looked out of place, so tried to link it back to the func arg, but couldn't figure out how other than with a hand written href.  That is probably wrong.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/163.html" title="Last updated on Oct 10, 2019, 9:55 PM UTC (f35b9e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/163/9e19e9f...f35b9e6.html" title="Last updated on Oct 10, 2019, 9:55 PM UTC (f35b9e6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5471,6 +5471,9 @@
     <a data-cite="ECMASCRIPT#sec-promise-objects">Promises</a> are defined in [[ECMASCRIPT]].
     General use within specifications can be found in [[promises-guide]].</p>
 
+  <p class="note">Interfaces are marked `[Exposed=(Window,Worker)]`,
+    but the API is also intended for use outside of a browser context.</p>
+
   <section>
     <h3>The <dfn>JsonLdProcessor</dfn> Interface</h3>
 
@@ -5493,7 +5496,7 @@
       in this document mention this directly.</p>
 
     <pre class="idl">
-      [Exposed=Window]
+      [Exposed=(Window,Worker)]
       interface JsonLdProcessor {
         constructor();
         static Promise&lt;JsonLdDictionary> compact(
@@ -5787,7 +5790,7 @@
       which has a <a>default graph</a> accessible via the <a data-link-for="RdfDataset">defaultGraph</a> attribute.</p>
 
     <pre class="idl changed">
-      [Exposed=Window]
+      [Exposed=(Window,Worker)]
       interface RdfDataset {
         constructor();
         readonly attribute RdfGraph defaultGraph;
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4770,7 +4770,7 @@ <h2>Algorithm</h2>
         that all default to <code>false</code>.</p>
 
       <p>The <var>dataset</var> is <a data-link-for="RdfDataset">iterable</a> to iterate over <a>graphs</a> and <a>graph names</a>
-        contained within the <a>RdfDataset</a>. Each <a>graph</a> is also <code>iterable</code>
+        contained within the <a>RdfDataset</a>. Each <a>graph</a> is also <a data-link-for="RdfGraph">iterable</a>
```
gkellogg:
```suggestion
        contained within the <a>RdfDataset</a>. Each <a>graph</a> is also `<a data-link-for="RdfGraph">iterable</a>`
```

gkellogg: Grave is the same as `<code></code>`. Could go inside the link as well.

davidlehn:
> > Looking for comments if this patch is doing the right thing.
> > 
> > * The "iterable" properties didn't have a automatic `<code>` around them when using the dfn stuff.  Not sure why?  Other properties worked.  Added that by hand and made the RdfGraph version linkable.
> 
> ReSpec doesn't provide a good means of documenting "iterable", and there was a fair amount of back and forth on the ReSpec repo. They don't treat it as a documentable interface. We could perhaps do something directly in markup to remedy.
> 

It seems I don't know how this works.  These specific API names are whitelisted somewhere else?  I just assumed it should style whatever we use properly.  Adding `<code>` directly makes it look right.


> > * That `graphName` word looked out of place, so tried to link it back to the func arg, but couldn't figure out how other than with a hand written href.  That is probably wrong.
> 
> Can you be more specific? It is linked from 1.2 of the Deserialize JSON-LD to RDF algorithm.

The second sentence of the definition of `graphName` itself starts with an unstyled "graphName".  It looked out of place and grammer checker assumed it should be capitalized.  I thought it would be easy to just link back to itself, and that turned into a challenging task.

gkellogg:
> It seems I don't know how this works. These specific API names are whitelisted somewhere else? I just assumed it should style whatever we use properly. Adding <code> directly makes it look right.

`iterable` isn't a method, but a property of an interface, so it's treated differently. Note that it doesn't return a value.

PR #164, opened 2019-10-05
Add tests for checking lowercasing of @language

rubensworks:
I recently discovered that my implementation did not ensure lowercased languages, while this has been [required](https://w3c.github.io/json-ld-syntax/#dfn-language-tagged-string) since JSON-LD 1.0.
Apparently there weren't any test cases for this in the test suite, so this PR adds one (for expand and toRdf) for all the different language features.

### DIFF
--- a/tests/expand-manifest.jsonld
+++ b/tests/expand-manifest.jsonld
@@ -869,6 +869,13 @@
       "expect": "invalid vocab mapping",
       "option": {"specVersion": "json-ld-1.0"}
     }, {
+      "@id": "#t0117",
+      "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
+      "name": "Language tags with mixed casing",
+      "purpose": "Language tags must be normalized to lowercase",
+      "input": "expand/0117-in.jsonld",
+      "expect": "expand/0117-out.jsonld"
+    }, {
       "@id": "#tc001",
       "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
       "name": "adding new term",
--- a/tests/expand/0117-in.jsonld
+++ b/tests/expand/0117-in.jsonld
@@ -1,0 +1,25 @@
+{
+  "@context": {
+    "property": { "@id": "http://example.com/vocab/property", "@language": "DE" },
+    "indexMap": { "@id": "http://example.com/vocab/indexMap", "@language": "eN", "@container": "@index" },
+    "property2": { "@id": "http://example.com/vocab/property2", "@container": "@language" }
+  },
+  "@id": "http://example.com/node",
+  "property": [
+    {
+      "@id": "http://example.com/propertyValueNode",
+      "indexMap": {
+        "expands to english string": "simple string"
+      }
+    },
+    "einfacher String",
+	{
+	  "@value": "woord",
+	  "@language": "NL-be"
+	}
+  ],
+  "property2": {
+    "EN": "The Queen",
+    "dE": [ "Die Königin", "Ihre Majestät" ]
+  }
+}
--- a/tests/expand/0117-out.jsonld
+++ b/tests/expand/0117-out.jsonld
@@ -1,0 +1,39 @@
+[
+  {
+    "@id": "http://example.com/node",
+    "http://example.com/vocab/property": [
+      {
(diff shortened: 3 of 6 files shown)


gkellogg:
I wonder if now is the time to revisit this requirement. RDF only says implementations _may_ normalize to lowercase. I suspect it's there to ease testing, but there is also a consideration for RDF dataset normalization.

It may be that we should remove the requirement to normalize, since normal BCP47 practice is to use mixed case (i.e., "en-US" vs "en-us"). We could then put something the testing directions to cause them to output lower case language tags, or do case-insensitive comparison of `@language` in value objects.

RDF Dataset Normalization should probably do it's own normalization of language tags, as sources may come from other RDF serializations that don't require such normalization.

rubensworks:
I fully agree to remove the requirement to normalize language tags.

Since the normalization was required in JSON-LD 1.0,
should we still have tests to check whether this normalization is done?
Or should we just not check this at all?

If we still need tests, the I'll update this PR to only apply to 1.0 processors, and add a clarification in the README saying that test suites should check language tags in a case-insensitive manner for 1.1 tests. We may want to introduce a test flag for enabling strict language tag checks for this.

If we don't need the tests, then I guess this PR can be closed, and only a clarification in the test suite README is needed.

In any case, I've openened a new PR (https://github.com/w3c/json-ld-syntax/pull/282) to remove of the language tag normalization requirement.

gkellogg:
Note how the API document currently describes treating language tags:

>   <a>JSON-LD Processors</a> MUST NOT
    attempt to correct malformed <a>IRIs</a> or language tags;
    however, they MAY issue validation warnings. IRIs are not modified other
    than conversion between <a data-lt="relative IRI">relative</a> and
    <a>absolute IRIs</a>.

dlongley:
I'm fine with removing the language tag normalization requirement so long as the statement quoted above that says "JSON-LD processors MUST NOT..." cannot be construed to mean that a JSON-LD processor that supports canonicalization to N-Quads would somehow be a problem (as this is a future standardization target w/RDF canonicalization).

gkellogg:
@dlongley, I don't see this as a problem as RDF canonicalization is not really a function of a JSON-LD processor, but of a separate RDF canonicalization processor (which you happen to have integrated). It is often the case, and is true for my own processors, that the underlying RDF store does language normalization to lower-case, but that may not be the case. That statement refers to the algorithms defined within the spec; what happens outside in an RDF layer is outside of the scope of the document.

If you don't think that's clear, or needs to be elaborated, please suggest specific text to add to the spec.

dlongley:
@gkellogg,

I'm mostly in agreement, but perhaps to be safe we could tack something onto that quoted text above (or after the serializer/deserializer comment) such that it says something like this:

```
For clarity, JSON-LD processors MAY perform normalization
if they support common additional features that execute
after the core algorithms in this specification, such as
serializing RDF to a concrete syntax.
```

davidlehn:
Does anyone know why we did the normalization before?  I thought that was intentional.  One use case that will be odd is if you compact data using `"@container": "@language"`.  If the input data languages use different cases you would end up with different cased keys.  There should be a compact test for this one way or another.

I have no idea if mixed case tags are common in practice.  But I imagine if you have them and need to process that data, you'll end up doing case normalization at some point anyway to make the data easy to consume.

Input/Context:
```js
{
  "@context": {
    "@version": 1.1,
    "label": {
      "@id": "ex:label",
      "@container": ["@language", "@set"]
    }
  },
  "label": [
    {"@value": "en foo", "@language": "en-us"},
    {"@value": "alt en foo", "@language": "en-US"},
    {"@value": "de foo", "@language": "de"}
  ]
}
```
Output:
```js
{
  "@context": {
    "@version": 1.1,
    "label": {
      "@id": "ex:label",
      "@container": ["@language", "@set"]
    }
  },
  "label": {
    "en-us": ["en foo"],
    "en-US": ["alt en foo"],
    "de": ["de foo"]
  }
}
```

gkellogg:
We made many decisions based on what was easy to test.

Regarding language maps, presuming that the data being transformed is consistent, that wouldn't pose an issue. Otherwise, you might think it would be a problem for `@index` as well.

gkellogg:
This is obsolete now.

PR #165, opened 2019-10-07
Base direction

gkellogg:
For w3c/json-ld-syntax#11.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/165.html" title="Last updated on Oct 10, 2019, 8:38 PM UTC (b39ecba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/165/839a0db...b39ecba.html" title="Last updated on Oct 10, 2019, 8:38 PM UTC (b39ecba)">Diff</a>


gkellogg:
I have all the algorithm updates except for to/from RDF (a bit of fromRdf is in).

Changes to term selection, in particular, but also value expansion and compaction are fairly major, and really need a thorough review. But, it is based on updates to my implementation, so I believe the core logic is sound.

index.html
```diff
@@ -2214,19 +2264,23 @@ <h3>Algorithm</h3>
...
+                        <li class="changed">If <var>item</var> is `null`,
+                          continue to the next entry in <var>language value</var>.</li>
+                        <li><var>item</var> must be a <a>string</a>,
                           otherwise an
                           <a data-link-for="JsonLdErrorCode">invalid language map value</a>
                           error has been detected and processing is aborted.</li>
-                        <li>Append a <a class="changed">map</a> to
-                          <var>expanded value</var> that consists of two
+                        <li>Initialize a new <a class="changed">map</a> <var>v</var>
+                          consisteing of two
```
pchampin: typo
```suggestion
                          consisting of two
```

index.html
```diff
@@ -2395,11 +2449,14 @@ <h3>Algorithm</h3>
...
-              <code>@language</code> <a>entry</a> and an <code>@type</code> <a>entry</a>.
-              Otherwise, an
-              <a data-link-for="JsonLdErrorCode">invalid value object</a>
+              <code class="changed">@direction</code>,
+              `@index`,
+              `@language`,
+              `@type`,
+              and `@value`.
+              It must not contain either
+              `@language` or `@direction` <a>entries</a> and an `@type` <a>entry</a>.
```
pchampin: I had to read this sentence several times to get it right (and I knew in advance what it was supposed to say...).

I suggest to rephrase it as follows (sorry, can not make a patch on multiple lines):

> It must not contain a `@type` entry if it contains either `@language` or `@direction` entries.

index.html
```diff
@@ -2531,22 +2588,48 @@ <h3>Algorithm</h3>
...
+            <li class="changed">Otherwise, if a non-`null` <a>direction mapping</a> is associated with
+              <var>active property</var> in <var>active context</var>,
+              add <code>@language</code> to <var>result</var> and set its
+              value to the <a>default language</a> from <a>active context</a>,
+              if <a>active context</a> has a <a>default language</a>,
+              and the <a>language mapping</a> is not `null`.
+              Add a <code>@direction</code> to <var>result</var> and set its
+              value to the <a>direction mapping</a>; unless the
+              <a>direction mapping</a> is set to <code>null</code> in
+              which case no such <a>entry</a> is added.</li>
```
pchampin: "unless the direction mapping is set to null..." is useless here, as the condition at the start of the step prevents it from being null.

index.html
```diff
@@ -3264,9 +3350,9 @@ <h3>Algorithm</h3>
 
       <ol>
         <li>Initialize <var>result</var> to an empty <a class="changed">map</a>.</li>
-        <li>Initialize <var>default language</var> to <code>@none</code>. If the
-          <var>active context</var> has a <a>default language</a>,
-          set <var>default language</var> to it.</li>
+        <li>Initialize <var>default language</var> to <code>@none</code>.
+          If the <var>active context</var> has a <a>default language</a>,
+          set <var>default language</var> to the <a>language tag</a> from that <a>default language</a>.</li>
```
pchampin: "it" was a bit dry, but this phrasing makes it look like the language tag is somehow an attribute or a part of the default language...
I suggest reverting it to "it"
```suggestion
          set <var>default language</var> to it.</li>
```

index.html
```diff
@@ -4003,15 +4150,35 @@ <h3>Algorithm</h3>
           value matches the <a>type mapping</a> of <var>active property</var>,
           return the value associated with the <code>@value</code> <a>entry</a>
           of <var>value</var>.</li>
-        <li>Otherwise, if <var>value</var> has an <code>@language</code> <a>entry</a> whose
-          value matches the <a>language mapping</a> of
-          <var>active property</var>, return the value associated with the
+        <li class="changed">Otherwise, if <var>value</var> has both `@language` and `@direction <a>entries</a>,
```
pchampin: typo
```suggestion
        <li class="changed">Otherwise, if <var>value</var> has both `@language` and `@direction` <a>entries</a>,
```

index.html
```diff
@@ -4003,15 +4150,35 @@ <h3>Algorithm</h3>
...
+          and the `@language` <a>entry</a> matches either the <a>language mapping</a> of
+          <var>active property</var> or the <a>default language</a> of <var>active context</var>
```
pchampin: shouldn't that be

* it matches the language mapping of the active property OR
* **the language mapping of the active property is null** and it matches the default language of the active context
?

Same for the the direction?

index.html
```diff
@@ -4003,15 +4150,35 @@ <h3>Algorithm</h3>
...
+          and the `@direction` <a>entry</a> matches either the <a>direction mapping</a> of
+          <var>active property</var> or the <a>default base direction</a> of <var>active context</var>,
+          return the value associated with the
+          <code>@value</code> <a>entry</a> of <var>value</var>.</li>
+        <li class="changed">Otherwise, if <var>value</var> has a `@language` <a>entry</a>,
+          and the `@language` <a>entry</a> matches either the <a>language mapping</a> of
+          <var>active property</var> or the <a>default language</a> of <var>active context</var>
```
pchampin: Shouldn't that be:
```suggestion
          <var>active property</var> and the <a>default language</a> of <var>active context</var>
```

index.html
```diff
@@ -4003,15 +4150,35 @@ <h3>Algorithm</h3>
...
+          and the <a>direction mapping</a> of <var>active property</var> is `null`,
+          or there is no <a>default base direction</a> in <var>active context</var>,
+          return the value associated with the
+          <code>@value</code> <a>entry</a> of <var>value</var>.</li>
+        <li class="changed">Otherwise, if <var>value</var> has a `@direction` <a>entry</a>,
+          and the `@direction` <a>entry</a> matches either the <a>direction mapping</a> of
+          <var>active property</var> or the <a>default base direction</a> of <var>active context</var>
+          and the <a>language mapping</a> of <var>active property</var> is `null`,
+          or there is no <a>default language</a> in <var>active context</var>,
```
pchampin: Shouldn't that be
```suggestion
          and there is no <a>default language</a> in <var>active context</var>,
```

index.html
```diff
@@ -4760,15 +4927,20 @@ <h3>Overview</h3>
         Unless the <a data-link-for="JsonLdOptions">useRdfType</a> flag is set to true, <code>rdf:type</code>
         predicates will be serialized as <code>@type</code> as long as the associated object is
         either an <a>IRI</a> or <a>blank node identifier</a>.</p>
+
+      <p>If the </p>
```
pchampin: The suspense is killing me ;-)
```suggestion
```

index.html
```diff
@@ -4950,13 +5124,21 @@ <h3>Overview</h3>
...
+      <p class="changed">Literals with datatype <code>rdf:JSON</code>
+        are transformed into a value object using the internal representation
+        based on the lexical-to-value mapping defined in
+        <a data-cite="JSON-LD11#dfn-json-datatype" class="externalDfn">JSON datatype</a> in [[JSON-LD11]],
+        and <code>@type</code> of <code>@json</code>.</p>
+      <p class="changed">With the {{JsonLdOptions/rdfDirection}} option set to `i18n-datatype`,
+        literals with datatype starting with `https://w3.org/ns/i18n#`
+        are transformed into a <a>value object</a> by decoding
+        the <a>language tag</a> and <a>base direction</a> from the datatype.</p>
+      <p class="changed">With the {{JsonLdOptions/rdfDirection}} option set to `compount-literal`,
```
```suggestion
      <p class="changed">With the {{JsonLdOptions/rdfDirection}} option set to `compound-literal`,
```

gkellogg:
@rubensworks Updated your typo. Can you suggest some specific tests? I can see `https://w3.org/ns/i18n#en-us_abc`, but `https://w3.org/ns/i18n#a b_rtl` and others aren't valid IRIs.

I could see combinations of invalid language tags and invalid base directions; not sure if we are checking language tags against BCP47 in any tests, actually. We'd also need to check when converting from RDF.

I could see the following combinations, used in both expansion tests, toRdf tests, and fromRdf tests where decoding both `@langugage`, and `https://w3.org/ns/i18n#` and `rdf:language/rdf:direction`:

* language tags: `a+b`, `a.b`, `en-1`, etc.
* base direction `a`, `foo`, ...

But, I'm not sure we should be putting any real requirements for implementations to completely validate BCP47 other than some obvious problems like illegal characters ('+', '!', '~', `%`). But, it's a slippery slope.

index.html
```diff
@@ -1607,6 +1633,18 @@ <h3>Algorithm</h3>
               of <var>definition</var> to <var>language</var>.</li>
           </ol>
         </li>
+        <li class="changed">If <var>value</var> contains the <a>entry</a> <code>@direction</code> and
+          does not contain the <a>entry</a> <code>@type</code>:
```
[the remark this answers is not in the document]
gkellogg: https://github.com/w3c/json-ld-api/blob/66293aedb0de2ef937e81cf252afaa786f9908a9/index.html#L2457-L2460

index.html
gkellogg: Maybe we should update other uses like this in the spec, that imply that somehow `language tag` is encoded in `default language`.

Not much of a fan of "it" here, however. Maybe:

> If the _active context_ has a **default language**, set _default language_ to **default language**.

index.html
gkellogg: Fixed in a subsequent commit.

gkellogg:
> In the "Value compaction" algorithm, I find steps 7-8-9 very verbose (and yet still incomplete, see my comments). May be I'll make a counter proposal at some point if I find a simpler way to express them.

Yes, it is pretty bad. I think this is a case where we go away from precise algorithm to describing the intentions:

> Compact to a string, if any values of `@language` or `@direction` match those from the context, considering specific values in the term definition or defaults in the context and if the expanded value is not missing values for `@language` or `@direction` that should be there based on the term definition and defaults in the context.

Or, words to that effect.

gkellogg:
> In the "Value compaction" algorithm, I find steps 7-8-9 very verbose (and yet still incomplete, see my comments). May be I'll make a counter proposal at some point if I find a simpler way to express them.

I simplified value compaction, but there may be some corner cases, so I need to update my implementation to see if this handles everything, but it should read much better.

index.html
```diff
@@ -1261,6 +1273,19 @@ <h3>Algorithm</h3>
...
+              <ol>
+                <li>Initialize <var>value</var> to the value associated with the
+                  <code>@direction</code> <a>entry</a>.</li>
+                <li>If <var>value</var> is <code>null</code>, remove
+                  any <a>base direction</a> from <var>result</var>.</li>
+                <li>Otherwise, if <var>value</var> is <a>string</a>, the
+                  <a>base direction</a> of <var>result</var> is set to
+                  <var>value</var>. If it is not `null`, `"ltr"`, or `"rtl"`, an
+                  <a data-link-for="JsonLdErrorCode">invalid base direction</a>
+                  error has been detected and processing is aborted.</li>
```
dlongley: Do we need to consider whether or not `@direction` be protected?

index.html
```diff
@@ -2395,11 +2449,14 @@ <h3>Algorithm</h3>
...
-              and <code>@index</code>. It must not contain both a
-              <code>@language</code> <a>entry</a> and an <code>@type</code> <a>entry</a>.
-              Otherwise, an
-              <a data-link-for="JsonLdErrorCode">invalid value object</a>
+              <code class="changed">@direction</code>,
+              `@index`,
+              `@language`,
+              `@type`,
+              and `@value`.
+              It must not contain either
```
```suggestion
              It must not contain both
```

index.html
```suggestion
              an `@type` <a>entry</a> and either an `@language` or an `@direction` <a>entry</a>.
```

index.html
dlongley: It seems to me that using the word `both` here (and with the following change below) might help clarify?

index.html
```diff
@@ -3330,43 +3413,81 @@ <h3>Algorithm</h3>
...
+                <li>If <var>type map</var> does not have an <a>entry</a> corresponding
                   to the <a>type mapping</a> in <a>term definition</a>,
                   create one and set its value to the <a>term</a>
                   being processed.</li>
               </ol>
             </li>
+            <li>Reference the value associated with the <code>@language</code>
+              <a>entry</a> in <var>type/language map</var> using the variable
+              <var>language map</var>.</li>
+            <li class="changed">if <a>term definition</a> has both
```
```suggestion
            <li class="changed">If <a>term definition</a> has both
```

index.html
```diff
@@ -3330,43 +3413,81 @@ <h3>Algorithm</h3>
...
+              a <a>language mapping</a> and a <a>direction mapping</a>:
+              <ol>
+                <li>Create a new variable <var>lang dir</var>.</li>
+                <li>If neither the <a>language mapping</a> and the <a>direction mapping</a>
```
```suggestion
                <li>If neither the <a>language mapping</a> nor the <a>direction mapping</a>
```

index.html
```diff
@@ -3330,43 +3413,81 @@ <h3>Algorithm</h3>
...
+                  are `null`, set <var>lang dir</var> to the concatenation
+                  of <a>language mapping</a> and <a>direction mapping</a>
+                  separated by an underscore (`"_"`).</li>
+                <li>Otherwise, if <a>language mapping</a> is not `null`,
+                  set <var>lang dir</var> to the <a>language tag</a> in <a>language mapping</a>.
+                <li>Otherwise, if <a>direction mapping</a> is not `null`,
+                  set <var>lang dir</var> to <a>direction mapping</a>
+                  preceded by an underscore (`"_"`).</li>
```
dlongley: I guess I'm ok with this as a solution the language map + base direction issue. I'm not sure what else we'd do that wouldn't be much more complicated.

index.html
```suggestion
      <p class="changed">With the {{JsonLdOptions/rdfDirection}} option set to `compound-literal`,
```

index.html
gkellogg: Simplified Value Expansion along the lines of Value Compaction.

index.html
gkellogg: I think this was resolved in subsequent changes, can you verify?

gkellogg:
> Wow! Lots of changes here. That being said, they seem ok to me (modulo some comments I left). I'm obviously saying this without having implemented.
> 
> Also, I'm not a fan of the compound literal approach and would probably like to see DB's implementations only support the i18n type approach if we support anything other than dropping `@direction` to start when going to RDF. We don't want the technical debt of the compound literal approach so may not support it unless there's clear uptake there. Not sure if anything additional here needs to be marked at risk.

We should probably add _at risk_ to both of these algorithms here, and in the syntax document, so they don't come back to bite us if there's not sufficient implementation support, or the community otherwise rejects them.

Looking at the implementation, the compound literal mechanism leverages off of the list support, but is certainly much more heavyweight. I wouldn't be sad if we settled to just use the i18n-datatype mechanism.

index.html
gkellogg: How would this be protected?

* If aliased, it is protected along with other keyword aliases, I don't think anything special is needed for `@direction`.
* The _default base direction_ of a context is protected along with that context in the same way that _default language_ is protected.

* The _direction mapping_ of a term can be protected in the same way _language mapping_ is protected.

index.html
gkellogg: It certainly threw a wrench in the value expansion/compaction and term selection logic :(. Fortunately, only minimal changes to the Compaction and Expansion algorithms, themselves.

index.html
gkellogg: @pchampin's suggested change seems to have addressed this.

index.html
```diff
@@ -966,13 +966,16 @@ <h2>Context Processing Algorithm</h2>
...
+        and an optional <dfn>previous context</dfn>,
         used when a non-propagated <a>context</a> is defined</span>.
       Each <a>term definition</a> consists of:</p>
     <ul>
       <li>an <dfn>IRI mapping</dfn>,</li>
       <li>a boolean flag <dfn>reverse property</dfn>,</li>
-      <li>an optional <dfn>type mapping</dfn>, or <dfn>language mapping</dfn>,</li>
+       <li>an optional <dfn>type mapping</dfn>,</li>
+       <li>an optional <dfn>language mapping</dfn>,</li>
+      <li class="changed">an optional <dfn>direction mapping</dfn>,</li>
```
davidlehn: (pardon this comment here, github review limitiation, and a bit off topic) Down a few lines the "protected" text looks like it's missing a word.

davidlehn: And while I'm here... should these be using `<a>` that links to `<dfn>` text it terms.html (or elsewhere)?

index.html
```diff
@@ -4663,7 +4787,34 @@ <h3>Algorithm</h3>
...
           <code>xsd:string</code> or <code>rdf:langString</code>, depending on if
           item has an <code>@language</code> <a>entry</a>.</li>
-        <li>Initialize <var>literal</var> as an <a>RDF literal</a> using
+        <li class="changed">If <var>item</var> contains an `@direction` <a>entry</a>
+          and {{JsonLdOptions/rdfDirection}} is not `null`,
+          <var>item</var> is a <a>value object</a> which is serialized using special rules.
+          <ol>
+            <li>If {{JsonLdOptions/rdfDirection}} is `i18n-datatype`,
+              set <var>datatype</var> to the result of appending <var>language</var>
+              and the value of `@direction` in <var>item</var> separated by an underscore ('"_"`)
```
davidlehn:
```suggestion
              and the value of `@direction` in <var>item</var> separated by an underscore (`"_"`)
```

index.html
dlongley: > The default base direction of a context is protected along with that context in the same way that default language is protected.

If that's true already, this is what I was looking for. Great.

index.html
```diff
@@ -2395,11 +2452,13 @@ <h3>Algorithm</h3>
...
-              and <code>@index</code>. It must not contain both a
-              <code>@language</code> <a>entry</a> and an <code>@type</code> <a>entry</a>.
-              Otherwise, an
-              <a data-link-for="JsonLdErrorCode">invalid value object</a>
+              <code class="changed">@direction</code>,
+              `@index`,
+              `@language`,
+              `@type`,
+              and `@value`.
+              It must not contain a `@type` <a>entry</a> if it contains either `@language` or `@direction` <a>entries</a>.
```
```suggestion
              It must not contain an `@type` <a>entry</a> if it contains either `@language` or `@direction` <a>entries</a>.
```

index.html
gkellogg: I changed that to `a <dfn>protected</dfn> flag`. This defines the values contained within a **term definition**, so this does seem to be the source of those definitions.

However, there may be some uses of `<a>protect</a>` which refer to the concept of protection.

iherman:
> We should probably add at risk to both of these algorithms here, and in the syntax document, so they don't come back to bite us if there's not sufficient implementation support, or the community otherwise rejects them.

While I agree with this, I wonder whether, at this point, we should not consider these as non-normative, with a note that future versions of JSON-LD would settle on a normative version (or versions) depending on the community's reaction. I am not sure about it, but I still feel a bit uneasy about this WG suggesting _normatively_ a solution in a space that goes beyond JSON-LD proper.

iherman:
On the `...i18n#lt_dir` option, although the specification itself is correct, we may want to emphasize (e.g., in a note) that the `lt` part _may_ be empty, i.e., an `...i18n#_ltr` _is_ a correct datatype.

rubensworks:
@gkellogg 

> Can you suggest some specific tests? I can see https://w3.org/ns/i18n#en-us_abc, but https://w3.org/ns/i18n#a b_rtl and others aren't valid IRIs.

Indeed, it was my goal to make them invalid IRIs :-) (they would definitely be applicable for the toRdf tests)
But you're right that if we go this route, then we should probably (also?) include valid IRIs, but invalid language tags, as you suggested below.

> I could see combinations of invalid language tags and invalid base directions; not sure if we are checking language tags against BCP47 in any tests, actually. We'd also need to check when converting from RDF.

There is indeed one test that [checks the validity of language tags](https://github.com/w3c/json-ld-api/blob/master/tests/toRdf-manifest.jsonld#L2329-L2337).

> I could see the following combinations, used in both expansion tests, toRdf tests, and fromRdf tests where decoding both @langugage, and https://w3.org/ns/i18n# and rdf:language/rdf:direction:

>    language tags: a+b, a.b, en-1, etc.
     base direction a, foo, ...

👍 

> But, I'm not sure we should be putting any real requirements for implementations to completely validate BCP47 other than some obvious problems like illegal characters ('+', '!', '~', %). But, it's a slippery slope.

We do it for IRIs, so in that sense it also makes sense to do it for BCP47. I don't have a strong opinion on this though, so perhaps this can be left open for a future JSON-LD version.
But if we decide to do proper BCP47 validation, then I'd be happy to contribute some tests for this.

gkellogg:
For the tests to work, all IRIs need to be parsable as N-Quads, so we can't effectively test truly illegal IRIs there.

The issue of whether we should validate language tags should be discussed on a call, as the current text only says that processors MAY warn on invalid language tags.

gkellogg:
> Indeed, it was my goal to make them invalid IRIs :-) (they would definitely be applicable for the toRdf tests)
> But you're right that if we go this route, then we should probably (also?) include valid IRIs, but invalid language tags, as you suggested below.

As I noted, using invalid IRIs is problematic, as you can't parse an N-Quads document with an invalid IRI.

> > I could see combinations of invalid language tags and invalid base directions; not sure if we are checking language tags against BCP47 in any tests, actually. We'd also need to check when converting from RDF.
> 
> There is indeed one test that [checks the validity of language tags](https://github.com/w3c/json-ld-api/blob/master/tests/toRdf-manifest.jsonld#L2329-L2337).
> 
> > I could see the following combinations, used in both expansion tests, toRdf tests, and fromRdf tests where decoding both @langugage, and https://w3.org/ns/i18n# and rdf:language/rdf:direction:
> 
> > language tags: a+b, a.b, en-1, etc.
> > base direction a, foo, ...
> 
> 👍
> 
> > But, I'm not sure we should be putting any real requirements for implementations to completely validate BCP47 other than some obvious problems like illegal characters ('+', '!', '~', %). But, it's a slippery slope.
> 
> We do it for IRIs, so in that sense it also makes sense to do it for BCP47. I don't have a strong opinion on this though, so perhaps this can be left open for a future JSON-LD version.
> But if we decide to do proper BCP47 validation, then I'd be happy to contribute some tests for this.

As it happens, [RDF Concepts[(https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal) _does_ require that language tags be well formed according to BCP47:

> ... a non-empty **language tag** as defined by [BCP47]. The language tag **must* be well-formed according to section 2.2.9 of [BCP47].

The other RDF syntaxes use something like the following to validate language tags:

```
[144s] LANGTAG                        ::= "@" [a-zA-Z]+ ( "-" [a-zA-Z0-9]+ )* 
```

I could see us changing from a MAY validate language tags to MUST be valid either to ` [a-zA-Z]+ ( "-" [a-zA-Z0-9]+ )* `, or the [more strict ABNF from BCP47](https://tools.ietf.org/html/bcp47#section-2.2.9):

       obs-language-tag = primary-subtag *( "-" subtag )
       primary-subtag   = 1*8ALPHA
       subtag           = 1*8(ALPHA / DIGIT)

This should probably be done in context processing, expansion, and fromRDF algorithms to be consistent. The RDF Concepts regex is probably the one to use, as we're in that family.

I created issue #167 for this.

gkellogg:
I'm going to merge this, as all review comments have been cleared, and the remaining issue of validating language tags has been raised in #167.

index.html
```diff
@@ -3957,77 +4078,73 @@ <h3>Algorithm</h3>
...
+            <li>If <var>value</var> has an `@index` <a>entry</a>,
+              and the <a>container mapping</a> associated to <var>active property</var>
+              includes <code>@index</code>,
+              set <var>result</var> to the value associated with the `@value` <a>entry</a>.</li>
+          </ol>
+        </li>
+        <li>If the value of `@language` in <var>item</var> exactly matches <var>language</var>,
+          if it is not `null`, or is not present, if it is `null`,
+          <span class="changed">and the value of `@direction` in <var>item</var> exactly matches <var>direction</var>,
+            if it is not `null`, or is not present, if it is `null`</span>:
```
pchampin: I think this would be clearer
```suggestion
            if it is not `null`, or is not present, if <var>direction</var> is `null`</span>:
```

index.html
```diff
@@ -3957,77 +4078,73 @@ <h3>Algorithm</h3>
...
+          <ol>
+            <li>If <var>value</var> has an `@index` <a>entry</a>,
+              and the <a>container mapping</a> associated to <var>active property</var>
+              includes <code>@index</code>,
+              or <var>value</var> has no `@index` <a>entry</a>,
+              set <var>result</var> to the value associated with the `@value` <a>entry</a>.</li>
+          </ol>
+          return <var>value</var>.</li>
```
pchampin: This seems to be a remains of a previous version
```suggestion
```

index.html
```diff
@@ -3957,77 +4078,73 @@ <h3>Algorithm</h3>
...
+          in <var>active context</var>, if any, otherwise to the <a>default language</a>
+          of <var>active context</var>.</li>
+        <li class="changed">Initialize <var>direction</var> to the <a>direction mapping</a> for <var>active property</var>
+          in <var>active context</var>, if any, otherwise to the <a>default base direction</a>
+          of <var>active context</var>.</li>
+        <li>If <var>value</var> has an <code>@id</code> <a>entry</a>
+          and has no other <a>entries</a> other than `@index`:
           <ol>
-            <li>If <var>number entries</var> is <code>1</code> and
+            <li>If <var>value</var> and
```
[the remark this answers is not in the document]
gkellogg: Looks like that was cleared up since your comment.

index.html
```diff
@@ -3957,77 +4078,73 @@ <h3>Algorithm</h3>
...
               the <a>type mapping</a> of <var>active property</var>
-              is set to <code>@id</code>, return the result of using the
+              is set to <code>@id</code>, set <var>result</var> to the result of using the
```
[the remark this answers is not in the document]
gkellogg: It could be done that way, and was previously, but I found it better to simply update _result_, to allow for key transformation, if it is still a _map_, before returning it.

index.html
```diff
@@ -3957,77 +4078,73 @@ <h3>Algorithm</h3>
...
               <a href="#iri-compaction">IRI compaction algorithm</a>,
               passing <var>active context</var>, <var>inverse context</var>,
               and the value of the <code>@id</code> <a>entry</a> for <var>var</var>.</li>
-            <li>Otherwise, if <var>number entries</var> is <code>1</code> and
-              the <a>type mapping</a> of <var>active property</var>
-              is set to <code>@vocab</code>, return the result of using the
+            <li>Otherwise, the <a>type mapping</a> of <var>active property</var>
+              is set to <code>@vocab</code>, set <var>result</var> to the result of using the
```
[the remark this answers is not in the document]
gkellogg: Same consideration here, keep _result_ through the remaining steps of the algorithm, and return at the end.

index.html
gkellogg: It is now "return _result_".

index.html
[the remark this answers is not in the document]
pchampin: My understanding was that the IRI compaction algorithm returned a string, so there would be no more key to transform... hence my proposal to return it immediately.
Or did I miss something in the IRI compaction algorithm?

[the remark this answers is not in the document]
gkellogg: No, you're right, and it could be returned early, but there's something to be said for having the exit at the end of the algorithm. Having returns scattered through the algorithm might be less obvious, and the "Otherwise" clauses make it clear that you get to the end, where we see if it's a map.

But, if the consensus is to do early returns, I'll go along with that. @dlongley?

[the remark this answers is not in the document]
gkellogg: (We should probably move this to a new issue or PR, as this PR is closed, and it can be confusing what conversations were resolved in the code, and what is still open for consideration).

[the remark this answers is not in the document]
pchampin: @gkellogg 
> there's something to be said for having the exit at the end of the algorithm

very true

> and the "Otherwise" clauses make it clear that you get to the end

correct, I overlooked that.

So I'm happy with keeping those as 'set result'...

Issue #167, opened 2019-10-10
Require valid language tags


gkellogg:
Currently, JSON-LD doesn't do any language-tag validation. However, [RDF Concepts[(https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal) _does_ require that language tags be well formed according to BCP47:

> ... a non-empty **language tag** as defined by [BCP47]. The language tag **must* be well-formed according to section 2.2.9 of [BCP47].

The other RDF syntaxes use something like the following to validate language tags:

```
[144s] LANGTAG                        ::= "@" [a-zA-Z]+ ( "-" [a-zA-Z0-9]+ )* 
```

I could see us changing from a MAY validate language tags to MUST be valid either to ` [a-zA-Z]+ ( "-" [a-zA-Z0-9]+ )* `, or the [more strict ABNF from BCP47](https://tools.ietf.org/html/bcp47#section-2.2.9):

       obs-language-tag = primary-subtag *( "-" subtag )
       primary-subtag   = 1*8ALPHA
       subtag           = 1*8(ALPHA / DIGIT)

This should probably be done in context processing, expansion, and fromRDF algorithms to be consistent. The RDF Concepts regex is probably the one to use, as we're in that family.

We do say that a language tag MUST be well formed:

> language-tagged string

>A language-tagged string consists of a string and a non-empty language tag as defined by [BCP47]. The language tag MUST be well-formed according to section 2.2.9 Classes of Conformance of [BCP47], and is normalized to lowercase.

The [Conformance Section](https://www.w3.org/TR/json-ld11-api/#conformance) of the API document waffles on this:

> JSON-LD Processors MUST NOT attempt to correct malformed IRIs or language tags; however, they MAY issue validation warnings. IRIs are not modified other than conversion between relative and absolute IRIs.

Note the fine distinction between correction and validate and that warnings MAP be issued, rather than errors MUST be generated.

This should probably change to processors MUST generate an error and either abort or ignore invalid IRIs or language tags (or base directions). We do require that these be well formed when generating RDF.

akuckartz:
SHOULD instead of MUST ?

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-10-11-json-ld.html#section4-3).

- `RESOLVED:  Change requirement to lowercase language tags to be a MAY from MUST`
[...]

PR #168, opened 2019-10-13
Language tags

gkellogg:
* Remove requirement to normalize language tags to lower case, but note testing considerations.
* Notes about normalizing language tags, issuing warnings when not well-formed, and not considering case when using for term selection.
* Add a test for term selection where case of language does not match.

For #167.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/168.html" title="Last updated on Oct 14, 2019, 9:18 PM UTC (1c62284)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/168/5e314e9...1c62284.html" title="Last updated on Oct 14, 2019, 9:18 PM UTC (1c62284)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1275,7 +1275,10 @@
                   If it is not a <a>string</a>, an
                   <a data-link-for="JsonLdErrorCode">invalid default language</a>
                   error has been detected and processing is aborted.
-                  <span class="changed note">Processors MAY normalize <a>language tags</a> to lowercase.</span>
+                  <span class="changed">If <var>value</var> is not a <a>well-formed</a> according to
+                    <a data-cite="BCP47#section-2.2.9">section 2.2.9</a> of [[BCP47]],
+                    processors SHOULD issue a warning.</span>
+                  <span class="changed note">Processors MAY normalize <a>language tags</a> to lower case.</span>
                 </li>
               </ol>
             </li>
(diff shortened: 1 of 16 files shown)


tests/compact-manifest.html
```diff
@@ -46,9 +46,10 @@ <h2>JSON-LD Object comparison</h2>
 <p>JSON-LD Object comparison compares JSON objects, arrays, and values recursively for equality.</p>
 
 <ul>
-<li>JSON objects are compared member by member without regard to the ordering of members within the object. Each member must have a corresponding member in the object being compared to. Values are compared recursively.</li>
+<li>JSON objects are compared entry by entry without regard to the ordering of entriess within the object. Each entry must have a corresponding entry in the object being compared to. Values are compared recursively.</li>
```
rubensworks:
```suggestion
<li>JSON objects are compared entry by entry without regard to the ordering of entries within the object. Each entry must have a corresponding entry in the object being compared to. Values are compared recursively.</li>
```
(also in the other manifests)

index.html
```diff
@@ -909,7 +909,9 @@ <h2>RDF Serialization/Deserialization</h2>
 
   <p><a>JSON-LD Processors</a> MUST NOT
     attempt to correct malformed <a>IRIs</a> or language tags;
-    however, they MAY issue validation warnings. IRIs are not modified other
+    however, they SHOULD issue validation warnings.
+    <span class="changed">Invalid language tags MUST be ignored.</span>
```
rubensworks: What does this mean exactly?

For example, in toRdf, should the whole triple be omitted, or should the object become a literal without language tag? (I assume the former?)

I also don't see this mentioned in the algorithms.

gkellogg: Probably should just stick with the previous wording, changing MAY to SHOULD. In fact, the algorithms issue a warning and continue with the (possibly invalid) language tag.

In toRdf, such triples would not be emitted, using the provision that only valid triples are emitted, inasmuch as such triples would be invalid.

gkellogg: Fixed in 0d4556303057f8712724a510628ee04ba0764694.

index.html
```diff
@@ -2133,12 +2142,17 @@ <h3>Algorithm</h3>
                   <var>value</var> is not a <a>string</a>, an
                   <a data-link-for="JsonLdErrorCode">invalid language-tagged string</a>
                   error has been detected and processing is aborted.
+                  <span class="changed">If <var>value</var> is not <a>well-formed</a> according to
```
```suggestion
                  <span class="changed">Otherwise if <var>value</var> is not <a>well-formed</a> according to
```
This is require so that the "Otherwise" below (line 2149) still makes sense.

gkellogg: Actually, I think the clauses should be inverted, so the Otherwise has to do with setting expanded value to value, and do the warning later.

> If expanded property is `@language` and value is not a string, an **invalid language-tagged string** error has been detected and processing is aborted, _otherwise, set `expanded value` to `value`_. If `value` is not well-formed according to section 2.2.9 of [BCP47], processors SHOULD issue a warning.  ...

gkellogg: Fixed in 1c62284b358db33cab889fbff62aee2558dce78c.

PR #174, opened 2019-10-21
Add wording to allow a default object

gkellogg:
containing an _IRI_ as the value of `@type` when framing.

For w3c/json-ld-framing#74.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/174.html" title="Last updated on Oct 21, 2019, 5:16 PM UTC (f6c7fb2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/174/2abfba2...f6c7fb2.html" title="Last updated on Oct 21, 2019, 5:16 PM UTC (f6c7fb2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2415,7 +2415,9 @@
                     transforming it into an <a>array</a>, if necessary.</span>
                   <span class="changed">
                     When the {{JsonLdOptions/frameExpansion}} flag is set, <var>value</var>
-                    may also be an empty <a>map</a>.
+                    may also be an empty <a>map</a>, or a <a>default object</a>.
+                    For a <a>default object</a>, the value of `@default` is restricted to be
+                    an <a>IRI</a>, which is expanded as above.
                   </span>
                   <div class="note">
                     No transformation from a <a>string</a> <var>value</var> to an <a>array</a>


pchampin:
This is strange: the 'diff' shows some change in the terminology section, which is not part of this PR.

It says
> A default object is a map that has a `@default` key. 

Shouldn't it say "that has a single entry, whose key is `@default`", or something like that?

pchampin:
This is strange: the 'diff' shows some change in the terminology section, which is not part of this PR.

It says
> A default object is a map that has a `@default` key.

This definition should be expanded a little. Can a default object contain other keys? I couldn't find any example in the spec or the tests where `@default` co-exists with other keys...

Then in section 2.2, I read
> The `@default` property in a node object or value object provides a default value (...)

That imply that such a node/value object would *also* be a default object (per the definition). I was able to build a working example in the playground with a hybrid "default + node object", but was not able to make a working example with a hybrid "default + value object".

Suggestions : 

* expand the definition of "default object" to make it explicit that it is not exclusive with "node object" (and "value object" ?)

*

gkellogg:
> Suggestions :
> 
> * expand the definition of "default object" to make it explicit that it is not exclusive with "node object" (and "value object" ?)

It is actually intended that a `default object` be a map containing only the entry for `@default`. The value will be interpreted depending on the context, and could expand to a node object or value object (or just an IRI for `@type`).

We should clarify that only the first value is used, so a `default object` containing multiple values will only cause the first value to be used. This may be a side-effect of the algorithm/implementation rather than a design decision. Only @dlongley knows for sure 😃.

PR #177, opened 2019-10-22
Fix error codes in tests being emitted as IRIs

rubensworks:
Currently, the expected error codes in NegativeEvaluationTests are defined using the `expect` context term.
However, this term is defined with `"@type": "@id"`, which causes the error codes to become IRIs, which is incorrect, as the spec describes them to be plain strings.

This PR adds a new `expectErrorCode` context entry, which expands to the same predicate as `expect`, but as a plain string. All NegativeEvaluationTests have been updated to make use of `expectErrorCode` instead of `expect`.


rubensworks:
@gkellogg I've applied the changes you requested.

I did not change README.md, as it did not have any explicit mentions to `expect`, and instead refers to the vocab, where the distinction between `expect` and `expectErrorCode` is mentioned. (If you think this would be beneficial, I can add this into README.md as well)

gkellogg:
Thanks, Ruben. This is great. If running the tests is unclear, we'll get feedback.

Can you do the same for the framing tests?

gkellogg:
I'll take care of the framing tests, as you don't work with that.

PR #184, opened 2019-10-30
Clarify that expansion steps looking for entries

### LINKED ISSUE #175: Clarify trailing text in expansion algorithm
The [expansion algorithm](https://w3c.github.io/json-ld-api/#expansion-algorithm) includes this trailing text:

> If, after the above algorithm is run, the result is a map that contains only an @graph entry, set the result to the value of @graph's value. Otherwise, if the result is null, set it to an empty array. Finally, if the result is not an array, then set the result to an array containing only the result.

It is not clear to me if this post-handling of the result is applied universally to all values returned out of the expansion algorithm. For example, step 1 ends `return null`. Why use `null` here instead of an empty array if the post-processing will always do the conversion?

Alternatively, based on the overview in 5.1.1, I think it's possible that the post-processing detailed in the trailing text is meant to apply only to steps 6 and beyond. If that's the case, the language "after the above *algorithm* is run" (emphasis added) is misleading.

The use of `return` throughout the algorithm which is then required to have post-processing applied to it is confusing and a strange overload to what I would think of as the normal definition of `return`. Can this please be clarified and the language improved to avoid the confusion as to what parts of the algorithm the post-processing is applied to, and to remove any unnecessary values (such as the `null` return in step 1) which are almost immediately replaced with other values?

### LINKED ISSUE #183: Clarify type expectations for the result of expansion
The final few steps of the [expansion algorithm](https://w3c.github.io/json-ld-api/#algorithm-3) use language that seems to assume that `result` is a map.

In 17:

> If result contains only the entry @language

And in 18.1:

> If result is an empty map or contains the entries @value or @list, set result to null.

However, step 16 seems to allow `result` to become a non-map value in 16.2:

> If result contains the entry @set, then set result to the entry's associated value.

If steps 17 and 18 are meant to be conditional on `result` being a map, can this please be explicitly stated?

gkellogg:
only apply when _result_ is a map.

Also,  split-up and remove final step of Expansion Algorithm.

Fixes #175 and fixes #183.

cc/ @kasie

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/184.html" title="Last updated on Oct 31, 2019, 10:10 PM UTC (2dc02c1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/184/2f36a51...2dc02c1.html" title="Last updated on Oct 31, 2019, 10:10 PM UTC (2dc02c1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2861,15 +2861,15 @@
                 containing only the <code>@id</code> <a>entry</a> is retained.</span></li>
           </ol>
         </li>
+        <li>If <var>result</var> is a
+          <a class="changed">map</a> that contains only an <code>@graph</code> <a>entry</a>,
+          set <var>result</var> that value.</li>
+        <li>Otherwise, if <var>result</var> is <code>null</code>,
+          set <var>result</var> to an empty <a>array</a>.</li>
+        <li>If <var>result</var> is not an <a>array</a>,
+          set <var>result</var> to an <a>array</a> containing only <var>result</var>.</li>
         <li>Return <var>result</var>.</li>
       </ol>
-
-      <p>If, after the above algorithm is run, the result is a
-        <a class="changed">map</a> that contains only an <code>@graph</code> <a>entry</a>, set the
-        result to the value of <code>@graph</code>'s value. Otherwise, if the result
-        is <code>null</code>, set it to an empty <a>array</a>. Finally, if
-        the result is not an <a>array</a>, then set the result to an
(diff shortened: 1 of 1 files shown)


kasei:
Thanks @gkellogg . Just FYI, you're tagging someone else in that cc. `s/ie/ei/`

PR #186, opened 2019-10-30
Clarifications on _container mapping_

### LINKED ISSUE #181: Clarify text regarding container mappings in algorithms
In the [Create Term Definitiion](https://w3c.github.io/json-ld-api/#algorithm-0) algorithm, step 22 deals with an entry's '@container' value. In 22.1:

> Initialize container to the value associated with the @container entry, which MUST be either @graph, @id, @index, @language, @list, @set, @type, or an array containing exactly any one of those keywords, an array containing @graph and either @id or @index optionally including @set, or an array containing a combination of @set and any of @index, @id, @type, @language in any order .

(Note that this definition allows both scalar and array types.)

However, subsequent steps talk about a container mapping as though the type were known to be an array:

* In 22.4: "If container mapping is @type" (which might imply that it can't be an array containing '@type'(?))
* In 23.1: "or container mapping does not include @index" (implying that it must be an array?)

And in the [expansion algorithm](https://w3c.github.io/json-ld-api/#algorithm-3):

* In 5.2.2: "If the container mapping of active property includes @list" (implying that it can't *be* '@list'?)
* In 13.7: "Otherwise, if container mapping includes @language" (implying that it can't *be* '@language'?)
* In 13.8: "Otherwise, if container mapping includes @index, @type, or @id"
* And again in 13.8.3.1, 13.8.3.2, 13.8.3.7.1, 13.8.3.7.2, 13.8.3.7.3, , 13.8.3.7.4, 13.8.3.7.5, 13.11 and 13.12

As an aside, the text of 22.4 should probably clarify "container mapping" as "the [container mapping] of definition".

gkellogg:
Now it is always represented as an array in a term definition.

Fixes #181.

cc/ @kasei

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/186.html" title="Last updated on Oct 31, 2019, 11:39 PM UTC (62d07f1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/186/d930363...62d07f1.html" title="Last updated on Oct 31, 2019, 11:39 PM UTC (62d07f1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1490,7 +1490,8 @@
               error has been detected and processing is aborted.</li>
             <li>If <var>value</var> contains an <code>@container</code> <a>entry</a>,
               set the <a>container mapping</a> of <var>definition</var>
-              to its value; if its value is neither <code>@set</code>, nor
+              to <span class="changed">an <a>array</a> containing</span> its value;
+              if its value is neither <code>@set</code>, nor
               <code>@index</code>, nor <code>null</code>, an
               <a data-link-for="JsonLdErrorCode">invalid reverse property</a>
               error has been detected (reverse properties only support set- and
@@ -1613,8 +1614,9 @@
               generate an <a data-link-for="JsonLdErrorCode">invalid container mapping</a>
               error and abort processing if <a>processing mode</a> is `json-ld-1.0`.</li>
             <li>Set the <a>container mapping</a> of <var>definition</var> to
-              <var>container</var>.</li>
-            <li class="changed">If <a>container mapping</a> is `@type`:
+              <var>container</var>
+              <span class="changed">coercing to an <a>array</a>, if necessary</span>.</li>
+            <li class="changed">If the <a>container mapping</a> of <var>definition</var> includes `@type`:
               <ol>
                 <li>If <var>type mapping</var> in <var>definition</var> is undefined, set it to `@id`.</li>
                 <li>If <var>type mapping</var> in <var>definition</var> is neither `@id` nor `@vocab`,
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1613,8 +1614,9 @@ <h3>Algorithm</h3>
               generate an <a data-link-for="JsonLdErrorCode">invalid container mapping</a>
               error and abort processing if <a>processing mode</a> is `json-ld-1.0`.</li>
             <li>Set the <a>container mapping</a> of <var>definition</var> to
-              <var>container</var>.</li>
-            <li class="changed">If <a>container mapping</a> is `@type`:
+              <var>container</var>
+              <span class="changed">coercing to an <a>array</a>, if necessary</span>.</li>
+            <li class="changed">If the <a>container mapping</a> of <var>definition</var> includes `@type`:
```
pchampin: Should it be "includes `@type`", or "has a single item `@type`" ??
(the second is how I would have interpreted the old text... In any case, this shows that the previous wording was ambiguous and worth changing)

index.html
```diff
@@ -3268,7 +3270,7 @@ <h3>Algorithm</h3>
           return that result.</li>
         <li class="changed">If <var>element</var> is a
           <a>list object</a>, and the <a>container mapping</a> for
-          <var>active property</var> in <var>active context</var> is <code>@list</code>,
+          <var>active property</var> in <var>active context</var> <span class="changed">includes</span> <code>@list</code>,
```
pchampin: Same question as above...

index.html
gkellogg: The only possibilities are `"@type"`, `["@type"]`, and `["@type", "@set"]`, and this step should pertain to all of those. Any other combinations won't make it passed create term definition. So, "contains `@type`" should be correct.

pchampin: Ok then.

Issue #195, opened 2019-11-03
Containing a colon does not mean an IRI is absolute or blank node identifier


lo48576:
https://github.com/w3c/json-ld-api/blob/dd3fdfdf76834464f03a817ba136606acef783d9/index.html#L1486-L1490

[RFC 3986 §3.3. Path](https://tools.ietf.org/html/rfc3986#section-3.3) says that path segments can have colon characters.
So, "If the result is neither an IRI nor a blank node identifier" is not equal to "it contains no colon `:`".
Counter example is `./foo:bar`: it **is** "neither an IRI nor a blank node identifier", but **not** "it contains no colon".

I didn't check other places saying "contains a colon" is correct or not, but "contains a colon" and "is an IRI or a blank node identifier" should be used in non-confusing way, and should be corrected if unexpected one is used...

pchampin:
Per RFC 3986`./foo:bar` is not an IRI (which is always in absolute form), it is an *IRI reference*. We came to this realization when addressing w3c/json-ld-syntax#288.

Previously, the part of the document you cite read

> If the result is nether an *absolute* IRI nor a blank node identifier...

(emphasis is mine) but we resolved to align our terminology to that of RFC3986, and replace "relative IRI" with "relative IRI reference", and "absolute IRI" with "IRI" (because "absolute IRI" rules out fragments identifiers, while "IRI" does not).

I [expressed my concern](https://github.com/w3c/json-ld-syntax/issues/288#issuecomment-547975144) that "most people reading 'IRI' may wrongly think: it might be absolute or relative". I find it quite ironic that the first person to confirm my concern is the very person who pointed out the problem in our terminology... :->

pchampin:
You are correct, checking for a colon is not enough for deciding that we have an IRI or blank node identifier...

I scanned the mention of "colon" in the spec, and I think the only other place where this mistake needs fixing is step 6 of [§5.2.2](https://w3c.github.io/json-ld-api/#algorithm-4) (IRI Expansion).

gkellogg:
In 1.0 we made a decision to make IRI detection simple by just looking for a colon. This certainly doesn’t protect against such obvious attacks, but is really an sue for interpreting most JSON. To do it right, we should use wording such as “has the form of an IRI as described in section xxx of RFC3987”, which would require more work in the part of processors to check for this at the expense of simplicity.

[...]

Issue #189, opened 2019-10-31
Processing terms starting with :


azaroth42:
What should a processor do with a term that starts with `:` ?

@kasei and my reading is that, with an `@vocab` definition, it should create a term by concatenation.

For example:
```
{"@context": {"@vocab": "http://schema.org/"},
 "@id": "foo:bar-id",
 "@type": "foo:bar-type",
 "foo:bar": "is an absolute iri property",
 "term": "is schema.org/term",
 ":fish": "is schema.org/:fish"
}
```

In the playground the term is ignored completely:
  http://tinyurl.com/yxk2vka2

In Gregg's distiller it generates the `schema.org/:fish` property.

The text in https://www.w3.org/TR/json-ld11-api/#algorithm-4 step 6 says "contains a colon", which it does, and then to split it. This would in theory generate a prefix of empty string and a term of 'fish'. As there isn't an expansion possible, it seems like the playground drops it.

We think that step 6 should be modified to say that it contains a colon, **as the second or subsequent character**. This would make Gregg's implementation correct, and the vocab would used.

There should also be a test for this (admittedly pathological) case.

gkellogg:
Perhaps: "contains a colon anywhere after the first character."

Other potential terms are:

* `:foo:`
* `:foo`
* `foo:`

Note that we specifically handle a term `foo:` in 17.5 of Create Term Definition. We should also update step 18.

azaroth42:
What would we expect `:foo:` to do? Given that the first character is `:`, I would think this is also just a term, and thus would be concatenated with the vocabulary to form an IRI. But it would still match the "contains a colon after the first character" test. 

Perhaps it's a special case for the first or last character being a colon?

gkellogg:
It would be treated much like `foo:`, meaning that constructing a compact IRI would not add a new `:` between prefix and suffix, and expanding it would  not remove that trailing colon, so that the term with the training colon would be retained. In this case, it would also begin with a colon.

IIRC, the trainling-colon term was added to ensure that it could be used, and was intended to be used as a prefix. No good reason to create a term with a leading colon, but we should handle it one way or the other.

iherman:
Agree with closing.

PR #203, opened 2019-11-06
Terms with colon

gkellogg:
* Allow terms starting with a colon to not be treated as IRIs. For #189
* Add "document relative" when expanding values of `@reverse` in term creation.
  Remove the check for IRI or blank node after calling IRI expansion, as if it were a native value, it would have been detected, and otherwise, strings will be percent-encoded to create IRIs.
  For #195.

cc/ @azaroth42 @kasei @lo48576

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/203.html" title="Last updated on Nov 7, 2019, 9:23 PM UTC (356e4f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/203/0d787f2...356e4f4.html" title="Last updated on Nov 7, 2019, 9:23 PM UTC (356e4f4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1520,14 +1520,12 @@
               return; processors SHOULD generate a warning.</li>
             <li>Otherwise, set the <a>IRI mapping</a> of <var>definition</var> to the
               result of using the <a href="#iri-expansion">IRI Expansion algorithm</a>,
-              passing <var>active context</var>, the value associated with
-              the <code>@reverse</code> <a>entry</a> for <var>value</var>, <code>true</code>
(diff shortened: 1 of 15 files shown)


gkellogg:
You’re right, it was set in my implementation, and it looked like it was missing from the spec, but it should be symmetric to `@id`. For some reason it seemed right yesterday, I’ll reverse this and add back some of that language on invalid IRI.

PR #208, opened 2019-11-11
Algorithm clarifications

### LINKED ISSUE #206: Clarification in algorithms
This is a bunch of minor changes in the algorithms that I think would make them clearer...

Context Processing
* [x]  §4.1 It would be nice to indicate the datatype of each "attribute" of a term definition (especially for implementers using a strongly typed language)
* [x] §4.1 step 2:  the "MUST be a boolean" is confusing, because no error handling is specified here; it would be helpful to add (as a note) that the erreor will be handled later (in step 5.11)
* [x] §4.1 step 5.2.1: _value_ is not defined (should be _context_)
* [x] §4.1 step 5.2.1: it is not not clear what base URI should be used, probably that from _result_?
* [x] §4.1 step 5.6.4: Shouldn't this step be preceded by sometheling similar to 5.2.3?
* [x] §4.1 step 5.6.4: Shouldn't this step be preceded by sometheling similar to 5.2.3?
* [x] §4.1 step 5.13: Shouldn't _override_protected_ be also be passed to create-term-definition?

Create Term Definition
* [x]  §4.2 step 12: if `@protected` is not a boolean, shouldn't this raise an error  (invalid `@protected` value)? NB: it does not exist yet
* [x]  §4.2 step 16.4: only `@context` is checked for an "invalid keyword alias", but the specs also forbids aliasing `@base`, `@container`, `@import`, `@prefix`, `@propagate`, `@protected`, `@version`, `@vocab`.
* [x]  §4.2 step 16.5: the parameters passed to IRI Expansion should be explicit.

IRI Expansion
* [x] §5.2 step 8: it is not clear which *base IRI* is referred to. I'm guessing this is _active_context.base_. What if it is null?

I'll make a PR to that effect this evening (French timezone).

gkellogg:
* Type entries in a term definition and describe the properties of the active context.
* Add note to step on error handling for `@propagate`.
* Change _value_ to _context_ in context processing, and clarify that the base IRI used for resolving relative IRIs is that of the document containing the _local context_.
* Describe and use **invalid `@protected` value**.
* Be more explicit about parameters passed to IRI expansion.

Fixes #206.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/208.html" title="Last updated on Nov 11, 2019, 11:58 PM UTC (45f27b5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/208/c6147ce...45f27b5.html" title="Last updated on Nov 11, 2019, 11:58 PM UTC (45f27b5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2490,7 +2490,7 @@
                       error has been detected and processing is aborted.
                       <span class="changed">
                         When the {{JsonLdOptions/frameExpansion}} flag is set, <var>value</var>
-                        may be an empty <a>map</a>, or an <a>array</a> of one
+                        MAY be an empty <a>map</a>, or an <a>array</a> of one
                         or more <a>strings</a>.</span></li>
                     <li>Otherwise,
                       set <var>expanded value</var> to the result of using the
@@ -2513,7 +2513,7 @@
                       error has been detected and processing is aborted.
                       <span class="changed">
                         When the {{JsonLdOptions/frameExpansion}} flag is set, <var>value</var>
-                        may also be an empty <a>map</a>, or a <a>default object</a>
+                        MAY be an empty <a>map</a>, or a <a>default object</a>
                         where the value of `@default` is restricted to be
                         an <a>IRI</a>.
                         All other values mean that <a data-link-for="JsonLdErrorCode">invalid type value</a>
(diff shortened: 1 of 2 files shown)


gkellogg:
I may add further clarifications before pulling.

gkellogg:
@kasei, I attempted to address some of your concerns in this PR, regarding pseudo-steps in expansion involving the frameExpansion flag as well as initialize/set confusion.

From https://github.com/w3c/json-ld-api/pull/208/commits/3f18f02fcc6e20815f85e790b66c513aa5dffd52:

> Normalize initialize/set language. We use "initialize" for variables, the first time they're introduced (although, in some cases, they're introduced in parallel, and "set" is more natural). Use "set" for subsequent changes, or when describing how a value in some other structure is initialized or updated.

kasei:
@gkellogg these changes look like a great improvement. Thank you!

PR #223, opened 2019-11-22
Qualify that a potential prefix term must end with a gen-delim

### LINKED ISSUE #205: toRdf/0118 fails in JSON-LD 1.1
Since we added a provision in Create Term Definition that for simple terms to be used as prefixes, they need to end in a genDelim character, toRdf/0118 does not render the same results.

The relevant part of this test is the following:

```json
{
  "@context": {"term": "_:term", "termId": { "@id": "term", "@type": "@id" }},
  "termId": "term:AppendedToBlankNode"
}
```

This previously expanded to the following:

```json
[{
  "_:term": [{"@id": "_:termAppendedToBlankNode"}]
}]
```

Now it generates the following:

```json
[{
  "_:term": [{"@id": "term:AppendedToBlankNode"}]
}]
```

This is because the new test does not use "term" as a prefix, so it's treated as a scheme.

We could either update the algorithm to allow blank node terms to be used as prefixes, or make this a 1.0-only test. As the general trend for us is to deprecate blank node terms, this might be the way to go. Note that a workaround would be to make it an expanded term definition with `@prefix` set to true. But, this might affect contexts such as that used by Activity Streams, so the safest thing would be to update 16.6 as follows:

> If _term_ contains neither a colon (`:`) nor a slash (`/`), _simple term_ is `true`, and if the IRI mapping of _definition_ ends with a URI gen-delim character, **or is a blank node identifier**, set the prefix flag in _definition_ to `true`.

We could add another note about the use of blank node properties as being obsolete.

gkellogg:
charac…ter, **if it is an IRI**.

Fixes #205.

cc/ @kasei.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/223.html" title="Last updated on Nov 25, 2019, 9:42 PM UTC (21ed62e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/223/b92159a...21ed62e.html" title="Last updated on Nov 25, 2019, 9:42 PM UTC (21ed62e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1650,7 +1650,7 @@
             <li class="changed">If <var>term</var> contains neither a colon (`:`) nor a slash (`/`),
               <span class="changed"><var>simple term</var> is <code>true</code></span>, and if the
               <a>IRI mapping</a> of <var>definition</var> ends with a URI
-              <a data-cite="RFC3986#section-2.2">gen-delim</a> character,
+              <a data-cite="RFC3986#section-2.2">gen-delim</a> character, if it is an <a>IRI</a>,
               set the <a>prefix flag</a> in <var>definition</var> to <code>true</code>.</li>
           </ol>
         </li>
--- a/index.html
+++ b/index.html
@@ -1648,9 +1648,10 @@
               an <a data-link-for="JsonLdErrorCode">invalid IRI mapping</a>
               error has been detected and processing is aborted.</li>
             <li class="changed">If <var>term</var> contains neither a colon (`:`) nor a slash (`/`),
-              <span class="changed"><var>simple term</var> is <code>true</code></span>, and if the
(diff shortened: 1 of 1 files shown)


gkellogg:
> I think that "is an IRI that ends with a gen-delim character" would be easier to read...

That wording doesn't quite work, as it can either be a blank node, or an IRI which ends in a gen-delim, which is embedded in other clauses (unless I misinterpreted the resolution). Perhaps the following:

> If _term_ contains neither a colon (`:`) nor a slash (`/`), _simple term_ is `true`, and if the IRI mapping of _definition_ is either an IRI that ends with a URI gen-delim character or a blank node label, set the prefix flag in _definition_ to `true`.

PR #234, opened 2019-12-06
Fix typos in tests/manifest.html and template.haml

fsteeg:
Just two small typo fixes.


fsteeg:
Marked as non substantive for IPR from ash-nazg.

Issue #237, opened 2019-12-09
Add test for JSON-LD inside a JSON literal.


davidlehn:
Add test, or tests, to ensure JSON literals are not processed as JSON-LD.
This should at least be an expand test that ensures the JSON-LD-in-JSON stays unexpanded.
This was an issue in the jsonld.js implementation.

davidlehn:
Expand #js08 already exists.  @dlongley What additional tests, if any, would be useful here?  Something with a `@context`?

dlongley:
Yes, we should add something with a `@context` that uses a URL. jsonld.js would have previously erroneously resolved it and replaced it inline with the context document.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-12-13-json-ld.html#section4-2).

- `RESOLVED:  Accept test for JSON-LD embedded in JSON to test suite (no editorial action)`
<details><summary><i class="differentiate">View the transcript</i></summary>
JSON-LD featuring JSON featuring JSON-LD test<br/>
Rob Sanderson: <a href="https://github.com/w3c/json-ld-api/issues/237">https://github.com/w3c/json-ld-api/issues/237</a><br/>
David I. Lehn: a literal JSON can contain JSON-LD (e.g. a <code>@context</code> key)<br/>
… we need to make sure that processors do not try to process that<br/>
Rob Sanderson: this is only editorial; the text already says “do not touch this”<br/>
Proposed resolution: Accept test for JSON-LD embedded in JSON to test suite (no editorial action) (Rob Sanderson)<br/>
Rob Sanderson: +1<br/>
David I. Lehn: how should we process such issues?<br/>
Ivan Herman: +1<br/>
Tim Cole: +1<br/>
Ivan Herman: I don’t think we should use meeting time for this<br/>
Harold Solbrig: +1<br/>
David I. Lehn: +1<br/>
Pierre-Antoine Champin: +1<br/>
Resolution #3: Accept test for JSON-LD embedded in JSON to test suite (no editorial action)<br/>
Gregg Kellogg: +1<br/>
Rob Sanderson: but we need to point to a resolution in the PR solving the issue<br/>
</details>

davidlehn:
@gkellogg I don't think the requested test exists yet?  Something with a `@context` in the JSON as @dlongley mentioned above.

gkellogg:
[...]

PR #238, opened 2019-12-10
Add JSON literal alias tests.

davidlehn:
Tests for an implementation issue found in jsonld.js:
https://github.com/digitalbazaar/jsonld.js/pull/346

### DIFF
--- a/tests/expand/js14-in.jsonld
+++ b/tests/expand/js14-in.jsonld
@@ -1,0 +1,12 @@
+{
+  "@context": {
+    "@version": 1.1,
+    "value": "@value"
+  },
+  "ex:foo": {
+    "@type": "@json",
+    "value": {
+      "test": 1
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


tests/expand/js14-in.jsonld
```diff
@@ -0,0 +1,12 @@
+{
+  "@context": {
+    "@version": 1.1,
```
dlongley: Do we need an explicit version here?

davidlehn: I was matching the style of the other js* tests.

gkellogg:
Technically, it’s a 1.1 test, so `@version` is appropriate, but no longer necessary.

gkellogg:
Also, expansion tests should be replicated in toRdf, as @rubensworks does not go through an intermediate expansion step; this has been the pattern for other tests.

gkellogg:
For #237

PR #240, opened 2019-12-10
Add toRdf JSON literal alias tests.

davidlehn:
This goes along with https://github.com/w3c/json-ld-api/pull/238.
Note that the added `js#` ids do not line up with expand tests due to some extra tests in the toRdf namespace.  We might need to realign and/or sync them to avoid confusion.


gkellogg:
For #237.

Issue #241, opened 2019-12-11
Branching logic issue in Create Term Definition step 16


kasei:
[Create Term Definition](https://w3c.github.io/json-ld-api/#create-term-definition) step 16.1 uses passive language that doesn't seem to have any action expressed:

> If value contains the @id entry is null, the term is not used for IRI expansion, but is retained to be able to detect future redefinitions of this term.

If no action is taken as a result of this step, step 16.4 will call IRI Expansion with a null value and raise an Invalid IRI mapping error. I don't believe this is the intention of step 16, and ask that it be clarified.

I believe the intended set of steps is something like this:

```
16. If value contains the entry @id and its value does not equal term:
    16.1. If value contains the @id entry is null, the term is not used for IRI expansion, but is retained to be able to detect future redefinitions of this term.
    16.2. Otherwise:
        16.2.1: [old 16.2, without "Otherwise" prefix] If the value associated with the @id entry is not a string, an invalid IRI mapping error has been detected and processing is aborted.
        16.2.2: [old 16.3]
        16.2.3: [old 16.4]
        16.2.4: [old 16.5]
        16.2.5: [old 16.6]
```

gkellogg:
If the value is `null`, the remaining sub-steps in 16.1 should be skipped.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2019/2019-12-13-json-ld.html#section4-5).

- `RESOLVED:  Add to 16.1 that the following substeps should be skipped if the test for null is true`
<details><summary><i class="differentiate">View the transcript</i></summary>
Create Term Definition 16.1 needs more indents<br/>
Rob Sanderson: <a href="https://github.com/w3c/json-ld-api/issues/241">https://github.com/w3c/json-ld-api/issues/241</a><br/>
Gregg Kellogg: we added some text in 16.1 to retain the information that <code>@id</code> has value null<br/>
Proposed resolution: Add to 16.1 that the following substeps should be skipped (Rob Sanderson)<br/>
Gregg Kellogg: in my code, the following substeps are skipped<br/>
Proposed resolution: Add to 16.1 that the following substeps should be skipped if the test for null is true (Rob Sanderson)<br/>
Gregg Kellogg: +1<br/>
Rob Sanderson: +1<br/>
Pierre-Antoine Champin: +1<br/>
Harold Solbrig: +1<br/>
Tim Cole: +1<br/>
[...]

PR #247, opened 2019-12-17
Clarify 16.1 in Create Term Definition to satisfy #241.

gkellogg:
@kasei, please let us know if this satisfies issue #241.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/247.html" title="Last updated on Dec 18, 2019, 6:04 PM UTC (0255501)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/247/dd3dc99...0255501.html" title="Last updated on Dec 18, 2019, 6:04 PM UTC (0255501)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6889,7 +6889,7 @@
   </ul>
 </section>
 <section class="appendix informative" id="changes-from-cr">
-  <h2>Changes since Candidate Release of 12 November 2019</h2>
+  <h2>Changes since Candidate Release of 12 December 2019</h2>
   <ul>
     <li>Clarified step <a href="#ctd-id-null">16.1</a> of the
       <a href="#create-term-definition">Create Term Definition algorithm</a>
--- a/index.html
+++ b/index.html
@@ -6888,7 +6888,7 @@
     <li>Ignore terms and IRIs that have the form of a keyword (`"@"1*ALPHA`).</li>
   </ul>
 </section>
-<section class="appendix informative" id="changes-from-cg">
+<section class="appendix informative" id="changes-from-cr">
   <h2>Changes since Candidate Release of 12 November 2019</h2>
   <ul>
     <li>Clarified step <a href="#ctd-id-null">16.1</a> of the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -6887,6 +6888,15 @@ <h4>JsonLdErrorCode</h4>
     <li>Ignore terms and IRIs that have the form of a keyword (`"@"1*ALPHA`).</li>
   </ul>
 </section>
+<section class="appendix informative" id="changes-from-cr">
+  <h2>Changes since Candidate Release of 12 November 2019</h2>
```
iherman: Should be 12 December :-)

gkellogg:
If we just do editorial updates, then I think we can just track cumulative changes since the first CR. Of course, we can decide to do something differently in the future, but breaking it up into sub-appendices for each interim update might not be helpful.

iherman:
> If we just do editorial updates

Hopefully, that will be the case:-)

We shall see. As I said, it is a minor issue.

kasei:
I would have preferred a branching style as suggested in #241, and used throughout every algorithm in the API document. This is the only place in the entire document that talks about "skipping" "sub-steps". If there is a reason this language was chosen, I can live with it, but it sticks out as being different (and less explicit) than semantically similar steps used elsewhere in the document.

gkellogg:
I was looking to be less disruptive to the algorithmic flow, because this is intended to be a simple edit, rather than a substantial change to the algorithm. I was concerned that grouping the rest of the steps under an encompassing "Otherwise" might look too heavyweight; otherwise, that's what I would normally have done.

But, I'm willing to reconsider pending input from the WG.

davidlehn:
FWIW, I had the same thought as @kasei when I first saw the patch.  General pattern elsewhere is "if(cond) { steps }", but this is more like "if(cond) { return }".  That's fine in code, but a bit odd in prose.

PR #263, opened 2019-12-19
Add test for JSON literal with a context.

gkellogg:
Tests that no processing is done on a `@context`.

This is a test for https://github.com/w3c/json-ld-api/issues/237.

### DIFF
--- a/tests/expand-manifest.html
+++ b/tests/expand-manifest.html
@@ -6472,6 +6472,34 @@ Test tjs14 Expand JSON literal with aliased @value
 </dd>
 </dl>
 </dd>
+<dt id='tjs21'>
+Test tjs21 Expand JSON literal with @context
+</dt>
+<dd>
+<dl class='entry'>
+<dt>id</dt>
+<dd>#tjs21</dd>
+<dt>Type</dt>
+<dd>jld:PositiveEvaluationTest, jld:ExpandTest</dd>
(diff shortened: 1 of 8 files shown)


gkellogg:
@davidlehn the test isn't very useful for RDF, as a value object generates no triples.

Also, I serialize the 333333333.33333329 differently than you have, I need to check against the algorithm.

gkellogg:
Note that that number is an example in https://tools.ietf.org/html/draft-rundgren-json-canonicalization-scheme-05#section-3.2.2, which should serialize as 333333333.3333333.

That spec's updated from -05 to -16 since, with some minor changes, but we reference the -05 version.

I'll update both tests based on my output.

gkellogg:
Actually, the expansion algorithm doesn't run canonicalization, so it may be due to differences in native Ruby number conversion.

gkellogg:
Never mind, this seems to be related to a previous PR, as I'm seeing issues with js12, not js21

davidlehn:
These comments look related to https://github.com/w3c/json-ld-api/pull/256.  Maybe chat over there.

Re test not useful for RDF, which one?

Re the numbers, perhaps some of those toRdf canonicalization tests I moved over to the 'expand' tests shouldn't have been?

gkellogg:
I'll delete the comments as not relevant; the PR has been merged.

PR #266, opened 2019-12-20
Improve fix to #241

pchampin:
Building on @gkellogg's PR #247 to address issue #241, I made a further change that's hopefuly not too disruptive, and should address @kasei's [comment](https://github.com/w3c/json-ld-api/pull/247#issuecomment-567258296).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/266.html" title="Last updated on Dec 20, 2019, 11:26 PM UTC (1bb3603)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/266/0319468...1bb3603.html" title="Last updated on Dec 20, 2019, 11:26 PM UTC (1bb3603)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1616,14 +1616,12 @@
         </li>
         <li>Set the <a>reverse property</a> flag of <var>definition</var>
           to <code>false</code>.</li>
-        <li>If <var>value</var> contains the <a>entry</a> <code>@id</code> and its value
-          does not equal <var>term</var>:
+        <li class="changed">If <var>value</var> contains the <a>entry</a> <code>@id</code> and its value
+          neither equals <var>term</var> nor is <code>null</code>
+          <span id="ctd-id-null" class="note">If `null`, the term is not used for IRI expansion, but is
+            retained to be able to detect future redefinitions of this term.</span>:
           <ol>
-            <li id="ctd-id-null" class="changed">If the value associated with the `@id` <a>entry</a>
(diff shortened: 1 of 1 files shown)


pchampin:
Forgot to mention: the first commit in this PR is fixing a plain typo. So it should be retained even if we don't keep this PR.

index.html
```diff
@@ -1616,14 +1616,12 @@ <h3>Algorithm</h3>
         </li>
         <li>Set the <a>reverse property</a> flag of <var>definition</var>
           to <code>false</code>.</li>
-        <li>If <var>value</var> contains the <a>entry</a> <code>@id</code> and its value
-          does not equal <var>term</var>:
+        <li class="changed">If <var>value</var> contains the <a>entry</a> <code>@id</code> and its value
+          neither equals <var>term</var> nor is <code>null</code>
```
davidlehn: Maybe this should end with a ":"?  Otherwise it runs into the inline "note" below.  It'll look odd either way.

Related is that all the other inline "notes" also look odd compared to the full "NOTE" blocks.  Were they supposed to be inline?

gkellogg: ReSpec supports both, and sometimes it makes sense to do. Full block notes are pretty heavyweight.

davidlehn: I think this still needs something done.  It reads "... nor is null If null, the term ...".  The inline note style is there for the "If null...", but the final rendered result could be improved.

PR #273, opened 2019-12-23
Change inline note to block note in Create Term Definition step 16.

gkellogg:
For #241.

@davidlehn and @kasei, I change the inline note to a block note, but I don't think it improves it. Please suggest any other changes you think would improve both formatting and clarity.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/273.html" title="Last updated on Dec 23, 2019, 8:09 PM UTC (12c9e21)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/273/3e9626a...12c9e21.html" title="Last updated on Dec 23, 2019, 8:09 PM UTC (12c9e21)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1620,7 +1620,8 @@
           to <code>false</code>.</li>
         <li class="changed">If <var>value</var> contains the <a>entry</a> <code>@id</code> and its value
           neither equals <var>term</var> nor is <code>null</code>:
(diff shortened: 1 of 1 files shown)


kasei:
@gkellogg I do think this is an improvement. I think I would also update the note text (as it is more free-standing now) to begin "If the '@id' entry in value is null..."

Issue #270, opened 2019-12-20
Possible bug in Expansion Algorithm step 13.4.16


kasei:
[Expansion Algorithm](https://w3c.github.io/json-ld-api/#expansion-algorithm) step 13.4.16 says:

> Unless expanded value is null, and expanded property is `@value`, and input type is `@json`, set the expanded property entry of result to expanded value.

I believe test [t0122](https://w3c.github.io/json-ld-api/tests/expand-manifest.html#t0122) highlights a case where this is problematic. It's possible by way of 13.4.3.2 and (IRI Expansion step 2) that `expanded value` is null here, but the test expects the resulting value to be an empty string. I'm not sure where else the `null` value might be converted to an empty string. Is it possible the test is wrong? Or does the wording in 13.4.16 need updating?

gkellogg:
You might be referring to [compact#tjs11](https://w3c.github.io/json-ld-api/tests/compact-manifest.html#tjs11), which doesn't have a related expansion test. [t0122](https://w3c.github.io/json-ld-api/tests/expand-manifest.html#t0122) is about catching IRIs having the form of a keyword. I'll add a couple of tests for this.

And, that test should read, instead:

> Unless _expanded value_ is `null`, and _expanded property_ is `@value`, and _input type_ is **not**  `@json`, set the _expanded property_ entry of _result_ to _expanded value_.

But, this step overlaps a change in PR #275.

kasei:
I'm not doing anything related to compaction, so I doubt there's any confusion with tjs11.

gkellogg:
Can you describe what it is about t0122 that relates to this processing step?

gkellogg:
To be clear, [t0122](https://w3c.github.io/json-ld-api/tests/expand-manifest.html#t0122) contains the following input file:

```json
{
  "@context": {
    "@base": "http://example.org/"
  },
  "http://example.org/vocab/at": {"@id": "@"},
  "http://example.org/vocab/foo.bar": {"@id": "@foo.bar"},
  "http://example.org/vocab/ignoreme": {"@id": "@ignoreMe"}
}
```

which doesn't seem to relate to JSON literals. One test I was going to add uses the following input file:

```json
{
  "@context": {
    "@version": 1.1,
    "e": {"@id": "http://example.org/vocab#null", "@type": "@json"}
  },
  "e": null
}
```

I think this would trigger the issue you were concerned about.

kasei:
@gkellogg 

[...]

PR #283, opened 2019-12-24
Adds tests for null json literals, and fixes one expected result

gkellogg:
Note that this identifies an issue with the expected result for expand/0122, which was inconsistent with the spec.

For #270.


davidlehn:
- I think new js22 is same as js18?
- These should have toRdf tests too.
- Just noticed jsonld.js fails with an empty array `@json` value too. Could maybe just change js22 to that. (Or this could be filed as another issue.)

Issue #332, opened 2020-01-17
Revisit wording on Compaction 12.8


gkellogg:
There is some confusion over what "the value" means in 12.8.7.4.5 (see https://github.com/w3c/json-ld-api/issues/323#issuecomment-575702804), and in general this set of steps is confusing and inconsistent. Given that we are now bound to re-enter CR, this is an opportunity to clean this section up, extracting common terms, and making steps clear and consistent.

gkellogg:
This is an opportunity to create some general definitions for things like "compact the item" that can be used as "macros" to reduce the boilerplate where we perform IRI compaction with common parameters..

Issue #329, opened 2020-01-17
Ambiguous language in Compaction 12.8.6.2.2


kasei:
Compaction step 12.8.6.2.2:

> If `expanded item` contains the entry `@index`, then add an entry to `compacted item` where the key is the result of the IRI Compaction algorithm, passing `active context`, `inverse context`, `@index` as var, and the value associated with the `@index` entry in `expanded item` as value.

This text is confusing as `value` can be interpreted in two ways:

* As the value of the entry being set in `compacted item`
* As an optional argument to the IRI Compaction Algorithm

gkellogg:
As part of #322, I'll create algorithmic terminology to reduce the boilerplate text for calling IRI Compaction algorithm to something like the following:

> If expanded item contains the entry @index, then add an entry to compacted item where the key is the result _iri compacting_ `@index`, and the value is value of the `@index` entry in _expanded item_.

Then _iri compacting_ can be defined as an invocation of the IRI Compaction Algorithm using _active context_ (or some other explicit provided context), _inverse context_, and `vocab: true` with a provided argument.

The pattern for adding entries is to specify the _key_ and _value_ of the entry, so in this case the first _value_ refers to that new entry.

Do you think there is some standardized markup that would clarify adding such entries?

This should clean up many of these confusing statements where you can get lost in the boilerplate.

gkellogg:
Updated in PR #339.

kasei:
@gkellogg #339 looks good for this. Thanks.

Issue #327, opened 2020-01-17
Unclear language in Compaction Algorithm


kasei:
[Compaction](https://w3c.github.io/json-ld-api/#algorithm-6) step 12.8.7.4 says:

> Otherwise, `container` does not include `@graph` or otherwise does not match one of the previous cases, redo `compacted item`.

The language here is unclear:

* Is this meant to be an "else if" (e.g. there is a missing "if" after "Otherwise,")?
* Is the "otherwise does not match one of the previous cases" language just a restating of the initial "Otherwise"?
* It's not clear what "redo `compacted item`" means here

gkellogg:
The remainder after "Otherwise" is simply a statement of the conditions that led to this point, and that the following steps will redo the compacted item. It could be made a parenthetical, if you feel that would improve it, or made into a note.

gkellogg:
This is addressed in PR #339.

kasei:
@gkellogg #339 looks good for this. Thanks.

PR #339, opened 2020-01-19
Reduce boilerplate language by introducing macros.

### LINKED ISSUE #330: Imprecise language referring to variable in Compaction Algorithm
[Compaction](https://w3c.github.io/json-ld-api/#algorithm-6) step 12.8.9.1 says:

> If `compactArrays` is false, `as array` is true and `compacted item` is not an array, set it to a new array containing only `compacted item`.

Assuming "it" here is in reference to `compacted item`, it would be much clearer to state this explicitly.

gkellogg:
This is an attempt to simplify language by extracting common steps into "macros" described as algorithm terms.

* Be more clear about what it meant by "it" in "set it" in various parts of the text. Fixes #330.
* Added "add value" macro to algorithm terms to manage adding values entries in objects and properly deal with array conversion.
* Added _iri compacting_ macro and use where more verbose IRI Compaction language was used.
* Added _iri expanding_ macro and use where more verbose IRI Expansion language was used.
* Rewrite many entries in Compaction and elsewhere to use "add value" to simplify language.

For #327, #329, #330, and #332.

cc/ @kasei

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/339.html" title="Last updated on Jan 21, 2020, 9:56 PM UTC (191cab9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/339/16a8156...191cab9.html" title="Last updated on Jan 21, 2020, 9:56 PM UTC (191cab9)">Diff</a>

### DIFF
--- a/common/algorithm-terms.html
+++ b/common/algorithm-terms.html
@@ -43,12 +43,13 @@
   <dt><dfn data-cite="JSON-LD11-API#dfn-iri-compacting">iri compacting</dfn></dt><dd class="algorithm">
     Used as a macro within various algorithms as to reduce the language used to describe
     the process of compacting a <a>string</a> <var>var</var> representing an <a>IRI</a> or <a>keyword</a>
-    relative to an <var>active context</var> either specified directly, or coming from the scope of
+    using an <var>active context</var> either specified directly, or coming from the scope of
     the algorithm step using this term and <var>inverse context</var> also taken
     from the scope of the algorithm step using this term.
     An optional <var>value</var> is used, if explicitly provided.
-    Unless specified, the <var>vocab</var>
-    flag defaults to `true`, and the <var>reverse</var> flag defaults to `false`.
+    Unless specified,
+    the <var>vocab</var> flag defaults to `true`,
+    and the <var>reverse</var> flag defaults to `false`.
     <ol>
       <li>Return the result of using the <a href="#iri-compaction">IRI Compaction algorithm</a>,
         passing <var>active context</var>, <var>inverse context</var>,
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1381,12 +1381,9 @@ <h3>Algorithm</h3>
...
                   of <var>result</var> is set to
-                  <span class="changed">the result of using the 
-                    <a href="#iri-expansion">IRI Expansion algorithm</a>,
-                    passing <var>result</var> for <var>active context</var>,
-                    <var>value</var>,
-                    <code>true</code> for <var>vocab</var>, and
-                    <code>true</code> for <var>document relative</var>.
+                  <span class="changed">the result of
+                    <a>iri expanding</a> <var>value</var>
+                    using <code>true</code> for <var>document relative</var>.
```
davidlehn: Double period. I think this is the one to remove.
```suggestion
                    using <code>true</code> for <var>document relative</var>
```

common/algorithm-terms.html
```diff
@@ -20,6 +40,44 @@
     used internally to help determine if object embedding is appropriate,</span>
     the <a>explicit inclusion flag</a>,
     and the <a>omit default flag</a>.</dd>
+  <dt><dfn data-cite="JSON-LD11-API#dfn-iri-compacting">iri compacting</dfn></dt><dd class="algorithm">
```
davidlehn: Use uppercase "IRI" instead of lowercase "iri" for both "IRI compacting" and "IRI expanding".

kasei:
@gkellogg I see this has already been merged, but I'd like to mention that it feels very strange for the macros to have opposite default values for the `vocab` parameter than the underlying algorithms.

gkellogg:
@kasei please cite an example of where you find the language strange.

kasei:
[IRI Compaction](https://w3c.github.io/json-ld-api/#algorithm-7):

> If not passed, `value` is set to `null` and both `vocab` and `reverse` are both set to `false`.

Compared to [IRI Compacting](https://w3c.github.io/json-ld-api/#dfn-iri-compacting):

> Unless specified, the `vocab` flag defaults to `true`, and the `reverse` flag defaults to `false`.

Similarly for Expansion.

gkellogg:
Okay, I understand. The algorithms take false default values, but the most common usage is for `vocab` to be set to `true`. The whole point is to reduce boilerplate, so the macro uses `true`. It’s generally (but not exclusively) useful for flags to default to `false` in algorithm definitions.

kasei:
Right. And nothing about the current state is wrong, it just stood out to me as being rather strange to have two different default values depending on if you call the algorithm directly, or use the macro shortcut.

PR #340, opened 2020-01-20
Add an embedded context which is unused, but should indicate an error.

### LINKED ISSUE #331: Possible context syntax error tests
How deeply should contexts be checked, even if not used?  I was noticing the odd behavior in jsonld.js 2.0.1 of the following.

Missing t1 and t2 ids, no data, syntax error `"@context terms must define an @id."`:
```json
{
  "@context": {
    "@version": 1.1,
    "t1": {
      "@context": {
        "t2": {
          "@context": {
            "type": null
          }
        }
      }
    }
  }
}
```

Missing a t2 id, no data, works:
```json
{
  "@context": {
    "@version": 1.1,
    "t1": {
      "@id": "ex:t1",
      "@context": {
        "t2": {
          "@context": {
            "type": null
          }
        }
      }
    }
  }
}
```

Missing t2 id, t1 used, syntax error `"@context terms must define an @id."`:
```json
{
  "@context": {
    "@version": 1.1,
    "t1": {
      "@id": "ex:t1",
      "@context": {
        "t2": {
          "@context": {
            "type": null
          }
        }
      }
    }
  },
  "t1": {"t2": "x"}
}
```

Has t1 and t2 ids, works with or without data:
```json
{
  "@context": {
    "@version": 1.1,
    "t1": {
      "@id": "ex:t1",
      "@context": {
        "t2": {
          "@id": "ex:t2",
          "@context": {
            "type": null
          }
        }
      }
    }
  },
  "t1": {"t2": "x"}
}
```

I'm guessing this is exposing quirks of jsonld.js. Should it always be checking for those syntax errors in the context, regardless of the data?  Looks like some test coverage may be needed for this.

gkellogg:
Fixes #331.

### DIFF
--- a/tests/expand-manifest.html
+++ b/tests/expand-manifest.html
@@ -3896,6 +3896,34 @@ Test tc031 @context resolutions respects relative URLs.
 </dd>
 </dl>
 </dd>
+<dt id='tc032'>
+Test tc032 Unused embedded context with error.
+</dt>
+<dd>
+<dl class='entry'>
+<dt>id</dt>
+<dd>#tc032</dd>
+<dt>Type</dt>
+<dd>jld:NegativeEvaluationTest, jld:ExpandTest</dd>
+<dt>Purpose</dt>
+<dd>An embedded context which is never used should still be checked.</dd>
+<dt>input</dt>
+<dd>
+<a href='expand/c032-in.jsonld'>expand/c032-in.jsonld</a>
+</dd>
+<dt>expect</dt>
+<dd>
+invalid scoped context
+</dd>
+<dt>Options</dt>
+<dd>
+<dl class='options'>
+<dt>specVersion</dt>
+<dd>json-ld-1.1</dd>
+</dl>
+</dd>
+</dl>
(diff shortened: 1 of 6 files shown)


davidlehn:
You have those marked as "invalid scoped context".  jsonld.js was throwing "invalid IRI mapping" since the inner part didn't have an `@id`.  I'm not sure which is more correct.

These tests don't catch the specific bug in jsonld.js.  Need a similar test that has the first level "t1" id, no second level "t2" id, and **no** data.  (The second example in #331.)  That should fail since the inner context is invalid even though no context data is being used, but jsonld.js will pass.

gkellogg:
23.3 says the following:

> Invoke the Context Processing algorithm using the _active context_, _context_ as _local context_, and `true` for _override protected_. **If any error is detected, an "invalid scoped context" error has been detected and processing is aborted.

So, while the context processing may through "invalid IRI mapping", it should be caught in this step to generate "invalid scoped context".

Please add any additional tests you think are necessary.

PR #341, opened 2020-01-20
Simplify language for returning an array if there is a single value

### LINKED ISSUE #334: Ambiguous conditional language used in Compaction 3.3
[Compaction](https://w3c.github.io/json-ld-api/#algorithm-6) step 3.3 says:

> If `result` contains only one item (it has a length of 1), `active property` is not `@graph` or `@set`, or the container mapping for `active property` in `active context` does not include `@list` or `@set`, and `compactArrays` is `true`, set `result` to its only item.

I find the language here to be indecipherable. Based on some of the tests, my best guess is that the intention is something like:

```
If
	(
		(
			(`result` contains only one item (it has a length of 1))
			AND
			(`active property` is neither `@graph` nor `@set`)
		)
		OR
		(
			(the container mapping for `active property` in `active context` does not include `@list` or `@set`)
			AND
			(`compactArrays` is `true`)
		)
	)
```

Is that the intended semantics? Either way, could this language be clarified?

gkellogg:
in the Compaction algorithm.

Fixes #334.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/341.html" title="Last updated on Jan 21, 2020, 9:53 PM UTC (e6f944f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/341/b4c20a4...e6f944f.html" title="Last updated on Jan 21, 2020, 9:53 PM UTC (e6f944f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3417,16 +3417,13 @@
                   it to <var>result</var>.</li>
               </ol>
             </li>
-            <li>If <var>result</var> contains only one item (it has a length of
-              <code>1</code>),
-              <span class="changed">
-                <var>active property</var> is not <code>@graph</code> or <code>@set</code>,
-                or the <a>container mapping</a> for <var>active property</var> in
-                <var>active context</var> does not include <code>@list</code> or <code>@set</code>,
-              </span>
-              and {{JsonLdOptions/compactArrays}}
-              is <code>true</code>, set <var>result</var> to its only item.</li>
-            <li>Return <var>result</var>.</li>
+            <li>If <var>result</var> is empty or contains more than one value,
+              or {{JsonLdOptions/compactArrays}} is `false`,
+              or <span class="changed">or <var>active property</var> is either `@graph` or `@set`,
+              or <a>container mapping</a> for <var>active property</var> in
+              <var>active context</var> includes either <code>@list</code> or <code>@set</code></span>,
+              return <var>result</var>.</li>
+            <li>Otherwise, return the first value in <var>result</var>.</li>
           </ol>
         </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3417,16 +3417,13 @@ <h3>Algorithm</h3>
...
-                <var>active property</var> is not <code>@graph</code> or <code>@set</code>,
-                or the <a>container mapping</a> for <var>active property</var> in
-                <var>active context</var> does not include <code>@list</code> or <code>@set</code>,
-              </span>
-              and {{JsonLdOptions/compactArrays}}
-              is <code>true</code>, set <var>result</var> to its only item.</li>
-            <li>Return <var>result</var>.</li>
+            <li>If <var>result</var> is empty or contains more than one value,
+              or {{JsonLdOptions/compactArrays}} is `false`,
+              or <span class="changed">or <var>active property</var> is either `@graph` or `@set`,
```
davidlehn: Double "or".

index.html
```diff
@@ -3417,16 +3417,13 @@ <h3>Algorithm</h3>
...
+              or <a>container mapping</a> for <var>active property</var> in
+              <var>active context</var> includes either <code>@list</code> or <code>@set</code></span>,
+              return <var>result</var>.</li>
+            <li>Otherwise, return the first value in <var>result</var>.</li>
```
davidlehn: Seems odd to say "the first value" when, I think, you would only get here when you have one item in result.

gkellogg: would "return the value in _result_" scan better?

PR #352, opened 2020-01-23
Various

### LINKED ISSUE #345: Missing definition of "empty list object" in IRI Compaction
[IRI Compaction](https://w3c.github.io/json-ld-api/#algorithm-7) step 2.17 says:

> Otherwise, append `type/language value` and `@none`, in that order, to `preferred values`. If `value` is an empty list object, set `type/language` to @any.

I don't believe "empty list object" is defined anywhere.

### LINKED ISSUE #346: Confusing language about normalizing strings in IRI Compaction
[IRI Compaction](https://w3c.github.io/json-ld-api/#algorithm-7) step 2.9.1.2 says:

> Otherwise, if `value` contains an `@language` entry and does not contain an `@index` entry, then set `type/language value` to its associated value and, append `@language`, normalized to lower case, and `@language@set` to `containers`.

"associated value" presumably means the value of the `@language` entry?

I do not think "append `@language`, normalized to lower case" make sense. Is this a typo?

### LINKED ISSUE #347: Missing variable markup in IRI Compaction
[IRI Compaction](https://w3c.github.io/json-ld-api/#algorithm-7) 2.11 is missing the markup to indicate that `value` is a variable.

### LINKED ISSUE #348: IRI Compaction possibly missing type checks when using value as a map
[IRI Compaction](https://w3c.github.io/json-ld-api/#algorithm-7) step 2.11 says:

> If `processing mode` not `json-ld-1.0` and `value` does not contain an `@index` entry, append `@index` and `@index@set` to `containers`.

(I'm assuming "`processing mode` not `json-ld-1.0`" is missing an "is": "*is* not `json-ld-1.0`"?)

Steps 2.2 and 2.5 seem to condition entry checking on `value` being a map ("If value is a map containing the entry..."). Should 2.11 be doing this as well?

### LINKED ISSUE #349: Incorrect markup for variables in IRI Compaction step 5
[IRI Compaction](https://w3c.github.io/json-ld-api/#algorithm-7) step 5 uses markup that conflates variable names and "JSON-LD Specific Term Definitions" for the variables `term` and `term definition`.

### LINKED ISSUE #350: Does input value to Value Compaction needs copying?
[Value Compaction](https://w3c.github.io/json-ld-api/#algorithm-8) step 1 says:

> Initialize `result` to `value`.

The algorithm later goes on to mutate `result` (as in step 9). However, other places in the spec that do this sort of modification on a value that's been passed in will "copy" or "clone" the input value (as in Context Processing step 1 and Create Term Definition step 3). Should the initialization in Value Compaction use similar language?

I think this is important, because there seem to be places where Value Compaction is called on a value not expecting the value to be modified before later processing (as in Compaction step 7).

### LINKED ISSUE #351: Handling of @value-only maps in Value Compaction
The [Value Compaction Overview](https://w3c.github.io/json-ld-api/#overview-8) says:

> [Compacting `value` just into the value associated with the `@value` entry] can also be done if `@value` is the only entry in `value` (apart an `@index` entry in case the `container mapping` of `active property` includes `@index`) and either its associated value is not a string, there is no default language, or there is an explicit `null` language mapping for the active property.

However, I can't see any place in the algorithm that handles the specific of a `value` with only a `@value` entry whose value is not a string. Step 7.1 seems to get close, but requires "an `@index` entry, and the `container mapping` associated to `active property` includes `@index`".

gkellogg:
* Change "empty list object" to "list object with an empty array as a value of `@list`. Fixes #345.
* Fix language in IRI compaction on setting type/language to the normalized value of `@language`. Fixes #346.
* Add missing variable markup for _value_ in IRI compaction 2.11. Fixes #347.
* Change some other uses of _value_ and **entry** in IRI compaction. Fixes #348.
* Cleanup language in step 5 of IRI Compaction. Fixes #349.
* Update step 1 in value compaction to set result to a copy of value. Fixes #350.
* Account for value compacting a value object which is not a string, and there is no `@index` entry. Fixes #351.
* Don't remove "container key" from "compacted item" for `@index` in Compaction, as it will have been left out when compacting.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/352.html" title="Last updated on Jan 27, 2020, 6:05 PM UTC (a8a4c10)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/352/581ae99...a8a4c10.html" title="Last updated on Jan 27, 2020, 6:05 PM UTC (a8a4c10)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4095,7 +4095,8 @@
             </li>
             <li>Otherwise, append <var>type/language value</var> and <code>@none</code>, in
               that order, to <var>preferred values</var>.
-              <span class="changed">If <var>value</var> is an empty <a>list object</a>,
+              <span class="changed">If <var>value</var> is a <a>list object</a>
+                with an empty <var>array</var> as the value of `@list`,
                 set <var>type/language</var> to <code>@any</code>.</span></li>
             <li class="changed">Append <code>@any</code> to <var>preferred values</var>.</li>
             <li class="changed">If <var>preferred values</var>
--- a/index.html
+++ b/index.html
@@ -5598,7 +5598,7 @@
 
     <p>It is important to highlight that implementations do not modify the input parameters.
       If an error is detected, the {{Promise}} is
-      rejected with a <a>JsonLdError</a> with an appropriate {{JsonLdError/code}}
+      rejected with a <a>JsonLdError</a> having an appropriate {{JsonLdError/code}}
       and processing is stopped.</p>
 
     <p>If the <a data-link-for="JsonLdOptions">documentLoader</a>
@@ -6189,7 +6189,7 @@
       <p>The <dfn>LoadDocumentCallback</dfn> defines a callback that custom document loaders
         have to implement to be used to retrieve remote documents and contexts.
         The callback returns a {{Promise}} resolving to a <a>RemoteDocument</a>.
-        On failure, the {{Promise}} is rejected with an appropriate error <a data-link-for="JsonLdError">code</a>.</p>
+        On failure, the {{Promise}} with a a <a>JsonLdError</a> having an appropriate error <a data-link-for="JsonLdError">code</a>.</p>
 
       <pre class="idl">
         callback LoadDocumentCallback = Promise&lt;RemoteDocument> (
--- a/index.html
+++ b/index.html
@@ -4061,7 +4061,7 @@
               be the last <a>container mapping</a> value to be checked as it
               is the most generic.</li>
             <li class="changed">
-              If <a>processing mode</a> not `json-ld-1.0` and value does not contain an <code>@index</code> <a>entry</a>,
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -5601,7 +5596,7 @@ <h3>Algorithm</h3>
 
     <p>It is important to highlight that implementations do not modify the input parameters.
       If an error is detected, the {{Promise}} is
-      rejected passing a <a>JsonLdError</a> with the corresponding error
+      rejected <a>JsonLdError</a> with the corresponding error
```
davidlehn: This doesn't look right.  Was "passing a" supposed to change to "with a" or similar?
(Is this related to the listed bugs?)

gkellogg: See 8fc32cb408839583c4832095228885c20338e4ea.

Issue #265, opened 2019-12-20
Problematic/confusing algorithm text re: the handling of URL @context values


kasei:
[Context Processing](https://w3c.github.io/json-ld-api/#context-processing-algorithm) step 5.2.1 says:

> 5.2.1) Initialize context to the result of resolving context against the base IRI of the document containing the local context which is established as specified in section 5.1 Establishing a Base URI of [RFC3986]. Only the basic algorithm in section 5.2 of [RFC3986] is used; neither Syntax-Based Normalization nor Scheme-Based Normalization are performed. Characters additionally allowed in IRI references are treated in the same way that unreserved characters are treated in URI references, per section 6.5 of [RFC3987].

This seems rather verbose, as "the document containing the `local context`" will already have a base IRI. However, there doesn't seem to be any mention of maintaining provenance of base IRIs of local contexts in this algorithm. I *believe* this is something that could be passed as an optional input variable (defaulting to the base IRI of the active context), but it would require cooperation at the call-sites of the context processing algorithm in the [expansion algorithm](https://w3c.github.io/json-ld-api/#expansion-algorithm) and [create term definition](https://w3c.github.io/json-ld-api/#create-term-definition).

Steps 5.2.3 through 5.2.5 say:

> 5.2.3) If context was previously dereferenced, then the processor MUST NOT do a further dereference, and context is set to the previously established internal representation.
> 5.2.4) Otherwise, dereference context using the LoadDocumentCallback, passing context for url, and http://www.w3.org/ns/json-ld#context for profile and for requestProfile.
> 5.2.5) If context cannot be dereferenced, or cannot be transformed into the internal representation, a loading remote context failed error has been detected and processing is aborted. If the dereferenced document has no top-level map with an `@context` entry, an invalid remote context has been detected and processing is aborted; otherwise, set context to the value of that entry.

These steps are confusing for a few reasons:

* 5.2.3 talks about a stored version of the "established internal representation" of the dereferenced context IRI, but this seems to be hidden state rather than being passed as an input variable
* There doesn't seem to be a reason to store the entire representation of dereferenced context documents instead of just the `@context` entry
* The error checking portion of 5.2.5 (which is the bulk of the text) seems to only be applicable if the "otherwise" branch of 5.2.4 was taken. Perhaps it should be 5.2.4.1, leaving the setting of `context` in 5.2.4 (with or without extracting the `@context` entry, depending on the previous bullet point)

gkellogg:
[...]

PR #356, opened 2020-01-27
Updates to pass base URL

gkellogg:
to Expansion, Context Processing and Create Term definition.

The change touches a lot of parts, and leaves a little hand waiving when a context has been previously loaded.

For #265.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/356.html" title="Last updated on Jan 29, 2020, 7:43 PM UTC (2614999)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/356/2f57a54...2614999.html" title="Last updated on Jan 29, 2020, 7:43 PM UTC (2614999)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1086,7 +1086,6 @@
       <li class="changed">an optional <dfn>base URL</dfn> (<a>IRI</a>),</li>
       <li class="changed">an optional <a>context</a> (<a>context</a>),</li>
       <li>an optional <dfn>container mapping</dfn> (<a>array</a> of <a>strings</a>),
-      <li class="changed">an optional <a>context</a> (<a>context</a>),</li>
       <li class="changed">an optional <dfn>direction mapping</dfn> (`"ltr"` or `"rtl"`),</li>
       <li class="changed">an optional <dfn>index mapping</dfn> (<a>string</a>),</li>
       <li>an optional <dfn>language mapping</dfn> (<a>string</a>),</li>
--- a/index.html
+++ b/index.html
@@ -1080,16 +1080,19 @@
     <p>Each <a>term definition</a> consists of:</p>
     <ul>
       <li>an <dfn>IRI mapping</dfn> (<a>IRI</a>),</li>
-      <li>a flag <dfn>reverse property</dfn> (<a>boolean</a>),</li>
-       <li>an optional <dfn>type mapping</dfn> (<a>IRI</a>),</li>
-       <li>an optional <dfn>language mapping</dfn> (<a>string</a>),</li>
-      <li class="changed">an optional <dfn>direction mapping</dfn> (`"ltr"` or `"rtl"`),</li>
-      <li class="changed">an optional <a>context</a> (<a>context</a>),</li>
-      <li class="changed">an optional <dfn>nest value</dfn> (<a>string</a>),</li>
-      <li class="changed">an optional <dfn>prefix flag</dfn> (<a>boolean</a>),</li>
-      <li class="changed">an optional <dfn>index mapping</dfn> (<a>string</a>),</li>
+      <li class="changed">a <dfn>prefix flag</dfn> (<a>boolean</a>),</li>
       <li class="changed">a <dfn>protected</dfn> flag (<a>boolean</a>),</li>
-      <li>and an optional <dfn>container mapping</dfn> (<a>array</a> of <a>strings</a>).</li>
+      <li>a <dfn>reverse property</dfn> flag (<a>boolean</a>),</li>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1262,25 +1243,39 @@ <h3>Algorithm</h3>
...
                   passing <var>result</var> for <var>active context</var>,
-                  <var>context</var> for <var>local context</var>, and <span class="changed">a copy of</span> <var>remote contexts</var>.</li>
+                  <var>loaded context</var> for <var>local context</var>,
+                  <span class="changed">
+                    the {{RemoteDocument/documentUrl}} of <var>context document</var> for <var>base URL</var>,
+                  </span>
+                  and <span class="changed">a copy of</span> <var>remote contexts</var>.
+                  <div class="note">If <var>context</var> was previously dereferenced,
+                    <a>processors</a> MUST make provisions for retaining the <var>base URL</var>
+                    of that <a>context</a> for this step.</div>
```
dlongley: Can we say why? (I presume to enable the resolution of any relative context URLs that may be encountered during processing).

```suggestion
                    of that <a>context</a> for this step to enable the resolution of any
                    relative context URLs that may be encountered during processing.</div>
```

index.html
```diff
@@ -1209,46 +1215,21 @@ <h3>Algorithm</h3>
                 <li>Otherwise, initialize <var>result</var> as a
                   newly-initialized <var>active context</var>,
                   <span class="changed">setting <a>previous context</a> in <var>result</var>
-                    to the previous value of <var>result</var> if <var>propagate</var> is <code>false</code></span>.
+                    to the previous value of <var>result</var> if <var>propagate</var> is <code>false</code></span>
+                    and <a>base IRI</a> to the value of the {{JsonLdOptions/base}} option
+                    of the <a href="#the-application-programming-interface" class="sectionRef"></a>,
+                    if provided, otherwise to <var>base URL</var>.
```
pchampin: So "global configuration" (the API parameter) takes precedence over "local configuration" (algorithm parameter) ? This might very well be the way to go, I don't have a clear enough view of the big picture right now... But I'm pointing it out because it is a bit counter-intuitive...

index.html
```diff
@@ -1080,16 +1080,18 @@ <h2>RDF Serialization/Deserialization</h2>
...
-      <li class="changed">an optional <dfn>direction mapping</dfn> (`"ltr"` or `"rtl"`),</li>
+      <li class="changed">a <dfn>prefix flag</dfn> (<a>boolean</a>),</li>
+      <li class="changed">a <dfn>protected</dfn> flag (<a>boolean</a>),</li>
+      <li>a <dfn>reverse property</dfn> flag (<a>boolean</a>),</li>
+      <li class="changed">an optional <dfn>base URL</dfn> (<a>IRI</a>),</li>
       <li class="changed">an optional <a>context</a> (<a>context</a>),</li>
-      <li class="changed">an optional <dfn>nest value</dfn> (<a>string</a>),</li>
-      <li class="changed">an optional <dfn>prefix flag</dfn> (<a>boolean</a>),</li>
+      <li>an optional <dfn>container mapping</dfn> (<a>array</a> of <a>strings</a>),
+      <li class="changed">an optional <a>context</a> (<a>context</a>),</li>
```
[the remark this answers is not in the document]
gkellogg: Fixed in 2e78e1d1d069b5316c1253fa37af09a04d801b3b.

index.html
gkellogg: In this case, we're going back to the initial state, where _base IRI_ is either explicitly set, or comes from the document location, which is in _base URL_. It's true that this could get it wrong, as if you're deep into remote contexts, _base URL_ will be that of the containing context, rather than the original document.

Perhaps, we should have something accessible from the API which would record the `{{RemoteDocument/documentUrl}}` which could be used instead of _base URL_. Or, perhaps, update the Context Processing algorithm to remember the original _base URL_ passed in that doesn't get changed with each new remote context, just for this purpose.

This is an example of the "ripple effects" that making this change does, which I'm not sure actually clarifies the algorithm.

gkellogg: I solved this in dcb7a57c by adding an _original base URL_ component to an active context.

davidlehn:
Commit message should be "Update JCS", not "Update jsonld.js":
https://github.com/w3c/json-ld-api/pull/356/commits/2614999c9823ddc6195c8ad9a0aaf120e32f8486

gkellogg:
Actually, the file updated is coming/jsonld.js. But it is to update JCS.

davidlehn:
Ah, I totally misread what was going on there.

PR #360, opened 2020-01-28
Be more explicit about map values in IRI Compaction.

### LINKED ISSUE #357: Missing type checking of values in IRI Compaction?
In several places, the [IRI Compaction algorithm](https://w3c.github.io/json-ld-api/#algorithm-7) seems to assume that the `value` variable is a map. For example, in step 2.11:

> If `processing mode` is not `json-ld-1.0` and `value` does not contain an `@index` entry, append `@index` and `@index@set` to `containers`.

Step 2.5 seems to have what might have been an attempt at verifying that `value` is a map, but with missing language or markup:

> If `@index` is an entry in `value` is a map, and `value` is not a graph object then append the values `@index` and `@index@set` to `containers`.

However, IRI Compaction can be called with non-map values of `value`, as in [Compaction](https://w3c.github.io/json-ld-api/#algorithm-6) steps 12.7 and 12.7.1:

> 12.7) If `expanded value` is an empty array:
> 12.7.1) Initialize `item active property` by IRI compacting `expanded property` using `expanded value` for value and `inside reverse` for reverse.

Should IRI Compaction have better type checking? Is there a mistake in the argument value in12.7.1?

gkellogg:
Fixes #357.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/360.html" title="Last updated on Jan 28, 2020, 10:02 PM UTC (52b08a6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/360/2f57a54...52b08a6.html" title="Last updated on Jan 28, 2020, 10:02 PM UTC (52b08a6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1080,16 +1080,17 @@
     <p>Each <a>term definition</a> consists of:</p>
     <ul>
       <li>an <dfn>IRI mapping</dfn> (<a>IRI</a>),</li>
-      <li>a flag <dfn>reverse property</dfn> (<a>boolean</a>),</li>
-       <li>an optional <dfn>type mapping</dfn> (<a>IRI</a>),</li>
-       <li>an optional <dfn>language mapping</dfn> (<a>string</a>),</li>
-      <li class="changed">an optional <dfn>direction mapping</dfn> (`"ltr"` or `"rtl"`),</li>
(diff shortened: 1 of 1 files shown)


gkellogg:
Closing, because this seems to be infected with changes from another PR.

PR #362, opened 2020-01-28
Clarify that API options use default values from JsonLdOptions

### LINKED ISSUE #358: Confusing/contradictory text regarding options passed to Compaction
The definition of `compact()` in [The JsonLdProcessor Interface](https://w3c.github.io/json-ld-api/#the-jsonldprocessor-interface) section calls the Compaction algorithm:

> using `active context`, `inverse context`, `null` for `active property`, `expanded input` as `element`, and if passed, the `compactArrays` and `ordered` flags in `options`.

I find the "if passed" text confusing here. `options` is defined as:

> A set of options to configure the algorithms. This allows, e.g., to set the input document's base IRI.

It wasn't immediately clear to me that `options` is full [JsonLdOptions Type](https://w3c.github.io/json-ld-api/#the-jsonldoptions-type), with all the default values as specified in section 9.3. For that reason, I missed the fact that `compact()` needs to pass the default value of `compactArrays` as `true` to the Compaction algorithm. An explicit link between `compact()` and the `JsonLdOptions Type` section would help to clarify this. 

Further confusing the issue, the [Compaction Algorithm](https://w3c.github.io/json-ld-api/#algorithm-6) says:

> The optional inputs are the `compactArrays` flag and the `ordered` flag, used to order map entry keys lexicographically, where noted. If not passed, the both flags are set to `false`.

It's not clear to me if this `false` default value can ever be used since calls to the algorithm seem to always need a full options structure, in which `compactArrays` must have a value.

gkellogg:
and remove most conditional use or initialization of those options.

Fixes #358.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/362.html" title="Last updated on Jan 29, 2020, 11:41 PM UTC (a9c7c7b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/362/c25e18c...a9c7c7b.html" title="Last updated on Jan 29, 2020, 11:41 PM UTC (a9c7c7b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2378,7 +2378,8 @@
           the {{JsonLdOptions/ordered}} flag, used to order
           <a>map entry</a> keys lexicographically, where noted,
           and the <var>from map</var> flag, used to control reverting
-          previous <a>term definitions</a> in the <a>active context</a> associated with non-propagated <a>contexts</a>.</p>
+          previous <a>term definitions</a> in the <a>active context</a> associated with non-propagated <a>contexts</a>.
+          If not passed, the optional flags are set to <code>false</code></span>.</p>
 
       <p class="changed">The algorithm also performs processing steps specific to expanding
         a <a>JSON-LD Frame</a>. For a <a>frame</a>, the <code>@id</code> and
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2380,8 +2380,7 @@ <h3>Algorithm</h3>
           the {{JsonLdOptions/ordered}} flag, used to order
           <a>map entry</a> keys lexicographically, where noted,
           and the <var>from map</var> flag, used to control reverting
-          previous <a>term definitions</a> in the <a>active context</a> associated with non-propagated <a>contexts</a>.
-          If not passed, the both flags are set to <code>false</code></span>.</p>
+          previous <a>term definitions</a> in the <a>active context</a> associated with non-propagated <a>contexts</a>.</p>
```
davidlehn: Lost the "changed" span closing tag.

davidlehn: Same with two changes below.

gkellogg: Fixed by restoring text.

kasei:
@gkellogg yes, #358 improves the relevant language. Looks good, thanks.

Issue #370, opened 2020-02-14
Unclear how to set original base URL in Context Processing


kasei:
(Related to #265, #356)

[Context Processing Overview](https://w3c.github.io/json-ld-api/#overview) 4.1.1 says:

> When an  active context  is initialized, the value of the  original base URL  is initialized from the original  `documentUrl`  of the document containing the initial  context . This is necessary when resetting the  active context  by setting it to  null  to retain the original default  base  IRI .

What about when there is no `documentUrl`? Is it `null`, or is it the current base IRI? The `documentUrl` field is only relevant when the LoadDocumentCallback has been used.

This also impacts `compact()` step 3, and `expand()` step 4.

gkellogg:
This is not dissimilar from Turtle serialization from a form, rather than a remote document. Also, the JSON-LD Playground. Implementations provide a reasonable default to use as a document base.

Perhaps if step 7 of the `expand()` API definition provided for the case where there is no remote document having a documentUrl. But, really, the use of a relative URL to locate a context from a form is non-sensical. Maybe step 5.2.1 should generate an error if _context_ does not result in a valid IRI after resolution. Perhaps `invalid IRI mapping` could be used for this case too.

gkellogg:
Actually, the LoadDocumentCallback will generate a "loading document failed" error if the URL can't be resolved, so no new error detection is required.

I did add changes to the `expand()` API definition to fallback to `JsonLdOptions/base` if there is no `documentUrl`, which is how it's typically handled in form-based parsers. Implementations may provide a default for this, but we don't need to describe that non-normative behavior.

gkellogg:
@kasei see pending change in #377.

kasei:
@gkellogg I think the changes to `expand()` in #377 are good, but the description in 4.1.1 is still misleading as in some cases there will simply not be a relevant `documntUrl`.

gkellogg:
Updated in 50a50ca0, and to give `base` a default of `null`.

gkellogg:
Updated in #377.

PR #377, opened 2020-02-14
More context processing

### LINKED ISSUE #369: Base URL is not always absolute in Context Processing
(Related to #265, #356)

[Context Processing](https://w3c.github.io/json-ld-api/#algorithm) step 5.2.1:

> Initialize context to the result of resolving context against base URL.

This `base URL` comes directly as the argument to this algorithm. However, in at least one case, the value of `base URL` is not an absolute IRI.

The `expand()` defintion in JsonLdProcessor Interface step 5 says:

> If the expandContext option in options is set, update the active context using the Context Processing algorithm, passing the expandContext as local context and expandContext as base URL.

In test [t0077](https://w3c.github.io/json-ld-api/tests/expand-manifest.html#t0077), the value of "expandContext" is simply `"0077-context.jsonld"`, which will cause Context Processing to attempt to resolve a relative IRI against another relative IRI. I don't think that's what is intended here.

### LINKED ISSUE #372: Use of possibly undefined variables in Context Processing
(Related to #265, #356)

[Context Processing](https://w3c.github.io/json-ld-api/#algorithm) step 5.2.5 uses `loaded context` which is only defined in 5.2.4.3 (which is not necessarily evaluated prior to 5.2.5)

[Context Processing](https://w3c.github.io/json-ld-api/#algorithm) step 5.2.5 uses `context document` which is only defined in 5.2.4 (which is not necessarily evaluated prior to 5.2.5)

### LINKED ISSUE #373: Possible use of non-URL values as the base URL in expand()
(Related to #265, #356)

[`expand()`](https://w3c.github.io/json-ld-api/#dom-jsonldprocessor-expand) step 5:

> If an expandContext option has been passed, update the active context using the Context Processing algorithm , passing the expandContext as local context and  expandContext  as base URL  . If expandContext is a map having an `@context` entry , pass that entry's value instead.

What if `expandContext` is not a URL and not a map having a `@context` entry? This will result in a non-URL being passed as the base URL.

### LINKED ISSUE #374: Missing use of required Expansion input variable in expand()
(Related to #265, #356)

[`expand()`](https://w3c.github.io/json-ld-api/#dom-jsonldprocessor-expand) step 7:

> Set expanded output to the result of using the Expansion algorithm, passing the active context, document from remote document or input if there is no remote document as element, documentUrl as base URL, and frameExpansion and and ordered flags from options.

Expansion 5.1.2 says `active property` is a required input variable, but it is not supplied at this callsite.

### LINKED ISSUE #375: No default value defined for base URL in Create Term Definition
(Related to #265, #356)

[Create Term Definition](https://w3c.github.io/json-ld-api/#algorithm-0) says that `base URL` is an optional input variable, but does not define a default value. It is then used unconditionally in step 21.3 and 21.4.

gkellogg:
* Make _expandContext_ `@type: @id` and set option values to be relative to the manifest location. Fixes #369.
* Set a fallback in `expand()` processing if there is no `documentUrl`. This uses `JsonLdOptions/base`, which could be null, or something else appropriate for when the `input` or `expandContext` is not a URL. For #370.
* Describe that if `documentUrl` is not available, values come from `JsonLdOptions/base`, which now defaults to `null`. Fixes #375.
* Reorder clauses in 5.1.2 to make it more clear that setting _base IRI_ and _original base URL_ of the new _active context_ is unconditional, unlinke _previous_context_.
* Create values for _context document_ and _loaded context_ in the case that the previous document was cached in 5.2.3 of the Create Context Algorithm. Fixes #372.
* Describe how to handle a remote context that is not a valid IRI (not absolute) when it can't be resolved against _base URL_, because it is also not a valid IRI. For #373.
* Update `expand()` to pass the _original base URL_ from the current _active context_ when creating a new _active context_ using _expandContext_. Fixes #373.
* Pass `null` for _active property_ on initial call to the Expansion algorithm from `expand()`. Fixes #374.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/377.html" title="Last updated on Feb 18, 2020, 7:38 PM UTC (45f077a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/377/dcd4869...45f077a.html" title="Last updated on Feb 18, 2020, 7:38 PM UTC (45f077a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1069,7 +1069,7 @@
     <ul>
       <li>the active <a>term definitions</a> which specify how
         keys and values have to be interpreted (<a>array</a> of <a>term definitions</a>),</li>
-      <li>the current <a>base IRI</a> (<a>IRI</a>),</li>
+      <li>the current <dfn data-lt="context-base-iri" data-lt-noDefault>base IRI</dfn> (<a>IRI</a>),</li>
       <li class="changed">the <dfn>original base URL</dfn> (<a>IRI</a>),</li>
       <li>an optional <a>vocabulary mapping</a> (<a>IRI</a>),</li>
       <li>an optional <a>default language</a> (<a>string</a>),</li>
@@ -1131,7 +1131,7 @@
       <p>If <a>context</a> is a <a class="changed">map</a>,
         it is a <a data-cite="JSON-LD11#dfn-context-definition">context definition</a>.
         We first update
-        the <a>base IRI</a>,
+        the <a class="changed" data-lt="context-base-iri">base IRI</a>,
         <span class="changed">the <a>default base direction</a></span>,
         the <a>default language</a>,
         <span class="changed">context propagation</span>,
@@ -1163,7 +1163,7 @@
         otherwise from the {{JsonLdOptions/base}} API option.
         This is necessary when resetting the <a>active context</a>
         by setting it to `null`
-        to retain the original default <a>base IRI</a>.</p>
+        to retain the original default <a class="changed" data-lt="context-base-iri">base IRI</a>.</p>
 
       <p>Then, for every other <a>entry</a> in <a>local context</a>, we update
         the <a>term definition</a> in <var>result</var>. Since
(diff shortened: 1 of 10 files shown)


index.html
```diff
@@ -1247,7 +1253,10 @@ <h3>Algorithm</h3>
                 <li class="changed">If <var>context</var> was previously dereferenced,
                   then the processor MUST NOT do a further dereference, and
                   <a>context</a> is set to the
-                  previously established <a>internal representation</a>.
+                  previously established <a>internal representation</a>,
```
```suggestion
                  previously established <a>internal representation</a>:
```
With a comma, it looks like the rest of the sentence must be executed *in addition* to what precedes.
As I understand, the rest of the sentence is rather explaining *how* to do what precedes. Hence the proposal to use a colon instead.
A colon makes it more explicit than a comma, which may induce that the rest of the sentence

Issue #380, opened 2020-02-19
Expansion does not take property-scoped contexts for nested properties into account


niklasl:
There is currently a limitation in the JSON-LD 1.1 expansion algorithm which ignores property-scoped contexts for nested properties.

My specific case is:

```json
{
  "@context": {
    "@version": 1.1,
    "@vocab": "http://purl.org/dc/terms/",
    "bibo": "http://purl.org/ontology/bibo/",
    "Print": "bibo:Book",
    "name": "http://www.w3.org/2000/01/rdf-schema#label",
    "instanceOf": "@nest",
    "contributionByRole": {
      "@id": "@nest",
      "@context": {
        "agent": "@nest",
        "aut": "creator"
      }
    },
    "provisionActivityByType": {
      "@id": "@nest",
      "@context": {
        "Publication": {
          "@id": "@nest",
          "@context": {"date": "published", "agent": "publisher"}
        }
      }
    },
    "identifiedByType": {
      "@id": "@nest",
      "@context": {
        "Isbn": {"@id": "@nest"},
        "value": "bibo:isbn"
      }
    }
  },
  "@id": "book/one",
  "@type": "Print",
  "instanceOf": {
    "contributionByRole": {
      "aut": {
        "agent": {"name": "Some Body"}
      }
    }
  },
  "identifiedByType": {
    "Isbn": {
      "value": "1234567890"
    }
  },
  "provisionActivityByType": {
    "Publication": {
      "date": "1999",
      "agent": {"name": "PubCorp"}
    }
  }
}
```

From which I expected to get (recompacted for brevity):

```json
{
  "@context": {
    "@vocab": "http://purl.org/dc/terms/",
    "bibo": "http://purl.org/ontology/bibo/",
    "name": "http://www.w3.org/2000/01/rdf-schema#label"
  },
  "@id": "book/one",
  "@type": "bibo:Book",
  "bibo:isbn": "1234567890",
  "published": "1999",
  "publisher": {"name": "PubCorp"},
  "creator": {"name": "Some Body"}
}
```
[...]

PR #388, opened 2020-02-23
Add tests and spec text for scoped-contexts on `@nest` alias.

gkellogg:
For #380

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/388.html" title="Last updated on Nov 14, 2020, 7:29 PM UTC (e7b759e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/388/18bc029...e7b759e.html" title="Last updated on Nov 14, 2020, 7:29 PM UTC (e7b759e)">Diff</a>


index.html
```diff
@@ -3013,9 +3013,20 @@ <h3>Algorithm</h3>
...
-                  using <var>nested value</var> for <var>element</var>.
+                  using <var>nesting-key</var> for <var>active property</var>, and
+                  <var>nested value</var> for <var>element</var>.
+                  <div class="note">Steps <a href="#alg-expand-property-scoped-context">3</a>
+                    and <a href="#alg-expand-property-scoped-context2">8</a>
+                    may update the <var>active context</var> based on a
+                    property-scoped context associated with <var>nesting-key</var>.
+                    Updates to <var>active context</var> are restricted to the
+                    recursive operation, and do not propogate to subsequent iterations
+                    on <var>nested value</var> and <var>nesting-key</var>.</div>
```
```suggestion
                    on <var>nested values</var> and <var>nesting-key</var>.</div>
```

index.html
```diff
@@ -3013,9 +3013,20 @@ <h3>Algorithm</h3>
                   <var>nested value</var> expands to <code>@value</code>, an
                   <a data-link-for="JsonLdErrorCode">invalid @nest value</a> error
                   has been detected and processing is aborted.</li>
-                <li>Recursively repeat steps <a href="#alg-expand-each-key-value">13</a>
+                <li>Recursively repeat steps
+                  <a href="#alg-expand-property-scoped-context">3</a>,
+                  <a href="#alg-expand-property-scoped-context2">8</a>,
```
pchampin: why not also include step 9 (processing local context)?
I know that @niklasl didn't explicitly required it in #380,
but since we are there, not supporting local contexts seems as arbitrary as not supporting scoped contexts...

index.html
gkellogg: Either would be correct, but this is fine.

index.html
gkellogg: The Syntax document is explicit where `@context` can be found. For instance, not in a value object or list object. It's really only in node objects and graph objects, so adding this would require a change to that, which is outside our scope.

pchampin: Got it :)

pchampin (approved):
Good for me.

PR #390, opened 2020-02-23
Update language when `@type` is used as a term

### LINKED ISSUE #387: keyword redefinition and protection
When creating a term definition for `@type`, the document says at step 4 "value MUST be a map with only the entry `@container` and value `@set` and optional entry `@protected`", which I interpret as "it must contain an `@container` whose value is `@set` and may also have `@protected` but nothing else". However, looking at test [pr32](https://w3c.github.io/json-ld-api/tests/expand/pr32-in.jsonld), it seems that the following is not a keyword redefinition:

```json
{
  "@type": {"@protected": true}
}
```

After changing my implementation to allow the value to not contain any `@container`, the test passes (I get a protected term redefinition as expected, from step 27), without any new failure.

gkellogg:
so that entries can be either or both `@container` or `@protected`.

Fixes #387.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/390.html" title="Last updated on Feb 25, 2020, 12:47 AM UTC (c872e08)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/390/ad97b98...c872e08.html" title="Last updated on Feb 25, 2020, 12:47 AM UTC (c872e08)">Diff</a>


index.html
```diff
@@ -1538,10 +1538,12 @@ <h3>Algorithm</h3>
           and <a>processing mode</a> is `json-ld-1.0`,
           a <a data-link-for="JsonLdErrorCode">keyword redefinition</a> error has
           been detected and processing is aborted.
-          At this point, <var>value</var>
-          MUST be a <a>map</a> with only the entry <code>@container</code>
-          and value <code>@set</code>
-          and optional entry `@protected`.
+          At this point,
+          <var>value</var> MUST be a <a>map</a> with either or both of the following <a>entries</a>:
```
```suggestion
          <var>value</var> MUST be a <a>map</a> with only either or both of the following <a>entries</a>:
```
I wouldn't delete "only", because without it, it seems that other keys would be ok (as long as at least one of those is present). Granted, the sentence afterwards exclude any other key, but I find it clearer that way.

PR #394, opened 2020-02-25
add test where `@nest` alias uses `"@id": "@nest"`

pchampin:
catches an error that went unnoticed in some implementations


tests/compact-manifest.jsonld
```diff
@@ -1786,6 +1786,15 @@
       "context": "compact/n003-context.jsonld",
       "expect": "compact/n003-out.jsonld",
       "option": {"specVersion": "json-ld-1.1"}
+    }, {
+      "@id": "#tn003a",
+      "@type": ["jld:PositiveEvaluationTest", "jld:CompactTest"],
+      "name": "Nests using alias of @nest (defined with @id)",
+      "purpose": "Compaction using @nest",
+      "input": "compact/n003-in.jsonld",
```
davidlehn: In general (or always?) we haven't been reusing test files this way.  We usually just copy things over to a tnNNN+1 rather than a tn003 and tn003a.  That does get annoying when you don't have related tests next to each other... like here it would be tn011.  Not sure if it matters.

PR #395, opened 2020-02-29
Treat _inverse context_ as a field in the _active context_

gkellogg:
which is initialized on demand. This simplifies algorithm calling sequences and better captures implementation experience while not making any behavioral change.

cc/ @kasei


gkellogg:
> The only risk I see in this change is the semi-implicit nullification of the `inverse context` field every time another field is updated. But apart from that, this does indeed simplify the algorithms. Good move.

In practice, this never happens. Only when a new context is introduced is is “nulled out” and, previously, it would have had to be recreated anyway. Memoization is the real key, and cached values would need to be invalidated if there were an update. But we don’t really update, except as part of creation.

PR #409, opened 2020-03-07
Order JsonLdErrorCode descriptions.

davidlehn:
Sort the descriptions to match the ordered enum list above.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 15, 2020, 7:58 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fjson-ld-api%2F811419fba0804f24d66d9b4f59dd3bc6c0f1e065%2Findex.html%3FisPreview%3Dtrue)

```
Navigation timeout of 30000 ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/json-ld-api%23409.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6415,6 +6415,9 @@
       </pre>
 
       <dl>
+        <dt class="changed"><dfn data-dfn-for="RemoteDocument">contentType</dfn></dt>
+        <dd class="changed">The <dfn data-cite="RFC2045#section-5">Content-Type</dfn>
+          of the loaded document, exclusive of any optional parameters.</dd>
         <dt><dfn data-dfn-for="RemoteDocument">contextUrl</dfn></dt>
         <dd>If available, the value of the HTTP Link Header [[RFC8288]]
           using the <code>http://www.w3.org/ns/json-ld#context</code> link relation
@@ -6430,9 +6433,6 @@
         <dt><dfn data-dfn-for="RemoteDocument">document</dfn></dt>
         <dd>The retrieved document.
           This can either be the raw payload or the already parsed document.</dd>
-        <dt class="changed"><dfn data-dfn-for="RemoteDocument">contentType</dfn></dt>
-        <dd class="changed">The <dfn data-cite="RFC2045#section-5">Content-Type</dfn>
-          of the loaded document, exclusive of any optional parameters.</dd>
         <dt class="changed"><dfn data-dfn-for="RemoteDocument">profile</dfn></dt>
         <dd class="changed">The value of any <code>profile</code> parameter
           retrieved as part of the original {{RemoteDocument/contentType}}.</dd>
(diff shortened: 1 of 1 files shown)


davidlehn:
Also sorted RemoteDocument descriptions to match IDL above.

davidlehn:
Hmm.  Maybe this isn't needed and the RemoteDocument dl is just missing data-sort attribute?  If that's the case, the proper patch is probably to add the data-sort attribute and manually resort the JsonLdErrorCode list to match the sorting data-sort does.  (just piping through "sort" puts the IRI thing at top and orders some things slightly different)

davidlehn:
For IDL sections, I assume it's desired to have the IDL item list match the description list.  If descriptions are using `data-sort`, then should the IDL item order match that?  And if the IDL items are ordered in `data-sort` order, seems the markup descriptions should use that order too.  So then everything ends up manually sorted.  Is that the idea?  `data-sort` seems a nice indicator, but I guess it's redundant.

gkellogg:
Method parameters should stay in order, and of course, algorithm steps. Otherwise, I think we can use `data-sort`.

It's not too important that the HTML source be ordered if we're using `data-sort`, and sometimes it made sense to keep things out of order, but mostly to catch issues where things were't properly ordered in the source. It's fine to re-order the source to match what is ordered with `data-sort`.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2020/2020-03-13-json-ld#section2-2).

- No actions or resolutions
[api] Order JsonLdErrorCode descriptions<br/>
Benjamin Young: <a href="https://github.com/w3c/json-ld-api/pull/409">https://github.com/w3c/json-ld-api/pull/409</a><br/>
Gregg Kellogg: other is reordering error code. Normally we use &lt;dl> with data-sort attribute that orders them, but this maybe doesn’t use data-sort<br/>
… so need to get the text in the HTML to align with the sort order<br/>
David I. Lehn: ordering thing is not very exciting :)<br/>
… if we use data-sort it will put it in a particular order. And the IDL should be the same order as the -sort<br/>
… everything should be ordered the same. Wasn’t sure what to do, and ran out of time on it<br/>
Gregg Kellogg: up to the editors for what to do. Doesn’t affect the rendered document at all. Do things exist in the source where you expect them … so a good idea to have in the same order. data-sort corrects for past errors<br/>
… if someone wants to do it, that’s great<br/>
David I. Lehn: I sorted the values by piping through sort locally<br/>
… which put IRI at the top<br/>
Gregg Kellogg: Good to just do what the algorithm does<br/>
David I. Lehn: Will just make it look right :)<br/>
Gregg Kellogg: don’t sort method param lists<br/>
… as they’re in the order they’re used in the method definition<br/>
… but error codes, enum values don’t order by default<br/>
… and need to be done manually, but keeping the enum in the order below makes the most sense<br/>
… IRI comes first due to caps<br/>
… whereas in data-sort it’s after invalid<br/>
David I. Lehn: Which things to not sort?<br/>
Gregg Kellogg: in jsonld processor interface, the definitions of the methods, there’s definitions of the three params … those should not be ordered alphabeitcally but in the order they appear in the method definition<br/>
David I. Lehn: Right, okay. I’ll work on an update<br/>
Gregg Kellogg: Great, need to sign off<br/>
… will push syntax today<br/>
Benjamin Young: Great, thanks, bye!<br/>
… revisit 415?<br/>
Rob Sanderson: [crickets]<br/>
</details>

PR #417, opened 2020-03-15
Update implementation report on streaming parser

rubensworks:
jsonld-streaming-parser.js now passes all toRdf tests.

@gkellogg I see that the updated HTML only reports 367 toRdf test, while my test suite runner reports 377 tests (7 skipped). Is it possible that something else needs to be updated?


gkellogg:
There is a cache of the merged manifests in triples form that needs to be updated to report on all the tests. I’ll take care of it shortly.

gkellogg:
@rubensworks Try rebasing and running again.

Note that your skipped reports include those for specVersion json-ld-1.0.

We'll need to do this again after PR #416 is merged.

rubensworks:
Done.

> We'll need to do this again after PR #416 is merged.

Yes. I plan to add some more tests in the next couple of days as well.

Issue #408, opened 2020-03-07
Vocab scope in toRdf-tc010


rubensworks:
Test [toRdf-tc010](https://w3c.github.io/json-ld-api/tests/toRdf-manifest#tc010) defines the following input:
```json
{
  "@context": {
    "@vocab": "http://example/",
    "B": {"@context": {"c": "http://example.org/c"}}
  },
  "a": {
    "@context": {"@vocab": "http://example.com/"},
    "@type": "B",
    "a": "A in example.com",
    "c": "C in example.org"
  },
  "c": "C in example"
}
```

And it expects the following output:
```
_:b0 <http://example/a> _:b1 .
_:b0 <http://example/c> "C in example" .
_:b1 <http://example.org/c> "C in example.org" .
_:b1 <http://example.com/a> "A in example.com" .
_:b1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://example/B> .
```

It is however unclear to me why this last triple has `http://example/B` as object, instead of `http://example.com/B`.

In the [syntax document](https://w3c.github.io/json-ld-syntax/#example-47-expansion-using-embedded-and-scoped-contexts), it is mentioned that _"A property-scoped context is processed first, followed by any embedded context, followed lastly by the type-scoped contexts, in the appropriate order."_

This would be my understanding of this:
When handling the node value of `"a"`, this would mean that the `@context` value would be handled first, and set `@vocab` of this node to `http://example.com/`. Next, the type-scoped context would be processed, inheriting this vocab (?). So `B` would be expanded based on vocab set to`http://example.com/`.

-----

When removing the type-scoped context definition I do get the output I would expect (with `http://example.com/B`) with the distiller and the JSON-LD playground:
```json
{
  "@context": {
    "@vocab": "http://example/"
  },
  "a": {
    "@context": {"@vocab": "http://example.com/"},
    "@type": "B",
    "a": "A in example.com",
    "c": "C in example.org"
  },
  "c": "C in example"
}
``` 
Output:
```
_:b0 <http://example/a> _:b1 .
_:b0 <http://example/c> "C in example" .
_:b1 <http://example.com/a> "A in example.com" .
_:b1 <http://example.com/c> "C in example.org" .
_:b1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://example.com/B> .
```

Is this inconsistency intentional?

gkellogg:
[...]

PR #418, opened 2020-03-16
Add term scoping test for toRdf

rubensworks:
As discussed in #408, this adds a dedicated toRdf test for checking that terms should make use of @vocab relative to the scope in which the term was defined.

### DIFF
--- a/tests/expand-manifest.jsonld
+++ b/tests/expand-manifest.jsonld
@@ -1237,6 +1237,14 @@
       "expect": "expand/c034-out.jsonld",
       "option": {"specVersion": "json-ld-1.1"}
     }, {
+      "@id": "#tc035",
+      "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
+      "name": "Term scoping with embedded contexts.",
+      "purpose": "Terms should make use of @vocab relative to the scope in which the term was defined.",
+      "input": "expand/c035-in.jsonld",
+      "expect": "expand/c035-out.nq",
+      "option": {"specVersion": "json-ld-1.1"}
+    }, {
       "@id": "#tdi01",
       "@type": [ "jld:PositiveEvaluationTest", "jld:ExpandTest" ],
       "name": "Expand string using default and term directions",
--- a/tests/expand-manifest.jsonld
+++ b/tests/expand-manifest.jsonld
@@ -1242,7 +1242,7 @@
       "name": "Term scoping with embedded contexts.",
       "purpose": "Terms should make use of @vocab relative to the scope in which the term was defined.",
       "input": "expand/c035-in.jsonld",
-      "expect": "expand/c035-out.nq",
+      "expect": "expand/c035-out.jsonld",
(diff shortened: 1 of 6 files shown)


gkellogg:

rubensworks:

gkellogg:

gkellogg:
Does [tc005](https://w3c.github.io/json-ld-api/tests/toRdf-manifest.html#tc005) cover this use case? It has a duplicate in expansion.

rubensworks:
Yes, tc005 does indeed test the same thing. But the main goal of that test seems to be regarding property-scoped contexts, while this PR focuses on the term-scoping part.

I can't seem to find an equivalent test in #422, but maybe I'm reading over it?

gkellogg:
There are several tests that have embedded contexts which affect `@vocab` scoping, such as [toRdf-manifest#t0059](https://w3c.github.io/json-ld-api/tests/toRdf/e059-in.jsonld), if by "term scoping" you mean that properties used within a node object having its own context are evaluated relative to that embedded context.

rubensworks:
I would say t0059 is slightly different still.
Perhaps it follows the same logic in the processing model from the API document, but in alternative processing models (such as the one in my implementation), this touches upon a different part of the logic.

Concretely, I consider the following parts important:
```
{
  "@context": {
    "@vocab": "http://vocab.org/",
    "prop1": {} <---- This "binds" this property to the @vocab in this context.
  },
  "@id": "ex:outer",
  "foo": {
    "@context": {
      "@vocab": "http://vocab.override.org/" <---- This should not override the vocab for prop 1, only prop2 that has not been "bound" yet.
    },
    "@id": "ex:inner",
    "prop1": "baz1",
    "prop2": "baz2"
  }
}
```

Does this make my intention of the test clearer?

gkellogg:
It seemed to me that the other test I added did that, but if you’d like to update it, go ahead. Just, please also update the corresponding expand test. You should be able,to use the distiller or playground to get the expanded results, if necessary. Or, just push it to a branch on this repo so I can supplement with the appropriate content.

rubensworks:
@gkellogg I've added a corresponding expansion test.

I was not able to push my branch to this repo directly, I don't seem to have permissions for that. @iherman would it be possible to add me?

tests/expand-manifest.jsonld
```diff
@@ -1236,6 +1236,14 @@
       "input": "expand/c034-in.jsonld",
       "expect": "expand/c034-out.jsonld",
       "option": {"specVersion": "json-ld-1.1"}
+    }, {
+      "@id": "#tc035",
+      "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
+      "name": "Term scoping with embedded contexts.",
+      "purpose": "Terms should make use of @vocab relative to the scope in which the term was defined.",
+      "input": "expand/c035-in.jsonld",
+      "expect": "expand/c035-out.nq",
```
gkellogg:
```suggestion
      "expect": "expand/c035-out.jsonld",
```

PR #419, opened 2020-03-16
Add toRdf test for protected term override in prop-scoped context

rubensworks:
It appeared toRdf was missing a test for checking that protected terms can be overridden from property-scoped contexts.

### DIFF
--- a/tests/toRdf-manifest.jsonld
+++ b/tests/toRdf-manifest.jsonld
@@ -2715,6 +2715,14 @@
       "input": "toRdf/pr39-in.jsonld",
       "expect": "toRdf/pr39-out.nq"
     }, {
+      "@id": "#tpr40",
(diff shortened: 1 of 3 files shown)


gkellogg:

rubensworks:

gkellogg:

gkellogg:
It seems that [pr01](https://w3c.github.io/json-ld-api/tests/toRdf-manifest.html#tpr01) does essentially the same test, which comes from expansion.

rubensworks:
pr01 seems to be expecting an error, while the new test from this PR does not.

Is there a corresponding test for this in #422?

gkellogg:
Indeed, I can't find one that does quite this, although I thought we have it covered.

I'll add it to #422 with a duplicate for expansion.

gkellogg:
Closed via #422.

PR #420, opened 2020-03-16
Add toRdf tests on invalid keyword aliases

rubensworks:
Tests for checking the `invalid keyword aliases` error code seemed to be missing.

### DIFF
--- a/tests/toRdf-manifest.jsonld
+++ b/tests/toRdf-manifest.jsonld
@@ -1138,6 +1138,22 @@
       "expectErrorCode": "keyword redefinition",
       "option": {"specVersion": "json-ld-1.1"}
     }, {
+      "@id": "#tec03",
+      "@type": [ "jld:NegativeEvaluationTest", "jld:ToRDFTest" ],
+      "name": "Term definition with alias to @context",
+      "purpose": "Verifies that an exception is raised if a term definition attempts to alias @context",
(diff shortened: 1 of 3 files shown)


gkellogg:

rubensworks:

> I don't see what steps from the algorithm would result in an error from ec04:

This was actually based on the error message of jsonld.js:

> jsonld.SyntaxError: Invalid JSON-LD syntax; @context and @preserve cannot be aliased.

I assumed that this was according to the spec, but I didn't check to be honest, so I may be wrong there. Will remove that one test.

gkellogg:
Perhaps @davidlehn knows about the origin of that error in jsonld.js. For the time being, I'd just remove it from this PR.

gkellogg:

gkellogg:
tec03 seems to be the same as expansion [te019](https://w3c.github.io/json-ld-api/tests/expand-manifest.html#te019). Best thing is to duplicate att the ecxx tests cases from expansion to toRdf.

I'm going to close this PR and create a new PR that does that.

gkellogg:
It's a bit worse than that, the numbering has diverged, so it will take a bit more coordination. The "exxx" use cases should probably be pretty much duplicated between the two suites.

davidlehn:
I'm not sure how to track down that error at the moment.  If there are bad errors, would be nice to have positive tests to expose those.

Issue #379, opened 2020-02-19
Inconsistent management of empty terms


pchampin:
According to [Section 9. JSON-LD Grammar](https://w3c.github.io/json-ld-syntax/#json-ld-grammar) of the syntax document:

> the term MUST NOT be an empty string ("") as not all programming languages are able to handle empty JSON keys

However:
* there is no step in the [Context Processing algorithms](https://w3c.github.io/json-ld-api/#context-processing-algorithms) for detecting empty term definitions,
* there is no test for it in the test-suite,
* implementations are handling this very heterogeneously (see below).

Consider the following example:
```json
{
"@context": {
  "": "http://example.org/"
},
  "": "foo",
  ":bar": "baz"
}
```

* pyld and jsonld.js reject it with an error
* rdflib-jsonld accepts it and hallily expands both properties
* the Ruby distiller accepts it, expands "", but ignores ":bar".

For the record, this "blind spot" of the test-suite and algotithms dates back to JSON-LD 1.0.

What made me realize that (and makes me think this is a problem we should probably address) is that two contexts in [prefixcommons' registry](https://github.com/prefixcommons/biocontext/tree/master/registry) (mentionned by @hsolbrig in w3c/json-ld-syntax#329) have a term definition for the empty string, which makes them unusable with pyld or jsonld.js. Apparently, this went unnoticed because they are mostly using rdflib-jsonld...

I know it is late (too late?) to fix it, but I think the spec should be more explicit about how to handle a term definition when the term is the empty string (`""`). Rejecting it seems consistent with the grammar. And a test should therefore be added.

hsolbrig:
> the prefixcommons' registry ...have a term definition for the empty string, which makes them unusable with pyld or jsonld.js. ...  Apparently, this went unnoticed because they are mostly using rdflib-jsonld...

Interestingly, I think that it went unnoticed because they were parsing the document as plain old JSON -- something that we may want to keep in mind when adding features that require the expanded form.

iherman:
> Interestingly, I think that it went unnoticed because they were parsing the document as plain old JSON -- something that we may want to keep in mind when adding features that require the expanded form.

Wow, of course! There are a number of cases when JSON-LD files are processed directly using JSON parsers (I do it with some other projects I am involved with, too). This may be a warning to be put somewhere (Best practices, maybe?).

gkellogg:
[...]

PR #421, opened 2020-03-16
Reject empty term

pchampin:
***

### :boom: Error: write EPROTO 140222018242432:error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version:../deps/openssl/openssl/ssl/s23_clnt.c:772:
 :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 16, 2020, 11:38 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [HTML Diff Service](http://services.w3.org/htmldiff) - The HTML Diff Service is used to create HTML diffs of the spec changes suggested in a pull request.

:link: [Related URL](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fw3c%2Fjson-ld-api%2Fpull%2F421%2F257c09e.html&doc2=https%3A%2F%2Fpr-preview.s3.amazonaws.com%2Fw3c%2Fjson-ld-api%2Fpull%2F421.html)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/json-ld-api%23421.)._
</details>


pchampin:
This PR addresses #379

davidlehn:
Need to add the same toRdf test.  Also "e052" will bump into the recursion test PR naming.  May be easier to handle that here with less tests to rename?

Sad to see such a nice obfuscated json-ld feature get testsuited away.

gkellogg:
> Need to add the same toRdf test. Also "e052" will bump into the recursion test PR naming. May be easier to handle that here with less tests to rename?

Yes, we should compare with other PRs and rename as appropriate, and add a toRdf version. (It would be nice to have a way to not have to do this manually!)

> Sad to see such a nice obfuscated json-ld feature get testsuited away.

👍 But, that's what the spec says to do; perhaps 2.0 can roll that back.

tests/expand-manifest.html
```diff
@@ -5353,6 +5353,27 @@ <h2>
 </dd>
 </dl>
 </dd>
+<dt id='te052'>
```
davidlehn: Missed an html rebuild.

gkellogg:
I'm rebuilding them all as part of PR #416.

PR #430, opened 2020-03-24
Verify URI resolution relative to base without trailing slash

fsteeg:
This came up when fixing https://github.com/jsonld-java/jsonld-java/issues/279 in https://github.com/jsonld-java/jsonld-java/pull/280, where we noticed the original issue described on [the Apache Jena list](https://lists.apache.org/thread.html/r343f27b3390dfbfd7ce9dcee4576bc532705b39d26923dfb0baf5322%40%3Cusers.jena.apache.org%3E) was not covered by the spec tests.

Relative resolution in RFC 3986 [1][2] (see `8)` in spec [3]):

- `rel` relative to `example` is `example/rel`
- `rel` relative to `example/base` is also `example/rel`

Already tested elsewhere, for clarity here, with trailing slash:

- `rel` relative to `example/base/` is `example/base/rel`

[1] https://tools.ietf.org/html/rfc3986#section-5.2
[2] https://blog.cdivilly.com/2019/02/28/uri-trailing-slashes
[3] https://w3c.github.io/json-ld-api/#algorithm-4

### DIFF
--- a/tests/expand-manifest.html
+++ b/tests/expand-manifest.html
@@ -3170,6 +3170,48 @@ Test t0128 Two scoped context may include a shared context
 </dd>
 </dl>
 </dd>
+<dt id='t0129'>
+Test t0129 Base without trailing slash, without path
+</dt>
+<dd>
+<dl class='entry'>
+<dt>id</dt>
+<dd>#t0129</dd>
+<dt>Type</dt>
+<dd>jld:PositiveEvaluationTest, jld:ExpandTest</dd>
+<dt>Purpose</dt>
+<dd>Verify URI resolution relative to base (without trailing slash, without path) according to RFC 3986</dd>
+<dt>input</dt>
+<dd>
+<a href='expand/0129-in.jsonld'>expand/0129-in.jsonld</a>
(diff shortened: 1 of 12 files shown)


gkellogg:
0130 seems to be similar to [toRdf/0128](https://w3c.github.io/json-ld-api/tests/toRdf-manifest.html#t0128), but there isn't an expand version of this AFAICT.

Ruby passes them all okay, but pyld fails 0129, so clearly testing a corner case. Looks like the playground (jsonld.js) isn't passing 0129 either.

We'll need toRdf versions of these as well, and both expand-manifest.html (and toRdf-manifest.html) should be updated using the `rake` task, but if you have problems with this, I can update after the fact.

gkellogg:
Also, note that toRdf versions should be named "e129" and "e130" to match the naming scheme for other expansion tests brought over.

fsteeg:
Thanks @gkellogg, I've added toRdf tests and the `rake` output. Hope I got it right.

gkellogg:
Thanks for the extra effort, @fsteeg!

PR #431, opened 2020-03-26
Add jsonld-streaming-serializer implementation report

rubensworks:
This PR adds the implementation report for [jsonld-streaming-serializer](https://github.com/rubensworks/jsonld-streaming-serializer.js) with updated 1.1 support.

Note that not all tests pass. This is due to the serializer not supporting serialization using the `@list` keyword, and compound literals for `rdfDirection`. As far as I can see, these can not be implemented in a fully streaming manner.


gkellogg:
We're going to freeze the test suite shortly, and the reports manifests.nt file needs to be regenerated. I'll regenerate that now, presuming that now tests are coming, and you can re-run?

gkellogg:
Fixed by #444.

PR #440, opened 2020-03-29
Various

### LINKED ISSUE #434: compaction implementation difficulties
Hi,

as I'm trying to implement the compaction algorithm, I'm still having some troubles related to tests, and mostly with container mappings. Here is an example, with test t0029. The context is the following:

```json
{
  "@context": {
    "author": {"@id": "http://example.com/vocab/author", "@container": "@index" }
  }
}
```

and the inverse context (from my implementation) is:

```json
{
  "http://example.com/vocab/author": {
    "@index": {
      "@language": {"@none": "author"},
      "@type": {"@none": "author"},
      "@any": {"@none": "author"}
    }
  }
}
```

The data to compact is an array of two elements of this form:

```
{ "http://example.com/vocab/author":
  [ {"@index": "something", "@id": "somethingelse"}, … ]}
```

For the author key, I have to execute step 12.8 of the compaction algorithm, that uses the iri compaction algorithm to compact "http://example.com/vocab/author". Expanded item is the first element of the array above: `{"@index": "something", "@id": "somethingelse"}`. We try to iri compact `http://example.com/vocab/author` with that element.

In the iri compaction algorithm, vocab is true and var is an entry in inverse context, so we execute step 4. Some of the steps add values to containers. Initially, it is empty:

- 4.5: value contains `@index`, but it is not a map, so nothing to do
- 4.6: reverse is false, so nothing to do
- 4.7: value is not a list object, so skip
- 4.8: value is not a graph object, so skip
- 4.9: none of the above cases matched, so we execute this.
  - 4.9.1: value is not a value object, so skip
  - 4.9.2: we add `@id`, `@id@set`, `@type`, and `@set@type` to containers
  - 4.9.3: we add `@set` to containers
- 4.10: we add `@none` to containers
- 4.11: processing mode is indeed 1.1, but value is a map and does contain an `@index` key, so skip
- 4.12: processing mode is indeed 1.1, but value is a map that doesn't only contain `@value`, so skip

In the end, containers contains `[@id, @id@set, @type, @set@type, @set, @none]`. None of them are the expected `@index` which is the only one that can give the correct compaction.

### LINKED ISSUE #436: graph compaction issues
I'm still having issues with compaction. This time, compacting graphs is the problem. I take 0077 as an example. Here, I need to compact `{"input": {"@graph": [{"@value": "x"}]}}`. where `input` has `@graph` in its container mapping. The expected result is `{"input": {"@value": "x"}}`, but I can't remove the `@graph` in my implementation.

When running step 12.8.6, I recursively call the compaction algorithm on `{"@graph": [{"@value": "x"}]}` with `input` as the active property. This recursive call returns the `@graph` object, not compacted.

Note that here the specification says "[passing] expanded item for element if it does not contain the entry `@list` and is not a graph object containing `@list`, otherwise pass the value of the entry for element.".  This is confusing (that last part sounded like pass the value associated with element). Reversing the sentence might make it more readable, like so: "If expanded item does not contain the entry `@list` and is not a graph object containing `@list`, pass expanded item for element, otherwise pass the value of the `@list` entry for element."

Continuing execution, I arrive at step 18.8.8.3 because the expanded item is a graph object, and container only includes `@graph`. There I simply add compacted item to result, so I have `@graph` in the result.

I think there are two points where that could have gone wrong: either step 12.8.6 where I should actually have passed only the content of the graph object, or step 12.8.8 where I should have set the result to the content of the graph object, instead of the graph object itself.

### LINKED ISSUE #437: compacting types with mode 1.0
Test 0106 is entitled "Ensures that a single @type value is not represented as an array in 1.0". However, my implementation returns the type as an array with one element in this test.

First of all, the expansion algorithm already returns an array in that case, because of step 16 of the expansion algorithm. Skipping that step when processing mode is 1.0 doesn't trigger any test failure.

Then, when compacting a type in 12.2, as array is set to true, because type's container mapping contains `@set`, so the result is an array, whether or not the expansion gave a string or an array.

### LINKED ISSUE #438: compaction: remaining singleton values
In steps 12.8.9.6 and 12.8.9.8, we examine the values associated with a key. There can be one or more values. The first value is used, and when there are remaining values, the specification asks to set a value is compacted item to those remaining values.

However, it is not clear what to do when there is only one remaining value. As I understood it, the specification says to use an array even for a single value. However, in test m004, there are two types for each object. The remaining value is therefore an array with one value, but the expected output is a single value, not an array.

The specification should say what to do when there is only one remaining value. For 12.8.9.8, I followed test m004 and used the value instead of the array in my implementation, and I think the specification should say so too. I'm not sure about 12.8.9.6.

### LINKED ISSUE #439: compaction: type-scoped context
In test [c009](https://w3c.github.io/json-ld-api/tests/compact/c009-in.jsonld), we test that the type-scoped context is not propagated to nested nodes. However, I think the specification doesn't prevent it. When we examine the object, we execute step 11 which updates active context. This active context is passed to nested nodes. In the nested nodes, the active context contains a definition for baz whose type is vocab, leading to a different IRI in the result.

I believe that passing false for the propagate flag when calling the context processing algorithm in step 11 is enough, as the original context will be saved in the previous context of the new active context. When recursing on the compaction algorithm, step 5 will ensure that we use the original context instead of the new one in nested nodes.

This change helped me pass 8 tests (4 of which had additional issues that I now fixed) and didn't create any new test failure.

gkellogg:
* Update IRI Compaction to use `@index` for any value with an `@index` entry. Fixes #434.
* Update step 12.2.4 of the Compaction Algorithm to only look for `@set` if _processing mode_ is json-ld-1.1. Update the description of the _add\_value_ macro. Fixes #437.
* Update step 12.8.6 of the Comapation Algorithm to to clarify the value passed for _element_. Fixes #436.
* Update steps 12.8.9.6.3 and 12.8.9.2.2 of the Compaction Algorithm to invoke the _add value_ macro for adding remaining values back to _compacted item_. Fixes #438.
* Update step 11.1 of the Compaction Algorithm to pass `false` for _propagate_ when calling theContext Processing algorithm. Fixes #439.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/440.html" title="Last updated on Mar 30, 2020, 5:44 PM UTC (b63a6f9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/440/d519b2c...b63a6f9.html" title="Last updated on Mar 30, 2020, 5:44 PM UTC (b63a6f9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3515,7 +3515,7 @@
           Then, for each <var>term</var>
           in <var>compacted types</var> ordered lexicographically:
           <ol>
-            <li>If the <a>term definition</a> for <var>term</var> in <var>type-scoped context</var> has a
+            <li id="alg-compact-11_1">If the <a>term definition</a> for <var>term</var> in <var>type-scoped context</var> has a
               <a>local context</a>
               set <var>active context</var> to the result of the
               <a href="#context-processing-algorithm">Context Processing algorithm</a>,
@@ -3523,7 +3523,8 @@
               <a>local context</a> in <var>type-scoped context</var> as <var>local context</var>
               <span class="changed">
                 <var>base URL</var> from the <a>term definition</a> for <var>term</var>
-                in <var>type-scoped context</var></span>.
+                in <var>type-scoped context</var>,
+                and `false` for <var>propagate</var></span>.
             </li>
           </ol>
         </li>
@@ -7016,6 +7017,10 @@
       This simplifies calling sequences and better represents actual implementation experience.</li>
     <li>Updated step <a href="#alg-iric-value-map">4.5</a> of the <a href="#iri-compaction">IRI Compaction algorithm</a>
       to use `@index` for any value with an `@index` entry.</li>
+    <li>Update step <a href="#alg-compact-11_1">11.1</a> of the
+      <a href="#compaction-algorithm">Compaction Algorithm</a>
+      to pass `false` for <var>propagate</var> when calling the
+       <a href="#context-processing-algorithm">Context Processing algorithm</a>.</li>
     <li>Updated step <a href="#alg-compact-12_2_4">12.2.4</a> of the
       <a href="#compaction-algorithm">Compaction Algorithm</a>
       to only look for `@set` if <a>processing mode</a> is json-ld-1.1.</li>
--- a/index.html
+++ b/index.html
@@ -3568,8 +3568,9 @@
                 </li>
                 <li>Initialize <var>alias</var>
                   <span class="changed">by <a>IRI compacting</a> <var>expanded property</var></span>.</li>
-                <li class="changed">Initialize <var>as array</var>
-                  to <code>true</code> if the <a>container mapping</a> for <var>alias</var> in the
+                <li id="alg-compact-12_2_4" class="changed">Initialize <var>as array</var>
+                  to <code>true</code> if <a>processing mode</a> is `json-ld-1.1` and
+                  the <a>container mapping</a> for <var>alias</var> in the
                   <var>active context</var> includes <code>@set</code>,
                   otherwise to the negation of {{JsonLdOptions/compactArrays}}.</li>
                 <li class="changed">Use <a>add value</a> to add <var>compacted value</var>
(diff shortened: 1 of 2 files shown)


gkellogg:
cc/ @ropat.

index.html
```diff
@@ -3707,16 +3709,16 @@ <h3>Algorithm</h3>
...
-                  <span class="changed">and is not a <a>graph object</a> containing <code>@list</code></span>,
-                  otherwise pass the value of the <a>entry</a> for <var>element</var>.
-                  <span class="changed">Along with the {{JsonLdOptions/compactArrays}}
-                    and {{JsonLdOptions/ordered}} flags</span></li>
+                  <var>expanded item</var> for <var>element</var>,
+                  <span class="changed">along with the {{JsonLdOptions/compactArrays}}
+                    and {{JsonLdOptions/ordered}} flags</span>.
+                  If <var>expanded item</var> is a <a>list object</a> or a <a>graph object</a>,
+                  use the value of the `@list` <span class="changed">or `@graph`</span> entries,
+                  respectively, for <var>element</var> instead of <var>expanded item</var>.</li>
```
gkellogg: @ropat these are the lines that fix #436.

PR #442, opened 2020-04-01
Add tests for invalid prefix values, Closes #432

### LINKED ISSUE #432: Missing tests with expected error code
For reference, tests for the following error codes are missing:

* ~~`IRI confused with prefix`: Present for compact, but not for fromRdf.~~
* `invalid @prefix value`: Missing for both expand and toRdf.
* `context overflow`: Missing for both expand and toRdf.

(I only check the compact, expand, toRdf and fromRdf manifests)

rubensworks:
As discussed in #432.

### DIFF
--- a/tests/expand-manifest.html
+++ b/tests/expand-manifest.html
@@ -5845,6 +5845,27 @@ invalid term definition
 </dd>
 </dl>
 </dd>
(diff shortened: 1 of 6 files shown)


gkellogg:
Fixes #432.

PR #447, opened 2020-04-04
Json gold EARL report

gkellogg:
Currently shows FAIL for all HTML tests, could probably use `earl:untested` or `earl:inapplicable` for things not specifically intended to pass.

cc/ @ kazarena

### DIFF
--- a/reports/jsonld-gold-earl.ttl
+++ b/reports/jsonld-gold-earl.ttl
@@ -4,21 +4,27 @@
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
-<https://github.com/piprate/json-gold> a doap:Project,
-    <http://www.w3.org/ns/earl#TestSubject>,
+<https://github.com/piprate/json-gold/tree/v0.3.0> a doap:Version;
+  doap:created "2020-04-06"^^xsd:date;
+  doap:name "json-gold-v0.3.0";
+  doap:revision "v0.3.0" .
+
+<https://github.com/piprate/json-gold> a <http://www.w3.org/ns/earl#TestSubject>,
+    doap:Project,
     <http://www.w3.org/ns/earl#Software>;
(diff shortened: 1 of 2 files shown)


reports/jsonld-gold-earl.ttl
```diff
@@ -0,0 +1,18485 @@
...
+  foaf:homepage <https://github.com/kazarena>;
+  foaf:name "Stan Nazarenko" .
+
+ [
+    a <http://www.w3.org/ns/earl#Assertion>;
+    <http://www.w3.org/ns/earl#assertedBy> <https://github.com/kazarena>;
+    <http://www.w3.org/ns/earl#mode> <http://www.w3.org/ns/earl#automatic>;
+    <http://www.w3.org/ns/earl#result> [
+      a <http://www.w3.org/ns/earl#TestResult>;
+      dc:date "2020-04-06T17:15:22.676634"^^xsd:date;
```
davidlehn, line 37: @gkellogg Shouldn't these be xsd:dateTime?  Or if it's xsd:date then truncated to just yyyy-mm-dd?

gkellogg: The range of `dc:date` is `rdfs:Literal`, so it doesn't really matter from that perspective, and it's not used in the report generation in any case.

gkellogg:
From the description for `dc:date`:

> Date may be used to express temporal information at any level of granularity.  Recommended practice is to express the date, date/time, or period of time according to ISO 8601-1 [[ISO 8601-1](https://www.iso.org/iso-8601-date-and-time-format.html)] or a published profile of the ISO standard, such as the W3C Note on Date and Time Formats [[W3CDTF](https://www.w3.org/TR/NOTE-datetime)] or the Extended Date/Time Format Specification [[EDTF](http://www.loc.gov/standards/datetime/)].  If the full date is unknown, month and year (YYYY-MM) or just year (YYYY) may be used. Date ranges may be specified using ISO 8601 period of time specification in which start and end dates are separated by a '/' (slash) character.  Either the start or end date may be missing.

PR #457, opened 2020-04-10
CSS trick to visually mark failing tests

pchampin:
"failing" here means: "passed by less than 2 implementations"


gkellogg:
I bow to your CSS selector mastery! 😃

PR #463, opened 2020-04-12
Update implementation report for guile-jsonld

roptat:


iherman:
Mark it as non-substantive because there is no change on standard text. This will settle the IPR issue.

PR #465, opened 2020-04-15
Update PyLD test report for 2.0.0.

davidlehn:
- Renamed report since it's easier to create api+framing reports now.
- I generated output index.html and it looks ok, I think.  Lots of data, hard to check it all by sight.
- My tools had odd parsing errors rebuilding manifest.nt, I'm not sure how that effects things.
- Generating earl.jsonld and index.html caused huge diff bloat due to non-deterministic ordering.  So I left them out of this PR for now. Do the tools allow output property sorting to avoid that?  I'm not sure about the jsonld part, but perhaps the html template itself could do some of that?
- What time resolution is needed for these tests? I dropped the subsecond part for PyLD tests.  Seems like just the day would be enough for this purpose.  That could help a bit with any daily report update diff churn and make files a bit smaller.

### DIFF
--- a/reports/pyld-earl.ttl
+++ b/reports/pyld-earl.ttl
@@ -4,12 +4,12 @@
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
-<https://github.com/digitalbazaar/pyld> a doap:Project,
-    <http://www.w3.org/ns/earl#Software>,
-    <http://www.w3.org/ns/earl#TestSubject>;
+<https://github.com/digitalbazaar/pyld> a <http://www.w3.org/ns/earl#TestSubject>,
+    doap:Project,
+    <http://www.w3.org/ns/earl#Software>;
   dc:title "PyLD";
   dc:creator <https://github.com/dlongley>;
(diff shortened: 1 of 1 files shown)


gkellogg:
> * Renamed report since it's easier to create api+framing reports now.
> * I generated output index.html and it looks ok, I think.  Lots of data, hard to check it all by sight.

The main thing would be to see if the numbers reported in the summary section for PyLD look right.

> * My tools had odd parsing errors rebuilding manifest.nt, I'm not sure how that effects things.

This shouldn't be necessary for submitting a report. There is a rake task for building _manifests.nt_, but that should only be necessary when there are changes to test manifests.

> * Generating earl.jsonld and index.html caused huge diff bloat due to non-deterministic ordering.  So I left them out of this PR for now. Do the tools allow output property sorting to avoid that?  I'm not sure about the jsonld part, but perhaps the html template itself could do some of that?

The output is compacted using the algorithm, so properties should be sorted as specified by the compaction algorithm. That said, producing output with the minimum changes is not a goal.

> * What time resolution is needed for these tests? I dropped the subsecond part for PyLD tests.  Seems like just the day would be enough for this purpose.  That could help a bit with any daily report update diff churn and make files a bit smaller.

Time is not used in the output; it's a useful field in the EARL format, but not used in generating the implementation report.

I'll take care of regenerating the report.

Issue #480, opened 2020-04-23
Is the possibility of having `@id: null` in a expanded document intended?


timothee-haudebourg:
In the playground implementation, using a keyword-like value as `@id` causes the node id to explicitly expand to `null`. I see why it happens since it is the result of the IRI expansion. However this causes the expanded document to be an invalid JSON-LD document, since putting `null` for `@id` is forbidden.

[Here is an example where the expanded output is an invalid input](https://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40id%22%3A%22foo%22%2C%22scheme%3A%2F%2Ffoo%22%3A%7B%22%40id%22%3A%22%40foo%22%7D%7D).

gkellogg:
This could be addressed by merging steps 13.4.3.1 and 13.4.3.2 in the Expansion algorithm, to check that the result of IRI expanding the value is a string, which it shouldn't be because the `@foo` value gets ignored and replaced with `null`. The same might be said of `@type`, but as this is an odd-case, it might be simpler to add a blanket statement that if keyword-like values are ignored, they should not be used for adding values to results on expansion, without trying to track down every place that would check before and after expansion.

In this case, it's either an error, or no `@id` property should be added to the result.

timothee-haudebourg:
This odd case happens in the [expansion test 122](https://w3c.github.io/json-ld-api/tests/expand/0122-in.jsonld). The [output of this test](https://w3c.github.io/json-ld-api/tests/expand/0122-out.jsonld) is not a valid JSON-LD document.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2020/2020-04-24-json-ld#section3-2).

- `RESOLVED:  Defer api #480 until future version, as error case of assigning a non-URI as a URI leads to a further (worse) error case, and the solution would be far reaching`
<details><summary><i class="differentiate">View the transcript</i></summary>
Rob Sanderson: <a href="https://github.com/w3c/json-ld-api/issues/480">https://github.com/w3c/json-ld-api/issues/480</a><br/>
Gregg Kellogg: This is an implication of ignoring keyword-like things.<br/>
… The intention is that if you see something that looks like a keyword to be ignired, it would not result in any properties that would be added to the expansion output.<br/>
… In this case, this reveals that there may be other cases requiring a more substantial change. So we may have to work with this, and tackle it in the future.<br/>
[...]

Issue #473, opened 2020-04-17
Lists, sets, and multisets


timothee-haudebourg:
Hi it's me again,

I am confused about the use of the term "set" to describe `@set` objects. In the section [Lists and Sets](https://www.w3.org/TR/json-ld11/#lists-and-sets) it is said that
> A set represents an unordered set of values

Section [4.3.2](https://www.w3.org/TR/json-ld11/#sets) it is said that 
> While `@list` is used to describe ordered *lists*, the `@set` keyword is used to describe unordered **sets**.

So I thought to be correct assuming that I could optimize `@set` objects by removing duplicate values (using a set data structure internally), since it is described as an unordered **set**, a opposed to `@list` objects defined as an ordered **list**.

Now trying to pass the expansion tests, I cannot pass [test 27](https://w3c.github.io/json-ld-api/tests/expand/0027-in.jsonld) as it is indicated that "Duplicate values in `@list` and `@set` are not merged" which is not mentioned in the specification sections associated to `@list` and `@set`, and also means that a `@set` object is, in fact, not a set, but a [multiset](https://en.wikipedia.org/wiki/Multiset).

Note that I understand that `@set` is just syntactic sugar, associating multiple values to a subject property. Forcing an implementation to keep duplicate values in `@set` objects implies that duplicate RDF triples would have some kind of meaning, which they have not since a RDF graph is a *set* of triples, not a multiset.

Am I correct in my conclusion or is there something wrong with the expansion test 27? Or did I missed the explanation in the spec? Or is it just to avoid further complications (like, what should we do if there is a duplicate occurrence of the same node object, with the same id but not the same properties, etc.)

gkellogg:
The word _set_ is, indeed, something of a misnomer. Really, the distinction is between order-preserving and non-order-preserving arrays (or multiset, as you note). A true _set_ could not allow duplicates.

Although non-list arrays of JSON values do maintain their distinctness in the JSON-LD realm, they do not round-trip through RDF, and strictly speaking, this does not really conform to the RDF Graph model even in the pure JSON-LD realm. But, compatibility with JSON-LD 1.0 restricts our ability to do anything here, other than perhaps to add such a disclaimer.

I'd defer to a future version any change the semantics of `@set` (explicit or implied), but that's for the group to discuss.

[...]

PR #481, opened 2020-04-27
Timothee updates

### LINKED ISSUE #475: Create Term Definition algorithm, possible imprecision step 11
Step 11 of the [Create Term Definition Algorithm](https://www.w3.org/TR/json-ld11-api/#create-term-definition), it is said that
> If the `@protected` entry in `value` is `true` set the `protected` flag in `definition` to `true`

Strictly following this, a `"@protected": false` in a term definition is ignored.
However, this makes the [expansion test `pr13`](https://w3c.github.io/json-ld-api/tests/expand/pr13-in.jsonld) to fail.

### LINKED ISSUE #477: Expansion Algorithm confusing Step 13.8.3.7.2.3 not covered by the expansion tests
In the [Expansion Algorithm](https://www.w3.org/TR/json-ld11-api/#expansion-algorithm) I personally find the description of the Step 13.8.3.7.2.3 hard to understand:
> Initialize `index property values` to the concatenation of `re-expanded index`
with any existing values of `expanded index key` in `item`.

The variable `re-expanded index` holds a value object. I don't understand how to "concatenate" a value object with anything. I find the use of "any" confusing, does it mean "all", or "the first one you find"? Moreover, `expanded index key` is most likely to be an expanded IRI and I don't see any situation where it would appear in `item` (unless we consider the expanded keys in `item`? I'm lost).

In my implementation, I temporarily simplified this step by just defining `index property values` as `re-expanded index`. My plan was to come back to it while running the expansion tests, trying to understand reading the tests.

However I now pass all the tests without changing this part...

### LINKED ISSUE #478: What is the purpose of `from map` in the Expansion Algorithm?
The Expansion Algorithm requires a boolean `from map` variable as input. Its value is never changed across the expansion algorithms and is always set to `false`. There are only 3 occurrences of this variable in the whole API document. It doesn't seem to be used.

### LINKED ISSUE #479: Expansion Algorithm, possible imprecision step 13.4.6.2
While expanding `@included` step 13.4.6.2 it is said
> Set expanded value to the result of using this algorithm recursively passing active context, active property, value for element, base URL, and the frameExpansion and ordered flags, ensuring that the result is an array.

I believe the implementation is supposed to pass `@included` as active property instead of passing the current value. Otherwise, free floating literals inside the included block are automatically dropped, which makes the test [`expand/in07`](https://w3c.github.io/json-ld-api/tests/expand/in07-in.jsonld) to fail.

gkellogg:
* Add note that unordered arrays may contain duplicate values after expansion. For #473. Other parts of #473 are deferred.
* Updated step 11 of the Create Term Definition algorithm to use any boolean value of `@protected`, not just `true`. Fixes #475.
* Improve wording of expansion 13.8.3.7.2.3. Fixes #477.
* Update step 13.8.3.6 of the Expansion algorithm to pass `true` for the from map parameter to properly manage reverting active contexts. Fixes #478.
* Update step 13.4.6.2 of the Expansion algorithm to pass `null` for active property, as included blocks do not define a relationship to a referencing node. Fixes #479.
* Mark expand/0122 non-normative, as it results in an invalid JSON-LD result. Note: toRdf/e122 does not include a triple for this, so it remains normative. For #480.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/481.html" title="Last updated on Apr 27, 2020, 6:52 PM UTC (9bc2f71)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/481/c512a0d...9bc2f71.html" title="Last updated on Apr 27, 2020, 6:52 PM UTC (9bc2f71)">Diff</a>


gkellogg:
@iherman note that this updates the implementation report which now considers expand/0122 to be non-normative, as the resulting file is not legal JSON-LD.

gkellogg:
@iherman So, does that mean you'd like to back out the change that affects the implementation report, or just be sure it's in prior to acceptance?

gkellogg:
@iherman: we have agreement on everything, we can pull this PR pending your clarification to changing the implementation report to show the affected test as being non-normative. If you don't want this change, I can back out the commit.

iherman:
@gkellogg I am less worried about the implementation report, as long as it does not change our exit criteria results:-). We did agree (and I told the Director) that we intend to add new implementations, for example, to the report. I am worried about the specifications themselves.

I intend to set up the errata page/script for all documents this week. We should begin to use that if needed...

gkellogg:
> @gkellogg I am less worried about the implementation report, as long as it does not change our exit criteria results:-). We did agree (and I told the Director) that we intend to add new implementations, for example, to the report. I am worried about the specifications themselves.

IIRC, we agreed to make these changes to the spec, are you saying we shouldn’t pull the PR now? I’m confused.

iherman:
>> @gkellogg I am less worried about the implementation report, as long as it does not change our exit criteria results:-). We did agree (and I told the Director) that we intend to add new implementations, for example, to the report. I am worried about the specifications themselves.
>
> IIRC, we agreed to make these changes to the spec, are you saying we shouldn’t pull the PR now? I’m confused.

We are talking one another... reset:-) What I mean is:

- merge this PR, meaning the report as well as the spec gets a minor update
- no more PR-s on the spec
- PR-s on the report are o.k. for the future, too.

I am afraid we are in a violent agreement...

gkellogg:
Okay, thanks!

PR #483, opened 2020-04-30
Added the errata management file

iherman:
The file will be served from 

https://w3c.github.io/json-ld-api/errata/

(although we may want to set up redirections from W3C, to be discussed...)

See: https://raw.githack.com/w3c/json-ld-api/add-errata/errata/index.html for now.

If this is fine, the same file will be added to the syntax and framing specs (maybe to streaming?), with obvious copy-paste actions.

### DIFF
--- a/errata/errata.css
+++ b/errata/errata.css
@@ -1,0 +1,178 @@
+div#toc a, section#toc a {
+	text-decoration: none;
+}
+
+div#toc a:hover, section#toc a:hover {
+	background: yellow;
+	color: red;
+}
+
+div#toc ul, section#toc ul {
+	list-style-type: none;
+	margin-left: 1.5em;
+	padding-left: 0;
+}
+
+.tocvisible, .tochidden { cursor: pointer; }
+.tocvisible:before {
+	content: "▾ ";
+	font-size: 110%; 
+	font-family: monospace;
+}
+.tochidden:before  { 
+	content: "‣ "; 
+	font-size: 110%; 
+	font-family: monospace;
+}
+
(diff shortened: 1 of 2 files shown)


davidlehn:
The errata.css that's being pulled in sets line-height on body.  That makes the long h1 title look bad when it wraps.  The bottom of 1st line touches top of 2nd line.  May deserve an issue in another repo.

Are the accepted "Errata" issues supposed to be kept open?  A bit unclear in the text.  I'm guessing the search will find open or closed with the proper label.

azaroth42:
:+1: to @davidlehn.  I see it as:

Screen Shot 2020-04-30 at 1 29 46 PM

when I resize the browser width down

iherman:
> The errata.css that's being pulled in sets line-height on body. That makes the long h1 title look bad when it wraps. The bottom of 1st line touches top of 2nd line. May deserve an issue in another repo.
> 
> Are the accepted "Errata" issues supposed to be kept open? A bit unclear in the text. I'm guessing the search will find open or closed with the proper label.

I have made an extra commit:

- Added a local version of the css file and removed the line-height setting (no idea why that would be necessary
- Made the title shorter because it was too long anyway
- I have also added an explanation to the text, which says:

> All issues, labeled by “Errata”, are displayed in this report, whether they are opened or closed. Their status is added to the report on the individual errata.

iherman:
> Apart from having a section 1 and no section 2 (which I can live with), looks ok.

Yep; I have actually removed the TOC. For a single document case it does not make sense.

PR #488, opened 2020-05-12
fix version number in sophia report

pchampin:
I could not update the reports;
the command `rake report/manifests.nt` yields the following error:

>  The git source https://github.com/ruby-rdf/json-ld.git is not yet checked out. Please run `bundle install` before trying to start your application

even though I *did* run bundle install (with success).

### DIFF
--- a/reports/README.md
+++ b/reports/README.md
@@ -9,7 +9,7 @@ Run it as follows:
 
 ```sh
 $ gem install earl-report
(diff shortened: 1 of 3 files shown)


gkellogg:
I'll look into it, but  you shouldn't need to update report/manifests.nt unless any of the JSON-LD API or Framing test suites are modified. But, the proper command would be `rake reports/manifests.nt`, as it's the `reports` directory, not the `report` directory.

PR #489, opened 2020-05-12
Add tests to make sure that value objects ...

### LINKED ISSUE #484: Compacting IRIs associated to multi types values/nodes
The type mapping of a term definition in a context must be a string, which means it is a single type. So in an inverse context, we have maps from single types to terms. However in an expanded document, node objects and value objects may have multiple types, stored in an array.

I'm wondering how this is handled during the compaction phase. While reading the IRI compaction algorithm, I can imagine cases where the `type/language value` variable is set to an array of types, which in turns will be part of the `prefered values` when calling the Term Selection Algorithm.

Is this intended? I guess it is not strictly incorrect since the provided algorithm is untyped, the array would just be ignored during the term selection. Or should I read the verb "append" in the step 4.17 of the IRI compaction step as "concatenate" if it an array?

gkellogg:
and expanded term definitions have only a single value for `@type`, unless framing in the case of value objects.

Fixes #484.

cc/@timothee-haudebourg


codedfiles:
Duplicate of #

PR #490, opened 2020-05-14
Empty-property scoped context does not affect term selection.

gkellogg:
Adding a minimal/empty property-scoped context should not affect the using terms defined in outer context.

From https://github.com/w3c/wot-thing-description/issues/894.


davidlehn:
Perhaps also add an even more minimal test.  Surprised this wasn't covered before.
in:
```json
{
  "ex:title": "top",
  "ex:thing": {
    "ex:title": "sub"
  }
}
```
context:
```json
{
  "@context": {
    "thing": {
      "@id": "ex:thing",
      "@context": {}
    },
    "title": "ex:title"
  }
}
```
out:
```json
{
  "@context": {
    "thing": {
      "@id": "ex:thing",
      "@context": {}
    },
    "title": "ex:title"
  },
  "title": "top",
  "thing": {
    "title": "sub"
  }
}
```
And while we're at it, how about a basic expand test that jsonld.js also is failing now.  How was this also not covered?
in:
```json
{
  "@context": {
    "thing": {
      "@id": "ex:thing",
      "@context": {}
    },
    "title": "ex:title"
  },
  "title": "top",
  "thing": {
    "title": "sub"
  }
}
```
out:
```json
[
  {
    "ex:title": [
      {
        "@value": "top"
      }
    ],
    "ex:thing": [
      {
        "ex:title": [
          {
            "@value": "sub"
          }
        ]
      }
    ]
  }
]
```

davidlehn:
This is named with "minimal/empty" contexts, but in the case of jsonld.js, I think any non-null context data fails.  This may be a general additive context issue.

PR #500, opened 2020-06-25
Make a global JsonLd interface.

gkellogg:
This silences ReSpec, and is probably better than just removing [Exposed] altogether.

Thanks to @hober for the advice.

If acceptable, a similar change will be required for JSON-LD Framing.

CC/ @swickr @iherman

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/500.html" title="Last updated on Jun 26, 2020, 6:57 PM UTC (95a5df8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/500/e4e8dc5...95a5df8.html" title="Last updated on Jun 26, 2020, 6:57 PM UTC (95a5df8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5662,7 +5662,7 @@
       as long as they generally use the same methods, arguments, and options
       and return the same results.</span></p>
 
-  <p class="note">Interfaces are marked `[Exposed=Endpoint]`,
+  <p class="note">Interfaces are marked `[Exposed=JsonLd]`,
     which creates a global interface.
     The use of WebIDL in JSON-LD, while appropriate for use within browsers,
     is not limited to such use.</p>
@@ -5688,12 +5688,12 @@
 
     <pre class="idl">
       /*
-       * The Endpoint interface is created to expose the JsonLdProcessor interface.
+       * The JsonLd interface is created to expose the JsonLdProcessor interface.
        */
-      [Global=Endpoint, Exposed=Endpoint]
-      interface Endpoint {};
+      [Global=JsonLd, Exposed=JsonLd]
+      interface JsonLd {};
 
-      [Exposed=Endpoint]
+      [Exposed=JsonLd]
       interface JsonLdProcessor {
         constructor();
         static Promise&lt;JsonLdRecord> compact(
@@ -6092,7 +6092,7 @@
       which has a <a>default graph</a> accessible via the <a data-link-for="RdfDataset">defaultGraph</a> attribute.</p>
 
     <pre class="idl changed">
-      [Exposed=Endpoint]
+      [Exposed=JsonLd]
       interface RdfDataset {
         constructor();
         readonly attribute RdfGraph defaultGraph;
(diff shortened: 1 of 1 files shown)


hober:
I think this isn't quite what I meant either. I wasn't suggesting that JsonLdProcessor be the global itself, just that you need to define a global and then expose all of your WebIDL interfaces (including JsonLdProcessor, I assume) on it.

gkellogg:
So, would you creat an empty interface “Thingy”, and then the main JsonLdProcecessor interface exposed to “Thingy”, or is there some way to do this without an empty interface? Or, what might go into such an interface?

gkellogg:
So, would you creat an empty interface “Thingy”, and then the main JsonLdProcecessor interface exposed to “Thingy”, or is there some way to do this without an empty interface? Or, what might go into such an interface?

gkellogg:
I'm happy to bikeshed the use of "External" as a global interface.

iherman:
@gkellogg I trust you and @hober that this is correct WebIDL; but I do like the "melody" of it:-)

It is indeed bikeshedding, but why not just use `JSONLD` (with some form of a capitalization that WebIDL accepts) as the "endpoint"? It seems more natural...

hober:
Yeah, I was proposing an "empty" global, and @iherman's idea to call it JSONLD sounds reasonable to me.

dlongley:
More bikeshedding: Since this interface is generally specified on a "module" or "library" perhaps using a name like that would work. I don't have strong opinions here.

gkellogg:
> More bikeshedding: Since this interface is generally specified on a "module" or "library" perhaps using a name like that would work. I don't have strong opinions here.

I went with "JsonLd". Using "module" or "library" might be a little broad. Perhaps that's where things will end up, but I think that's a larger conversation than just JSON-LD. It does make the interfaces seem a bit redundant, but it does create a pattern other specs might follow, at least until its properly addressed in WebIDL, presuming the community goes in that direction.

With some approvals, I'll do the same thing to json-ld-framing, and merge this to master.

iherman:
@BigBlueHat it would be great if you could also explicitly signal that you approve this change before we try to make it official... @azaroth42, even if you are not formally a chair any more, the same would hold for you

azaroth42:
Yes, this seems like an excellent solution to me!

davidlehn:
I'm late to the discussion.  If anyone has advice on how this change effects the out-of-date jsonld.js webidl tests (or the jsonld.js api itself), please advise: https://github.com/digitalbazaar/jsonld.js/issues/359.

gkellogg:
The interface has always been JsonLdProcessor, going back to 1.0. In JSON-LD 1.0, WebIDL didn't have any Exposed attribute, so it wasn't relevant then. It is now, as a global, but not sure how that would actually impact an implementation.

But, it would be good if the WebIDL interface conformed to the new specification, for some value of "conform".

PR #516, opened 2020-12-17
Rename duplicate ids in toRdf manifest

clehner:
The `toRdf` tests manifest contains multiple entries for ids `#t0124` and `#t012t`. I rename them to `#te124` and `#te125` as that looks like how they are supposed to fit in with the sequence of other tests.

I checked the other test manifests and did not find any other deduplicate ids. Helper script: [dedup.js](https://github.com/w3c/json-ld-api/files/5711467/dedup.js.txt)

### DIFF
--- a/tests/toRdf-manifest.html
+++ b/tests/toRdf-manifest.html
@@ -5626,13 +5626,13 @@ invalid typed value
 </dd>
 </dl>
 </dd>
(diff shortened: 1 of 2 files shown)


gkellogg:
Good catch, thanks.

PR #538, opened 2022-03-01
Fix link to test suite on github

Khady:


pchampin:
I believe that the modified file is automatically generated, so the change should rather be made to `template.haml`. @gkellogg do you confirm?

gkellogg:
Yes, this would be in the manifest template, and affect all html versions of test manifests. The proper fix is the update the template and regenerate the manifests, which may require a different PR.

gkellogg:
Replaced by #539.

PR #559, opened 2023-03-01
Add JSON literal tests.

davidlehn:
- Various tests of when `@type` of `@json` is used, the JSON may be an array, and the interaction with a `@container` of `@set` and `@list`.
- Change compaction output for `@json` and `@set` with a JSON array.
- Add test for many other possibilities and processing algorithms.
- Propose new `multiple JSON literals` error code when trying to add multiple JSON literals as values for a property not declared a `@set`.

**NOTE**: This is a work-in-progress that needs feedback.  HTML rebuild has not yet been done.

Please put major discussion in the issue that links here as this PR may need to be replaced.


gkellogg:
Some of these assume changes to the algorithms that will need their own PRs:

* Tests expand/er57 and er58 along with toRdf/er57 and er58 need a check to be sure that the value is an array, or generate the `invalid set or list object` error. The presumed semantics are that each array value becomes its own JSON value.
* Test compact/ej01 is expected to raise `multiple JSON literals` because the term didn't use `@set`. This is arguably a breaking change as any data that did this before would now fail (of course two JSON values on input would now be one JSON value on output). This will certainly require checks in the compaction algorithm.
* Test compact/js07 (and others) also change the behavior to preserve the array value.

gkellogg:
Addresses #560.

tests/compact-manifest.html
```diff
@@ -4805,6 +4805,38 @@ <h2>
             </dd>
           </dl>
         </dd>
+        <dt id='tej01'>
+          Test tej01 Multiple JSON values for a JSON literal properties
```
TallTed, line 4809:
```suggestion
          Test tej01 Multiple JSON values for a JSON literal property
```

TallTed (changes requested):
editorial, human-facing typo

tests/expand-manifest.jsonld
```diff
@@ -1826,6 +1826,27 @@
       "purpose": "Verifies that an exception is raised when attempting to redefine @context.",
       "input": "expand/er56-in.jsonld",
       "expectErrorCode": "keyword redefinition"
+    }, {
+      "@id": "#ter57",
+      "@type": ["jld:NegativeEvaluationTest", "jld:ExpandTest"],
+      "name": "Array not found for @json @set.",
+      "purpose": "The value for a @json @set term MUST be an array.",
+      "input": "expand/er57-in.jsonld",
+      "expectErrorCode": "invalid set or list object"
```
pchampin, line 1835: I this test is accepted, it should have counterparts with other values for `@type` (`ex:datatype`, `@id`), and without any `@type`. I see no reason why `"@type": "@json"` should behave differently.

But more generally, I don't think that this test should be accepted nor that this behavior should be required. There are many other situations where parts of the context are ignored if the encountered value is not of the expected kind. For example, `@type` coertion is ignored if the value of the property is not a JSON scalar.

tests/expand-manifest.jsonld
```diff
@@ -1826,6 +1826,27 @@
...
+    }, {
+      "@id": "#ter58",
+      "@type": ["jld:NegativeEvaluationTest", "jld:ExpandTest"],
+      "name": "Array not found for @json @list.",
+      "purpose": "The value for a @json @list term MUST be an array.",
+      "input": "expand/er58-in.jsonld",
+      "expectErrorCode": "invalid set or list object"
```
pchampin, line 1849: Same comment as above (for `ter57`): I don't think that this test should be accepted.

I concur that having a simple value converted to an RDF list is a bit counter-intuitive, but it follows the principle that, in JSON-LD, `"prop": val` and `"prop": [val]` are equivalent.

tests/fromRdf-manifest.jsonld
```diff
@@ -379,6 +379,22 @@
       "input": "fromRdf/js11-in.nq",
       "expect": "fromRdf/js11-out.jsonld",
       "option": {"specVersion": "json-ld-1.1"}
+    }, {
+      "@id": "#tjs12",
+      "@type": ["jld:PositiveEvaluationTest", "jld:FromRDFTest"],
+      "name": "JSON literals (strings)",
+      "purpose": "Tests creating property with multiple rdf:type rdf:JSON to a JSON literal (null).",
```
pchampin, line 386: the 'name' is not accurate, and I don't understand the sentence under 'purpose' (using "`rdf:type`" to talk about the datatype is very confusing).
```suggestion
      "name": "Multiple JSON literals",
      "purpose": "Tests creating property with multiple literal values of datatype rdf:JSON",
```

tests/fromRdf-manifest.jsonld
```diff
@@ -379,6 +379,22 @@
...
+      "input": "fromRdf/js12-in.nq",
+      "expect": "fromRdf/js12-out.jsonld",
+      "option": {"specVersion": "json-ld-1.1"}
+    }, {
+      "@id": "#tjs13",
+      "@type": ["jld:PositiveEvaluationTest", "jld:FromRDFTest"],
+      "name": "Mixed JSON literals and non-JSON",
+      "purpose": "Tests creating property with mixed rdf:type rdf:JSON and non-JSON.",
```
pchampin, line 394:
```suggestion
      "name": "Mixed JSON literal and non-JSON",
      "purpose": "Tests creating property with mixed values: literal of datatype rdf:type and non-JSON.",
```

Actually, the test is mixing a literal of datatype `rdf:JSON` and an IRI, so the text is still a bit misleading (I would expect "non-JSON" to mean "a literal with another datatype than `rdf:JSON`).

tests/toRdf-manifest.jsonld
```diff
@@ -2251,6 +2251,20 @@
...
+      "purpose": "The value for a @json @list term MUST be an array.",
+      "input": "toRdf/er57-in.jsonld",
+      "expectErrorCode": "invalid set or list object"
+    }, {
+      "@id": "#ter58",
+      "@type": ["jld:NegativeEvaluationTest", "jld:ToRDFTest" ],
+      "name": "Array not found for @json @set.",
+      "purpose": "The value for a @json @set term MUST be an array.",
+      "input": "toRdf/er58-in.jsonld",
+      "expectErrorCode": "invalid set or list object"
```
pchampin, line 2267:
```suggestion
```
For the reasons explained above (in expand-manifest.jsonld), I believe that these tests should not be accepted.

pchampin (changes requested):
I disagree with tests er57 and er58 (in expand and toRdf) -- see my comments inline.

The rests of the tests make sense, but I agree with @gkellogg that some of them are breaking changes.

tests/compact-manifest.html
TallTed: This does not appear to have been applied. It should not be set as resolved.

PR #566, opened 2023-06-07
Fix context processing for reverse terms

### LINKED ISSUE #565: The algorithm for processing a reverse term returns too early
In section 4.2 [Create Term Definition](https://www.w3.org/TR/json-ld11-api/#create-term-definition) of the Context Processing Algorithms, step 13, processing a reverse term ends with:

> Set the term definition of term in active context to definition and the value associated with defined's entry term to true and return.

But with that early return, the algorithm stops processing a reverse term definition here, and thus reverse terms will not use any compaction features captured in the subsequent steps (step 19 and onward); e.g. indexed containers).

Thus this example:
```json
{
  "@context": {
    "@base": "https://example.org/",
    "@vocab": "https://example.net/ns#",
    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
    "statement": {"@reverse": "rdf:subject", "@container": "@index", "@index": "predicate"},
    "predicate": {"@id": "rdf:predicate", "@type": "@vocab"},
    "term": {"@id": "rdf:object", "@type": "@vocab"},
    "addedIn": {"@type": "@id"}
  },
  "@id": "/item/1",
  "statement": {
    "rdf:type": {"term": "A", "addedIn": "v1"}
  }
}
```
which *should* expand to:
```json
[
  {
    "@id": "https://example.org/item/1",
    "@reverse": {
      "http://www.w3.org/1999/02/22-rdf-syntax-ns#subject": [
        {
          "https://example.net/ns#addedIn": [
            {
              "@id": "https://example.org/v1"
            }
          ],
          "http://www.w3.org/1999/02/22-rdf-syntax-ns#object": [
            {
              "@id": "https://example.net/ns#A"
            }
          ],
          "http://www.w3.org/1999/02/22-rdf-syntax-ns#predicate": [
            {
              "@id": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"
            }
          ]
        }
      ]
    }
  }
]
```
instead results in:
```json
[
  {
    "@id": "https://example.org/item/1",
    "@reverse": {
      "http://www.w3.org/1999/02/22-rdf-syntax-ns#subject": [
        {
          "https://example.net/ns#addedIn": [
            {
              "@id": "https://example.org/v1"
            }
          ],
          "http://www.w3.org/1999/02/22-rdf-syntax-ns#object": [
            {
              "@id": "https://example.net/ns#A"
            }
          ],
          "@index": "rdf:type"
        }
      ]
    }
  }
]
```

since the term definition of `statement` is missing what term to use when interpreting the `@index` value.

The fix is simple: remove "and return" from the end of step 13, and revise step 14 to begin with "Otherwise".

This is consistent with playground and distiller behaviour, and by looking at implementation, e.g. [the ruby implementation does not return early](https://github.com/ruby-rdf/json-ld/blob/3.2.3/lib/json/ld/context.rb#L605).

I have a fix for [TRLD](https://github.com/niklasl/trld/pull/3) to reflect the above (applying that, example output changed accordingly to the expected behaviour).

niklasl:
This changes the algorithm for processing a reverse term so that it doesn't return early. Fixes #565.

Includes one test for expansion using a reverse term with `@container: @index` and `@index` set to a property, and a roundtripping compaction test based on the former.

(Note: the round-tripping compaction test _reuses_ the expand test data. That does not seem to be the practise; is there any rationale against that?)

### DIFF
--- a/tests/compact-manifest.jsonld
+++ b/tests/compact-manifest.jsonld
@@ -958,6 +958,15 @@
       "context": "compact/0113-context.jsonld",
       "expect": "compact/0113-out.jsonld"
     }, {
+      "@id": "#t0114",
+      "@type": ["jld:PositiveEvaluationTest", "jld:CompactTest"],
+      "name": "Reverse term with property based indexed container",
+      "purpose": "With reverse term using @container: @index and @index set to a property, ensure round-tripping from expaned form",
+      "input": "expand/0131-out.jsonld",
+      "context": "expand/0131-in.jsonld",
+      "expect": "expand/0131-in.jsonld",
+      "option": {"base": "https://example.org/", "specVersion": "json-ld-1.1"}
+    }, {
       "@id": "#tc001",
       "@type": ["jld:PositiveEvaluationTest", "jld:CompactTest"],
       "name": "adding new term",
--- a/index.html
+++ b/index.html
@@ -1664,10 +1664,10 @@
             <li>Set the <a>term definition</a> of <var>term</var> in
               <var>active context</var> to <var>definition</var> and the
               value associated with <var>defined</var>'s <a>entry</a> <var>term</var> to
-              <code>true</code> and return.</li>
+              <code>true</code>.</li>
           </ol>
         </li>
-        <li>If <var>value</var> contains the <a>entry</a> <code>@id</code> and its value
+        <li>Otherwise if <var>value</var> contains the <a>entry</a> <code>@id</code> and its value
           does not equal <var>term</var>:
           <ol>
             <li id="ctd-id-null">If the <code>@id</code> <a>entry</a> of <var>value</var>
--- a/tests/compact-manifest.html
+++ b/tests/compact-manifest.html
@@ -3212,6 +3212,40 @@ Test t0113 Compact property index using Absolute IRI index
 </dd>
 </dl>
 </dd>
+<dt id='t0114'>
+Test t0114 Reverse term with property based indexed container
+</dt>
+<dd>
+<dl class='entry'>
+<dt>id</dt>
+<dd>#t0114</dd>
+<dt>Type</dt>
+<dd>jld:PositiveEvaluationTest, jld:CompactTest</dd>
+<dt>Purpose</dt>
+<dd>With reverse term using @container: @index and @index set to a property, ensure round-tripping from expaned form</dd>
+<dt>input</dt>
+<dd>
+<a href='expand/0131-out.jsonld'>expand/0131-out.jsonld</a>
+</dd>
+<dt>context</dt>
+<dd>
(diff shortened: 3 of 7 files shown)


gkellogg (approved):

tests/compact-manifest.jsonld
```diff
@@ -957,6 +957,15 @@
       "input": "compact/0113-in.jsonld",
       "context": "compact/0113-context.jsonld",
       "expect": "compact/0113-out.jsonld"
+    }, {
+      "@id": "#t0114",
+      "@type": ["jld:PositiveEvaluationTest", "jld:CompactTest"],
+      "name": "Reverse term with property based indexed container",
+      "purpose": "With reverse term using @container: @index and @index set to a property, ensure round-tripping from expaned form",
+      "input": "expand/0131-out.jsonld",
```
davidlehn, line 965: I'd suggest changing this to be consistent with the independent files per-test and naming pattern used for all other tests.  Copy the files from `expand` to `compact` and rename them to match the param in/context/out pattern.  I see it's doing a round-tripping thing, but the change from the normal pattern could be confusing to others and our future selves.

davidlehn: > (Note: the round-tripping compaction test reuses the expand test data. That does not seem to be the practise; is there any rationale against that?)

Missed that comment... one reason is to be consistent.  Reusing cross algorithms, mixing up in/out/context naming... it will work but it's not straightforward.  I think it's better to just be clear what the files are and for which test and which params.

davidlehn:
Does this need parallel toRdf tests too?  t0131 is not going to line up between expand/toRdf since there already exists some toRdf tests over 0130.  Not sure if some other prefix would be better here for that reason alone.

gkellogg:
Yes to @davidlehn’s points. Also, there should be an entry in the “changes since Recommendation …” section.

tests/compact-manifest.jsonld
niklasl: Agreed! Addressed in a6a1dcb8af02b7a589e85be5a6c391c71f18412d.

niklasl:
> Yes to @davidlehn’s points. Also, there should be an entry in the “changes since Recommendation …” section.

Added an entry about this change in 8bfdf285ef6836cb80534ac2f8da95c9cc644bf6.

niklasl:
Note that I added _textual_ references in the `purpose` text to the original expand test id. We should consider adding proper interlinking of all tests instead, so we can query on what is based on what, and if there are any missing round-trip or to/from RDF tests. That would be a separate task of course.

dlongley (approved):
This seems ok to me and like an oversight previously.

TallTed (changes requested):
Trivial, but important.

pchampin (approved):

PR #570, opened 2023-06-22
Add similar id protected term redefinition test.

davidlehn:
- Add test of redefining equivalent protected id terms in different forms:
  - `"foo": {"@id": "ex:foo"}`
  - `"foo": "ex:foo"`

This currently fails in jsonld.js and with the distiller.  Should implementations be considering these forms equivalent as far as protected redefinition checks are concerned?

### DIFF
--- a/tests/expand-manifest.jsonld
+++ b/tests/expand-manifest.jsonld
@@ -2828,6 +2828,14 @@
       "input": "expand/pr40-in.jsonld",
       "expect": "expand/pr40-out.jsonld"
     }, {
+      "@id": "#tpr41",
+      "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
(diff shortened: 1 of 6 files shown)


gkellogg (commented):
I suppose this depends on what it means to be the "same as the previous", which is what is in step 27.1. Currently, my implementation does not pass this.

dlongley:
@gkellogg,

> I suppose this depends on what it means to be the "same as the previous", which is what is in step 27.1.

Yes, though surely the "same as the previous" was not intended to trigger errors on meaningless differences due to syntactic sugar. It seems we have an oversight here where we need to clarify that processors must de-sugar things to properly compare for "sameness".

gkellogg (approved):

dlongley (approved):

pchampin (approved):

PR #582, opened 2023-12-06
Add protected graph container null context test.

davidlehn:
Adds a test for a protected `@graph` `@container` property that nullifies it's `@context`.  This is similar to a pattern found in the VC Data Model 2.0 context.  There are other test variations of context nullification, graph properties, and protection rules, etc, but there doesn't seem to be one with this combination.  jsonld.js was failing this test but other implementations pass.

Are other variations of this test needed?  It looks like most base features are covered, but perhaps tricky combinations of features might expose edge case issues.  Help needed.

See:
https://github.com/w3c/vc-data-model/issues/1373

### DIFF
--- a/tests/expand-manifest.html
+++ b/tests/expand-manifest.html
@@ -9643,6 +9643,34 @@
             </dd>
           </dl>
         </dd>
+        <dt id='tpr43'>
+          Test tpr43 Clear protection in @graph @container with null context.
+        </dt>
+        <dd>
+          <dl class='entry'>
+            <dt>id</dt>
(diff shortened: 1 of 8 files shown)


gkellogg (approved):

davidlehn:
This had to be rebased due to conflicting "pr42" tests.  Ready to merge with others are ready.

@pchampin and other implementors, can you please review this.

@gkellogg The generate process output `expand-manifest.html` but not `toRdf-manifest.html`.  Is that a script error or an issue with the PR?

pchampin (approved):

PR #585, opened 2023-12-13
Add graph container array tests.

davidlehn:
- Add tests for `@graph` `@container` with and without nullified context with an array of plain literals.
- Add tests for `@graph` `@id` `@container` with and without nullified context with an array of IRIs.

The expand/toRdf parallel testing and number is a bit confusing.  Not all the tests have a twin.  I jumped up to 0200 thinking maybe going forward from there new tests might align better.

As of now, different implementations have different results for these sorts of tests.  For instance, for 0203:
- jsonld.js - n-quads drops the data, triggers a "safe mode" error for canonizing due to the drops.  ([link](https://json-ld.org/playground/#startTab=tab-canonized&json-ld=%7B%22%40context%22%3A%7B%22p%22%3A%7B%22%40id%22%3A%22ex%3Ap%22%2C%22%40type%22%3A%22%40id%22%2C%22%40container%22%3A%22%40graph%22%2C%22%40context%22%3Anull%7D%7D%2C%22%40id%22%3A%22ex%3Aouter%22%2C%22p%22%3A%5B%22ex%3Atest1%22%2C%22ex%3Atest2%22%5D%7D))
- RDF Distiller - errors with "undefined method `[]' for nil:NilClass" ([link](http://rdf.greggkellogg.net/distiller?command=serialize&format=jsonld&output_format=nquads))
- Sophia - n-quads has data with blank node ids - ([link](https://perso.liris.cnrs.fr/pierre-antoine.champin/2023/sowasm/?guess=&oformat=application%2Fn-quads&auto=&input=%7B%0A++%22%40context%22%3A+%7B%0A++++%22p%22%3A+%7B%0A++++++%22%40id%22%3A+%22ex%3Ap%22%2C%0A++++++%22%40type%22%3A+%22%40id%22%2C%0A++++++%22%40container%22%3A+%22%40graph%22%2C%0A++++++%22%40context%22%3A+null%0A++++%7D%0A++%7D%2C%0A++%22%40id%22%3A+%22ex%3Aouter%22%2C%0A++%22p%22%3A+%5B%0A++++%22ex%3Atest1%22%2C%0A++++%22ex%3Atest2%22%0A++%5D%0A%7D))

People have been looking at this construct in places like the VC work.  Providing URLs in the `verifiableCredential` property of a VP (I think).  This additional coverage is needed to understand what should happen here unless this is specifically undefined behavior.  (I'm not sure if some tests duplicate other tests, added variations for clarity.)

I'm not sure what the output should be of these tests.  Currently it has the jsonld.js empty output, but that may not be correct in all cases?

What other variations of these tests should be added?

Also related is what to do with a `"@container": ["@graph", "@id"]` with only a IRI provided instead of a mapping.  I'm not sure that has test either.  It outputs `<ex:outer> <ex:p> <ex:test> .`/`... "ex:test"` triples in jsonld.js and others, which seems closer to the behavior people might expect without the `@id`, but the feature was designed for only mappings instead of values like this?


gkellogg:
The implication in the Expand Algorithm step 13.12.1 is that _if_ must either be a _node object_ or a _graph object_, and not a _value object_. I agree that if it is a _value object_, should be omitted. However, I do get an expanded output after accounting for this:

```
[
  {
    "@id": "ex:outer",
    "ex:p": []
  }
]
```

I don't know of a step in the expansion algorithm which would eliminate this entry.

It did require a change in 13.12.1 to eliminate expanded values which are not either node objects or graph objects if _container_ includes `@graph`. We may need an Erratum and issue to add wording to this effect.

gkellogg (changes requested):
As I noted, I think the expansion results should be different, although it does not change the RDF results.

pchampin (commented):
Sorry for the late review. 

If we are to make a significant change to the spec for handling this corner case (which, IIUC, would be required to justify these test results), I think I would rather lean towards raising an error on these input.

gkellogg:
> If we are to make a significant change to the spec for handling this corner case (which, IIUC, would be required to justify these test results), I think I would rather lean towards raising an error on these input.

The way we've handled other such issues is to create an erratum and note it as a substantive change. Our charter allows us to address errata, which IMO would include this. Tests are appropriate once the erratum is accepted and changes have been added to the draft.

PR #606, opened 2024-07-10
The default for extractAllScripts in toRdf() should be true.

gkellogg:
Tests are already consistent with this interpretation, although it was never documented in the API properly.

Fixes #603.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/606.html" title="Last updated on Jul 12, 2024, 2:21 PM UTC (b2ece69)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/606/30aa531...b2ece69.html" title="Last updated on Jul 12, 2024, 2:21 PM UTC (b2ece69)">Diff</a>

Issue #603, opened 2024-06-17
toRdf() and flatten() API require extractAllScripts to default to true


gkellogg:
The JSON-LD section on [JSON-LD in HTML Documents](https://w3c.github.io/json-ld-syntax/#embedding-json-ld-in-html-documents) specifically says that all script all script elements are extracted, implying that the default value of |extractAllScripts| should be `true` for the `toRdf()` API entry point.

Tests for `flatten()` also say that it should extract all script elements, although there is no normative text to this effect.

Changing the behavior of `flatten()` would represent a change to the 1.1 behavior, but may be warranted.

The `toRdf()` entry point should be updated to say that the default is `true`.

TallTed:
> The `toRdf()` entry point should be updated to say that the default if `true`.

Typo... ```default if `true` ``` -> ```default is `true` ```

gkellogg:
<details>
<summary>Discussed on [today's call](https://json-ld.org/minutes/2024-07-10/#28)</summary>
<a id="link-topic-5.2" style="display:none;" href="#topic-5.2">✪</a></h2>
<div onmouseout="$('#link-28').hide()" onmouseover="$('#link-28').show()" id="28" class="information"><a href="https://github.com/w3c/json-ld-api/issues/603">https://github.com/w3c/json-ld-api/issues/603</a> -&gt; Issue 603 toRdf() and flatten() API require extractAllScripts to default to true (by gkellogg) [spec:enhancement] [needs discussion] [ErratumRaised] <a id="link-28" style="display:none;" href="#28">✪</a></div>
<div onmouseout="$('#link-29').hide()" onmouseover="$('#link-29').show()" id="29" class="comment"><span class="name">Anatoly Scherbakov</span>: Dlehn I wouldn&apos;t like to distract everyone as this is probably off-topic, but the issue I am facing now with python-yaml-ld is that I am redoing a lot of stuff that pyld already is doing. Basically, pyld already implements everything that&apos;s needed except a special DocumentLoader. So ideally, it feels like there shouldn&apos;t be a separate library for <a id="link-29" style="display:none;" href="#29">✪</a></div>
[...]

PR #607, opened 2024-07-10
Fix test output for html/f004 and add html/f005

gkellogg:
Add html/f005 with extractAllScripts set to true.

See #606.

For #603.

cc/@niklasl

### DIFF
--- a/tests/html-manifest.jsonld
+++ b/tests/html-manifest.jsonld
@@ -275,12 +275,25 @@
   }, {
     "@id": "#tf004",
     "@type": ["jld:PositiveEvaluationTest", "jld:FlattenTest", "jld:HtmlTest"],
-    "name": "Flattens all script elements by default",
+    "name": "Flattens first script element by default",
     "purpose": "Tests embedded JSON-LD in HTML multiple script elements",
     "input": "html/f004-in.html",
     "context": "html/f004-context.jsonld",
     "expect": "html/f004-out.jsonld",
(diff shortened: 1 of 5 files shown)


tests/html-manifest.jsonld
```diff
@@ -275,12 +275,25 @@
   }, {
     "@id": "#tf004",
     "@type": ["jld:PositiveEvaluationTest", "jld:FlattenTest", "jld:HtmlTest"],
-    "name": "Flattens all script elements by default",
+    "name": "Flattens first script element by default",
     "purpose": "Tests embedded JSON-LD in HTML multiple script elements",
     "input": "html/f004-in.html",
     "context": "html/f004-context.jsonld",
     "expect": "html/f004-out.jsonld",
     "option": {"specVersion": "json-ld-1.1", "processorFeature": "HTML Script Extraction"}
```
niklasl, line 283: For robustness, I would suggest to also spell out `"extractAllScripts": false` here. But perhaps that would break consistency in the test suite (by not always relying on the implicit defaults)?

gkellogg: Yes, explicitly saying with `extractAllScripts` `false` implies that the option test element contain that property; otherwise, it falls back to the default as specified in the API, which is that `extractAllScripts` is `false`.

anatoly-scherbakov (approved):

niklasl (approved):
LGTM

pchampin (approved):

PR #617, opened 2024-10-25
Update ci.yml for Echidna rules

gkellogg:


gkellogg:
@pchampin I based this on the auto-publish action for rdf-concepts. I commented out the group decision, although we may have one already that we can use.

Issue #555, opened 2022-12-13
blind spot in RDF to object conversion


pchampin:
I believe that the [RDF to object conversion](https://w3c.github.io/json-ld-api/#algorithm-16), step 2.3.4, has a blind spot.
This is in the case `useNativeTypes` is used:

> if the [datatype IRI](https://www.w3.org/TR/rdf11-concepts/#dfn-datatype-iri) of value equals xsd:integer or xsd:double and its [lexical form](https://w3c.github.io/json-ld-api/#dfn-lexical-form) is a valid xsd:integer or xsd:double according [[XMLSCHEMA11-2](https://w3c.github.io/json-ld-api/#bib-xmlschema11-2)], set converted value to the result of converting the [lexical form](https://w3c.github.io/json-ld-api/#dfn-lexical-form) to a JSON [number](https://tc39.es/ecma262/multipage/#sec-terms-and-definitions-number-value).

The problem is: some valid literals of these datatypes may fail to convert to JSON numbers, e.g. `"+INF"^^xsd:double`, or very large integers.

The spec is currently silent about that. We should either specify that an error must be raised, or that `useNativeType` should be ignored in that case, and a regular value object should be generated. I personally prefer the 2nd option.

gkellogg:
We should probably be using logic from JCS for dividing these different forms.

gkellogg:
JSON Canonicalization, which we use when serializing JSON literals

pchampin:
> JSON Canonicalization, which we use when serializing JSON literals

ah... thanks

[JCS Appendix D. ](https://www.rfc-editor.org/rfc/rfc8785#section-appendix.d)[Dealing with Big Numbers](https://www.rfc-editor.org/rfc/rfc8785#name-dealing-with-big-numbers) says
> numbers that do not have a natural place in the current JSON ecosystem MUST be wrapped using the JSON string type.

This is aligned with my suggestion above: whenever a literal can not be mapped to a JSON number, ignore `useNativeType` and keep it as a value object with explicit `@value` (as a string) and `@type`.

anatoly-scherbakov:
I tried to cover this with a PR: https://github.com/w3c/json-ld-api/pull/619

PR #619, opened 2024-11-08
RDF to Object conversion (closing #555)

anatoly-scherbakov:
# Why

Trying to solve #555.

# What

Clarify what we do when can't serialize a number from RDF into JSON.

***
<a href="https://pr-preview.s3.amazonaws.com/iolanta-tech/json-ld-api/pull/619.html" title="Last updated on Nov 29, 2024, 7:39 PM UTC (a6132a4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/619/a24a4b3...iolanta-tech:a6132a4.html" title="Last updated on Nov 29, 2024, 7:39 PM UTC (a6132a4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5600,8 +5600,8 @@
                   to a JSON <a>number</a>.
                 </li>
                 <li>
-                  Otherwise, if <a>lexical form</a> expresses a numerical value
-                  outside of the range mandated by [[IEEE-754-2008]], then set
+                  Otherwise, if <a>lexical form</a> expresses a numeric value
+                  outside of the ranges supported by [[IEEE-754-2008]], then set
                   <var>converted value</var> to a JSON string literal containing
                   said lexical form.
                 </li>
--- a/index.html
+++ b/index.html
@@ -5201,7 +5201,7 @@
 
         If the <a data-link-for="JsonLdOptions">useNativeTypes</a> flag is set
         to <code>true</code>, <a>RDF literals</a> are converted to JSON values
-        based on their <a>datatype IRI</a> and <a>lexical form</a> as described
+        based on their <a>datatype IRI</a> and <a>lexical form</a>, as described
         in <a class="sectionRef" href="#algorithm"></a>.
       </p>
 
--- a/index.html
+++ b/index.html
@@ -5166,6 +5166,14 @@
           <li>
             <code>xsd:integer</code> or <code>xsd:double</code> are converted
         to <a>JSON numbers</a>,
+            <ul>
+              <li>
+                unless such a conversion is impossible, for instance, in cases like
+                <code>"+INF"^^xsd:double</code>, in which instance <a data-link-for="JsonLdOptions">useNativeTypes</a> flag is ignored,
+                and the literal value is kept in string form,
+                in accordance with <a data-cite="RFC8785#section-appendix.d">RFC 8785, Appendix D. Dealing with Big Numbers</a>.
+              </li>
+            </ul>
           </li>
           <li>and <a>RDF literals</a>
         with a <a>datatype IRI</a>
--- a/index.html
+++ b/index.html
@@ -5205,7 +5205,8 @@
         in <a class="sectionRef" href="#algorithm"></a>.
       </p>
 
-      <table class="simple">
+      <table class="simple" id="literal-conversation-with-useNativeTypes>
+        <caption>Literal conversion when <a data-link-for="JsonLdOptions">useNativeTypes</a> is set to <code>true</code></caption>
         <thead>
           <tr>
             <th>RDF datatype</th>
--- a/index.html
+++ b/index.html
@@ -5234,7 +5234,7 @@
 
           <tr>
             <td>Any other value</td>
-            <td class="error">Raise an error</td>
+            <td class="error">Throw <a data-link-for="JsonLdErrorCode">invalid JSON literal</a></td>
           </tr>
 
           <tr>
@@ -5258,7 +5258,7 @@
                 <li><code>"-INF"^^xsd:double</code></li>
               </ul>
             </td>
-            <td class="error" rowspan="2">Raise an error</td>
+            <td class="error" rowspan="2">Throw <a data-link-for="JsonLdErrorCode">invalid JSON literal</a></td>
           </tr>
           <tr>
             <td>Any other value</td>
@@ -5275,7 +5275,7 @@
           </tr>
           <tr>
             <td>Any other value</td>
-            <td class="error">Raise an error</td>
+            <td class="error">Throw <a data-link-for="JsonLdErrorCode">invalid JSON literal</a></td>
           </tr>
         </tbody>
       </table>
--- a/index.html
+++ b/index.html
@@ -5207,9 +5207,11 @@
 
       <table class="simple">
         <thead>
-          <th>RDF datatype</th>
-          <th>RDF Literal</th>
-          <th>JSON</th>
+          <tr>
+            <th>RDF datatype</th>
+            <th>RDF Literal</th>
+            <th>JSON</th>
+          </tr>
         </thead>
         <tbody>
           <tr>
--- a/index.html
+++ b/index.html
@@ -284,6 +284,48 @@
     text-decoration: none;
     color: #666;
   }
+
+  table.simple {
+      border-collapse: collapse;
+      width: 100%;
+      font-size: 14px;
+  }
+
+  table.simple th,
+  table.simple td {
+      border: 1px solid #ddd;
+      padding: 8px;
+      vertical-align: top; /* Align text to the top */
+  }
+
+  table.simple th {
+      background-color: #f2f2f2; /* Light gray for headers */
+      font-weight: bold;
+      text-align: left;
+  }
+
+  table.simple tr:nth-child(even) {
+      background-color: #f9f9f9; /* Alternating row colors */
+  }
+
+  table.simple tr:hover {
+      background-color: #e9e9e9; /* Highlight on hover */
+  }
+
+  table.simple a {
+      color: #1a73e8; /* Make links a visible color */
+      text-decoration: none;
+  }
+
+  table.simple a:hover {
+      text-decoration: underline;
+  }
+
+  table.simple td.error {
+    background-color: #FFA500;
+    color: #000;
+    font-weight: bold;
+  }
 </style>
 </head>
 
@@ -5166,7 +5208,7 @@
       <table class="simple">
         <thead>
           <th>RDF datatype</th>
-          <th>Values</th>
+          <th>RDF Literal</th>
           <th>JSON</th>
         </thead>
         <tbody>
@@ -5192,11 +5234,11 @@
 
           <tr>
             <td>Any other value</td>
-            <td>Raise an error</td>
+            <td class="error">Raise an error</td>
           </tr>
 
           <tr>
-            <th rowspan="3"><code>xsd:double</code></th>
+            <th rowspan="4"><code>xsd:double</code></th>
             <td>Real number in the <code>[-2<sup>53</sup> + 1, 2<sup>53</sup> + 1]</code> range, representable as per [[IEEE-754-2008]]</td>
             <td><a>JSON Number</a></td>
           </tr>
@@ -5216,7 +5258,7 @@
                 <li><code>"-INF"^^xsd:double</code></li>
               </ul>
             </td>
-            <td rowspan="2">Raise an error</td>
+            <td class="error" rowspan="2">Raise an error</td>
           </tr>
           <tr>
             <td>Any other value</td>
@@ -5233,7 +5275,7 @@
           </tr>
           <tr>
             <td>Any other value</td>
-            <td>Raise an error</td>
+            <td class="error">Raise an error</td>
           </tr>
         </tbody>
       </table>
--- a/index.html
+++ b/index.html
@@ -5202,7 +5202,7 @@
         If the <a data-link-for="JsonLdOptions">useNativeTypes</a> flag is set
         to <code>true</code>, <a>RDF literals</a> are converted to JSON values
         based on their <a>datatype IRI</a> and <a>lexical form</a>, as described
-        in <a class="sectionRef" href="#algorithm"></a>.
+        in <a class="sectionRef" href="#algorithm-16"></a>.
       </p>
 
       <table class="simple" id="literal-conversation-with-useNativeTypes>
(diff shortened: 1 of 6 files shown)


anatoly-scherbakov:
I don't think I can invite reviewers to this PR :thinking: @pchampin @gkellogg

gkellogg:
> I don't think I can invite reviewers to this PR 🤔 @pchampin @gkellogg

I'll look at your permissions, you should have full permissions on this and other repos.

index.html
```diff
@@ -5156,17 +5156,38 @@ <h3>Overview</h3>
...
         and <a>blank node identifiers</a>.
+
         If the <a data-link-for="JsonLdOptions">useNativeTypes</a> flag is set to <code>true</code>,
         <a>RDF literals</a> with a
         <a>datatype IRI</a>
-        that equals <code>xsd:integer</code> or <code>xsd:double</code> are converted
-        to a <a>JSON numbers</a> and <a>RDF literals</a>
+        that equals
+
+        <ul>
```
gkellogg, line 5165: The paragraph will need to be broken up, as a `<p>` can not include a `<ul>`.

index.html
```diff
@@ -5156,17 +5156,38 @@ <h3>Overview</h3>
...
+        to <a>JSON numbers</a>,
+            <ul>
+              <li>
+                unless such a conversion is impossible, for instance, in cases like
+                <code>"+INF"^^xsd:double</code>, in which instance <a data-link-for="JsonLdOptions">useNativeTypes</a> flag is ignored,
+                and the literal value is kept in string form,
+                in accordance with <a data-cite="RFC8785#section-appendix.d">RFC 8785, Appendix D. Dealing with Big Numbers</a>.
+              </li>
+            </ul>
+          </li>
```
gkellogg, line 5177: This should also account for the other cases called out in JCS about numbers that are too large, or carry too much precision, to be represented by either xsd:integer or xsd:double. For example, `1.4e+9999`, and `9223372036854775807`.

gkellogg (changes requested):
The changes are to the informative summary, but the issue actually calls out 2.4.3 (was 2.3.4) which is normative; that's the important area to address.

gkellogg:
@anatoly-scherbakov, if you can create branches at json-ld-api rather than on your own fork, it is easier for others to collaborate.

gkellogg:
As an alternative, we may treat these as error conditions, which is what JCS does.

> [3.2.2.3. ](https://www.rfc-editor.org/rfc/rfc8785#section-3.2.2.3)[Serialization of Numbers](https://www.rfc-editor.org/rfc/rfc8785#name-serialization-of-numbers)
> ...
> Note: Since Not a Number (NaN) and Infinity are not permitted in JSON, occurrences of NaN or Infinity MUST cause a compliant JCS implementation to terminate with an appropriate error.

davidlehn:
Can you please replace the emoji in the commit titles with words.

index.html
```diff
@@ -5156,17 +5156,38 @@ <h3>Overview</h3>
         and generating a JSON-LD document in expanded form for all
         <a>RDF literals</a>, <a>IRIs</a>
         and <a>blank node identifiers</a>.
+
         If the <a data-link-for="JsonLdOptions">useNativeTypes</a> flag is set to <code>true</code>,
         <a>RDF literals</a> with a
         <a>datatype IRI</a>
-        that equals <code>xsd:integer</code> or <code>xsd:double</code> are converted
-        to a <a>JSON numbers</a> and <a>RDF literals</a>
+        that equals
```
davidlehn, line 5163: This list markup inside a sentence style is a bit difficult to read.  I'd suggest rewording the changes and using a similar style to what is elsewhere in the spec.

index.html
```diff
@@ -5156,17 +5156,89 @@ <h3>Overview</h3>
...
-        as described in
-        <a class="sectionRef" href="#data-round-tripping"></a>.
+
+        If the <a data-link-for="JsonLdOptions">useNativeTypes</a> flag is set
+        to <code>true</code>, <a>RDF literals</a> are converted to JSON values
+        based on their <a>datatype IRI</a> and <a>lexical form</a> as described
+        in <a class="sectionRef" href="#data-round-tripping"></a>.
+      </p>
+
+      <table class="simple">
```
gkellogg, line 5166: The table formatting needs some work; probably vertical alignment and borders to make it understandable.

gkellogg (changes requested):

index.html
anatoly-scherbakov: Yes, this is a rather draft version; I'd like it to look like `<table class="simple">` in YAML-LD spec but I haven't looked yet into what CSS markup is needed for that.

anatoly-scherbakov: …ReSpec has changed table formatting: https://github.com/speced/respec/issues/4667

A pity, I liked the previous blue style.

anatoly-scherbakov: image

Updated.

gkellogg (changes requested):
Minor change to table head:  th elements need to be wrapped inside a tr.

index.html
anatoly-scherbakov: Yes, that's what I did to obtain the formatting on the screenshot above.

index.html
```diff
@@ -5156,17 +5198,91 @@ <h3>Overview</h3>
...
-        with a <a>datatype IRI</a>
-        that equals <code>xsd:boolean</code> are converted to <code>true</code> or
-        <code>false</code> based on their
-        <a>lexical form</a>
-        as described in
-        <a class="sectionRef" href="#data-round-tripping"></a>.
+
+        If the <a data-link-for="JsonLdOptions">useNativeTypes</a> flag is set
+        to <code>true</code>, <a>RDF literals</a> are converted to JSON values
+        based on their <a>datatype IRI</a> and <a>lexical form</a> as described
```
TallTed, line 5204:
```suggestion
        based on their <a>datatype IRI</a> and <a>lexical form</a>, as described
```

index.html
```diff
@@ -5477,10 +5593,24 @@ <h3>Algorithm</h3>
...
+                  according to [[XMLSCHEMA11-2]], and if it is a numeric value
+                  conformant to [[IEEE-754-2008]], then set <var>converted value</var>
                   to the result of converting the
                   <a>lexical form</a>
-                  to a JSON <a>number</a>.</li>
+                  to a JSON <a>number</a>.
+                </li>
+                <li>
+                  Otherwise, if <a>lexical form</a> expresses a numerical value
+                  outside of the range mandated by [[IEEE-754-2008]], then set
```
TallTed, line 5604: There are several ranges supported by [IEE-754-2008 (and IEE-754-2019, the latest rev, to which we might want to update our references)](https://en.wikipedia.org/wiki/IEEE_754), depending on the subtype of the numeric.
```suggestion
                  Otherwise, if <a>lexical form</a> expresses a numeric value
                  outside of the ranges supported by [[IEEE-754-2008]], then set
```

tests/fromRdf-manifest.jsonld
```diff
@@ -195,6 +195,26 @@
       "purpose": "Check list generation with rdf:first property and rdf:nil value.",
       "input": "fromRdf/0026-in.nq",
       "expect": "fromRdf/0026-out.jsonld"
+    }, {
+      "@id": "#t0027",
+      "@type": ["jld:PositiveEvaluationTest", "jld:FromRDFTest"],
+      "name": "native types flag set to true is ignored if the number provided is invalid in JSON",
+      "purpose": "Literals with datatype xsd:boolean, xsd:integer, and xsd:double are serialized using native scalar values if possible",
```
TallTed, line 202:
```suggestion
      "purpose": "Literals with datatype xsd:boolean, xsd:integer, and xsd:double are serialized using native scalar values, if possible",
```

gkellogg (approved):
Looks good, subject to @TallTed's suggestions. We may want to revisit what happens if `Infinity` or `NaN` is attempted to be serialized.

anatoly-scherbakov:
@TallTed created a separate issue about the version of the IEEE754 standard we refer to: https://github.com/w3c/json-ld-api/issues/620

anatoly-scherbakov:
@gkellogg do you mean we could convert them to JSON Strings instead of raising a JSON-LD processing error?

anatoly-scherbakov:
GitHub tells me that

> You're not authorized to push to this branch. Visit https://docs.github.com/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches for more information.

Not sure why though, it might be related to the fact that I opened it in my fork.

index.html
```diff
@@ -5156,17 +5198,91 @@ <h3>Overview</h3>
...
+
+          <tr>
+            <td>Whole number out of that range</td>
+            <td>JSON string literal</td>
+          </tr>
+
+          <tr>
+            <td>Any other value</td>
+            <td class="error">Throw <a data-link-for="JsonLdErrorCode">invalid JSON literal</a></td>
+          </tr>
```
pchampin, line 5240: I don't think that is a good idea, but I'll comment on that in the algorithm itself, below.

index.html
```diff
@@ -5156,17 +5198,91 @@ <h3>Overview</h3>
...
+              Non-numeric values:
+
+              <ul>
+                <li><code>"NaN"^^xsd:double</code>,</li>
+                <li><code>"+INF"^^xsd:double</code>,</li>
+                <li><code>"-INF"^^xsd:double</code></li>
+              </ul>
+            </td>
+            <td class="error" rowspan="2">Throw <a data-link-for="JsonLdErrorCode">invalid JSON literal</a></td>
+          </tr>
```
pchampin, line 5264: Same as above; I don't think that is a good idea, see my comments in the algorithm.

pchampin (changes requested):
See my comments in the "Files changed" view.

gkellogg (changes requested):
I question if this is a class-2 editorial change; I think it's class-3 change, which will require Candidate Correction marking. We can discuss.

pchampin:
> I question if this is a class-2 editorial change; I think it's class-3 change, which will require Candidate Correction marking. We can discuss.

+1, this falls under "clears up an ambiguity or under-specified part of the specification" which is part of "class 3".
+1 also to discuss

index.html
```diff
@@ -5156,17 +5198,89 @@ <h3>Overview</h3>
...
+
+          <tr>
+            <th rowspan="3"><code>xsd:integer</code></th>
+            <td>Whole number in the <code>[-2<sup>53</sup> + 1, 2<sup>53</sup> + 1]</code> range, representable as per [[IEEE-754-2008]]</td>
+            <td><a>JSON Number</a></td>
+          </tr>
+
+          <tr>
+            <td>Whole number out of that range</td>
+            <td>JSON string literal</td>
```
gkellogg, line 5232: Actually, this needs to be serialized as a value object with `@type`. The integer `1e54` (`1000000000000000078291540404596243842305360299886116864`) is outside of that range, but the value object needs to include `"@type": "xsd:integer".

index.html
```diff
@@ -5156,17 +5198,89 @@ <h3>Overview</h3>
...
+        <tbody>
+          <tr>
+            <th><code>xsd:string</code></th>
+            <td>An RDF string literal</td>
+            <td>JSON string literal</td>
+          </tr>
+
+          <tr>
+            <th rowspan="3"><code>xsd:integer</code></th>
+            <td>Whole number in the <code>[-2<sup>53</sup> + 1, 2<sup>53</sup> + 1]</code> range, representable as per [[IEEE-754-2008]]</td>
```
gkellogg, line 5226: There are more constraints in the [lexical form of an xsd:integer](https://www.w3.org/TR/xmlschema11-2/#integer-lexical-representation) than this.

Also, while it's likely consistent with JCS/ECMAScript, such numbers also need to be representable as integers in JCS/ECMAScript. See [Appendix B](https://www.rfc-editor.org/rfc/rfc8785#json.ieee754.test). 
We should be careful to not mis-describe this, and just refer to XSD Datatypes 1.1 and JCS.

index.html
```diff
@@ -5156,17 +5198,89 @@ <h3>Overview</h3>
...
+
+          <tr>
+            <th rowspan="4"><code>xsd:double</code></th>
+            <td>Real number in the <code>[-2<sup>53</sup> + 1, 2<sup>53</sup> + 1]</code> range, representable as per [[IEEE-754-2008]]</td>
+            <td><a>JSON Number</a></td>
+          </tr>
+
+          <tr>
+            <td>Real number out of that range</td>
+            <td>JSON string literal</td>
```
gkellogg, line 5248: See above comment; the ranges should be consistent with XSD and JCS and probably shouldn't re-define the criteria.

index.html
```diff
@@ -5477,10 +5593,24 @@ <h3>Algorithm</h3>
...
+                  outside of the ranges supported by [[IEEE-754-2008]], then set
+                  <var>converted value</var> to a JSON string literal containing
+                  said lexical form.
+                </li>
+                <li>
+                  Otherwise, if <a>lexical form</a> expresses a non-numeric value,
+                  such as <code>"+INF"^^xsd:double</code>, an
+                  <a data-link-for="JsonLdErrorCode">invalid JSON literal</a>
+                  error has been detected and processing is aborted.
+                </li>
```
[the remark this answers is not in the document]
gkellogg: Any specific range examples should be careful to be informative, to not get ahead of XSD. Also, it will need to be consistent with JCS.

It might be that they way to do this is to just use JCS for serializing strings, and require that the result of serializing to JCS must result that is in the value space of xsd:integer or xsd:double, respectively.

tests/fromRdf-manifest.jsonld
```diff
@@ -195,6 +195,26 @@
...
+    }, {
+      "@id": "#t0028",
+      "@type": ["jld:NegativeEvaluationTest", "jld:FromRDFTest"],
+      "name": "native types flag set to true is ignored if the number provided is invalid in JSON",
+      "purpose": "+INF^^xsd:double causes a failure to serialize in JSON.",
+      "option": {
+        "useNativeTypes": true
+      },
+      "input": "fromRdf/0028-in.nq",
+      "expectErrorCode": "invalid JSON literal"
```
gkellogg, line 217: This is not invalid, it just can't be serialized as a native value.

gkellogg (changes requested):
I agree with @pchampin, but we need to be clear that it needs to be valid both as xsd:double and JCS number, and fallback to the expanded representation.

anatoly-scherbakov:
I believe I was misunderstanding a few things when writing this PR, sorry guys. Instead of reverting much of these changes, I just created another PR: https://github.com/w3c/json-ld-api/pull/625

And I am closing this one; it won't disappear so as to maintain the history, just will be closed.

PR #623, opened 2024-12-14
Fix issue 622

### LINKED ISSUE #622: Issue with Step 6 of `9.4.1 LoadDocumentCallback`
[Current spec](https://www.w3.org/TR/json-ld11-api/#loaddocumentcallback)
[Source code](https://github.com/w3c/json-ld-api/blob/c05d6af1a1b13d273a63a79253e4608c8be929c1/index.html#L6405-L6428)

Assume the ideal case, where that the server has responded with `Content-Type: application/ld+json`:
* We skip step 4, since we have the `+json` suffix
* We skip step 5, since `application/ld+json` is specifically excluded
* Then, we reach step 6

Currently this is worded as:
> "Otherwise, the retrieved document's [Content-Type](https://tools.ietf.org/html/rfc2045#section-5) **is** neither application/json, application/ld+json, nor any other media type using a +json suffix as defined in [[RFC6839](https://www.w3.org/TR/json-ld11-api/#bib-rfc6839)]. Reject the promise passing a [loading document failed](https://www.w3.org/TR/json-ld11-api/#dom-jsonlderrorcode-loading-document-failed) error."

I think this is a mistake, and it should read as follows, because of course we do have `application/ld+json`:
> "Otherwise, **if** the retrieved document's [Content-Type](https://tools.ietf.org/html/rfc2045#section-5) is neither application/json, application/ld+json, nor any other media type using a +json suffix as defined in [[RFC6839](https://www.w3.org/TR/json-ld11-api/#bib-rfc6839)], reject the promise passing a [loading document failed](https://www.w3.org/TR/json-ld11-api/#dom-jsonlderrorcode-loading-document-failed) error."

This is almost certainly a mistake because step 7 proceeds with the algorithm, even though the current wording of step 6 implies that it will always fail here.

Happy to put in a PR here.

multimeric:
Closes #622.

***
<a href="https://pr-preview.s3.amazonaws.com/multimeric/json-ld-api/pull/623.html" title="Last updated on Jan 27, 2025, 10:39 PM UTC (e97881b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/623/e9a9a01...multimeric:e97881b.html" title="Last updated on Jan 27, 2025, 10:39 PM UTC (e97881b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6570,7 +6570,8 @@
           <code>application/json</code>,
           <code>application/ld+json</code>,
           nor any other media type using a
-          <code>+json</code> suffix as defined in [[RFC6839]], reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
+          <code>+json</code> suffix as defined in [[RFC6839]],
+          then reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
         <li>Create a new <a>RemoteDocument</a> <var>remote document</var> using
           <var>url</var> as {{RemoteDocument/documentUrl}},
           <var>document</var> as {{RemoteDocument/document}},
--- a/index.html
+++ b/index.html
@@ -6566,12 +6566,11 @@
           <p class="note">The HTTP Link Header is ignored for documents served as <code>application/ld+json</code>,
             <code>text/html</code>, or <code>application/xhtml+xml</code>.</p>
         </li>
-        <li>Otherwise, the retrieved document's <a>Content-Type</a> is neither
+        <li>Otherwise, if the retrieved document's <a>Content-Type</a> is neither
           <code>application/json</code>,
           <code>application/ld+json</code>,
           nor any other media type using a
-          <code>+json</code> suffix as defined in [[RFC6839]].
-          Reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
+          <code>+json</code> suffix as defined in [[RFC6839]], reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
         <li>Create a new <a>RemoteDocument</a> <var>remote document</var> using
           <var>url</var> as {{RemoteDocument/documentUrl}},
           <var>document</var> as {{RemoteDocument/document}},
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -6420,12 +6420,11 @@ <h3>LoadDocumentCallback</h3>
...
             <code>text/html</code>, or <code>application/xhtml+xml</code>.</p>
         </li>
-        <li>Otherwise, the retrieved document's <a>Content-Type</a> is neither
+        <li>Otherwise, if the retrieved document's <a>Content-Type</a> is neither
           <code>application/json</code>,
           <code>application/ld+json</code>,
           nor any other media type using a
-          <code>+json</code> suffix as defined in [[RFC6839]].
-          Reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
+          <code>+json</code> suffix as defined in [[RFC6839]], reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
```
TallTed, line 6427:
```suggestion
          <code>+json</code> suffix as defined in [[RFC6839]],
          reject the <var>promise</var> passing a <a data-link-for="JsonLdErrorCode">loading document failed</a> error.</li>
```

pchampin:
thanks @multimeric for the IPR commitment.
@BigBlueHat @gkellogg we are good to move forward with this PR.

gkellogg (approved):
I added Candidate Correction markup and a log entry.

index.html
```diff
@@ -6903,6 +6912,12 @@ <h2>Change log</h2>
       <a data-cite="JSON-LD11#embedding-json-ld-in-html-documents">Embedding JSON-LD in HTML Documents</a> [[JSON-LD11]]
       for treating script elements as a single document,
       as described in <a href="#change_4">Candidate Correction 4</a>.</li>
+    <li>2025-01-25: Correct some corner cases in transforming RDF Number Literals
+      to JSON numbers when <a data-link-for="JsonLdOptions">useNativeTypes</a> is `true`,
+      as described in <a href="#change_5">Candidate Correction 5</a>.</li>
```
pchampin, line 6917:
```suggestion
```
I suspect this has leaked from another PR, right?
It should probably be removed from that one.

pchampin (changes requested):
There is a spurious change in the "changelog" section, which should be removed from this PR.
Otherwise, LGTM.

index.html
multimeric: This was added in 3b0a95c which I didn't make. Do you want me to fix it?

gkellogg: I added it to simplify a future merge, otherwise, it will lead to a conflict. If we do the PRs in an appropriate order, these will mesh together.

PR #624, opened 2024-12-14
Rename name to purpose and add name for t0041

multimeric:
`purpose` was missing from `t0041`. Some prior discussion here: https://github.com/json-ld/json-ld.org/pull/852.

### DIFF
--- a/tests/expand-manifest.jsonld
+++ b/tests/expand-manifest.jsonld
@@ -291,7 +291,8 @@
     }, {
       "@id": "#t0041",
       "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
-      "name": "@language: null resets the default language",
+      "name": "Language nullification",
+      "purpose": "@language: null resets the default language",
       "input": "expand/0041-in.jsonld",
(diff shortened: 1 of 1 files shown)


tests/expand-manifest.jsonld
```diff
@@ -291,7 +291,8 @@
     }, {
       "@id": "#t0041",
       "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
-      "name": "@language: null resets the default language",
+      "name": "Language nullification",
```
davidlehn, line 294: Is this better than the old test suite name that got dropped: `"@language: null"`?  Fine either way I suppose.

multimeric: Well I think it's clearer, but I'm happy to revert it

gkellogg: I’d keep the name the same, as the purpose field provides that further information.

gkellogg (changes requested):
Thanks for the fix. Change “name” back and I think we can merge quickly.

gkellogg (approved):

PR #625, opened 2025-01-12
(closes #555) Fall back to default logic in `useNativeTypes` mode for RDF numbers which are not JSON numbers

### LINKED ISSUE #555: blind spot in RDF to object conversion

anatoly-scherbakov:
This is my second attempt to fix this issue after https://github.com/w3c/json-ld-api/pull/619.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/625.html" title="Last updated on Jan 27, 2025, 10:44 PM UTC (33a652b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/625/f004001...33a652b.html" title="Last updated on Jan 27, 2025, 10:44 PM UTC (33a652b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5487,13 +5487,16 @@
                     </li>
                     <li>
                       If the conversion is successful, set <var>converted value</var>
-                      to its result.
+                      to its result;
+                    </li>
+                    <li>
+                      Otherwise, set <var>type</var> to <a>datatype IRI</a> of <var>value</var>.
                     </li>
                   </ol>
                 </li>
               </ol>
             </li>
-            <li class="changed">If <a>processing mode</a> is not `json-ld-1.0`,
+            <li class="changed">Otherwise, if <a>processing mode</a> is not `json-ld-1.0`,
               and <var>value</var> is a <a>JSON literal</a>,
               set <var>converted value</var> to the result of
               turning the lexical value of <var>value</var>
--- a/index.html
+++ b/index.html
@@ -5477,13 +5477,23 @@
                   <code>xsd:double</code> and its
                   <a>lexical form</a>
                   is a valid <code>xsd:integer</code> or <code>xsd:double</code>
-                  according [[XMLSCHEMA11-2]], set <var>converted value</var>
-                  to the result of converting the
-                  <a>lexical form</a>
-                  to a JSON <a>number</a>.</li>
+                  according to [[XMLSCHEMA11-2]],
+                  <ol>
+                    <li>attempt to convert the <a>lexical form</a> to a <a>JSON number</a>
+                      according to
+                      [[RFC8785]] <a data-cite="RFC8785#name-serialization-of-numbers">
+                        JSON Serialization Scheme (JCS) 3.2.2.3 Serialization of Numbers
+                      </a> procedure;
+                    </li>
+                    <li>
+                      If the conversion is successful, set <var>converted value</var>
+                      to its result.
+                    </li>
+                  </ol>
+                </li>
               </ol>
             </li>
-            <li class="changed">Otherwise, if <a>processing mode</a> is not `json-ld-1.0`,
+            <li class="changed">If <a>processing mode</a> is not `json-ld-1.0`,
               and <var>value</var> is a <a>JSON literal</a>,
               set <var>converted value</var> to the result of
               turning the lexical value of <var>value</var>
--- a/index.html
+++ b/index.html
@@ -5477,17 +5477,17 @@
                   <code>xsd:double</code> and its
                   <a>lexical form</a>
                   is a valid <code>xsd:integer</code> or <code>xsd:double</code>
-                  according to [[XMLSCHEMA11-2]],
+                  according to [[XMLSCHEMA11-2]]:
                   <ol>
-                    <li>attempt to convert the <a>lexical form</a> to a <a>JSON number</a>
+                    <li>Attempt to convert the <a>lexical form</a> to a <a>JSON number</a>
                       according to
                       [[RFC8785]] <a data-cite="RFC8785#name-serialization-of-numbers">
                         JSON Serialization Scheme (JCS) 3.2.2.3 Serialization of Numbers
-                      </a> procedure;
+                      </a> procedure.
                     </li>
                     <li>
                       If the conversion is successful, set <var>converted value</var>
-                      to its result;
+                      to its result.
                     </li>
                     <li>
                       Otherwise, set <var>type</var> to <a>datatype IRI</a> of <var>value</var>.
--- a/index.html
+++ b/index.html
@@ -5458,7 +5458,7 @@
             <li>If <a data-link-for="JsonLdOptions">useNativeTypes</a> is <code>true</code>
               <div class="candidate correction" id="change_5">
                 <span class="marker">Candidate Correction 5</span>
-                <p>This changes the behavior of using native numbers when <a data-link-for="JsonLdOptions">useNativeTypes</a> is `true`.
+                <p>This changes behavior when using native numbers where <a data-link-for="JsonLdOptions">useNativeTypes</a> is `true`.
                   For more information, refer to <a href="https://github.com/w3c/json-ld-api/issues/555">issue 555</a>.
                 </p>
               </div>
@@ -5492,7 +5492,7 @@
                     <ol>
                       <li>
                         Attempt to convert the <a>lexical form</a> to a <a>JSON number</a>
-                        according to
+                        according to the
                         [[RFC8785]] <a data-cite="RFC8785#name-serialization-of-numbers">
                           JSON Serialization Scheme (JCS) 3.2.2.3 Serialization of Numbers
                         </a> procedure.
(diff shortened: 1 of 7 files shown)


index.html
```diff
@@ -5477,10 +5477,20 @@ <h3>Algorithm</h3>
...
+                    <li>attempt to convert the <a>lexical form</a> to a <a>JSON number</a>
+                      according to
+                      <a data-cite="RFC8785#name-serialization-of-numbers">
+                        JSON Serialization Scheme (JCS) 3.2.2.3 Serialization of Numbers
+                      </a> procedure;
+                    </li>
+                    <li>
+                      If the conversion is successful, set <var>converted value</var>
+                      to its result.
+                    </li>
```
gkellogg, line 5491: This needs a corresponding "otherwise" to do what 2.5 does, or to explicitly state to use 2.5.

index.html
```diff
@@ -5477,10 +5477,20 @@ <h3>Algorithm</h3>
...
-                  according [[XMLSCHEMA11-2]], set <var>converted value</var>
-                  to the result of converting the
-                  <a>lexical form</a>
-                  to a JSON <a>number</a>.</li>
+                  according to [[XMLSCHEMA11-2]],
+                  <ol>
+                    <li>attempt to convert the <a>lexical form</a> to a <a>JSON number</a>
+                      according to
+                      <a data-cite="RFC8785#name-serialization-of-numbers">
+                        JSON Serialization Scheme (JCS) 3.2.2.3 Serialization of Numbers
```
gkellogg, line 5485: Needs an explicit `[[RFC8785]]` outside the anchor.

gkellogg (changes requested):
Minor changes, but this direction seems like the way to go.

index.html
anatoly-scherbakov: What if we remove the "Otherwise" at 2.5?..

gkellogg: > What if we remove the "Otherwise" at 2.5?..

That would mean that both 2.4 and 2.8 could be processed for something like `xsd:integer`.

I think we need a 2.4.3.3 which says "Otherwise, set type to the datatype IRI  of value".

We need to keep the "Otherwise, if" in 2.5.

anatoly-scherbakov: Introduced a new point.

index.html
```diff
@@ -5477,10 +5477,23 @@ <h3>Algorithm</h3>
...
+                      [[RFC8785]] <a data-cite="RFC8785#name-serialization-of-numbers">
+                        JSON Serialization Scheme (JCS) 3.2.2.3 Serialization of Numbers
+                      </a> procedure;
+                    </li>
+                    <li>
+                      If the conversion is successful, set <var>converted value</var>
+                      to its result;
+                    </li>
+                    <li>
+                      Otherwise, set <var>type</var> to <a>datatype IRI</a> of <var>value</var>.
```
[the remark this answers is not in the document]
anatoly-scherbakov: Adjusted.

TallTed (approved):

gkellogg:
There’s additional markup needed to show the controls for candidate change ins/del.

gkellogg (approved):
I added the other markup necessary, and called it Candidate Correction 5.

index.html
```diff
@@ -5477,30 +5483,30 @@ <h3>Algorithm</h3>
...
-                  according to [[XMLSCHEMA11-2]]:
-                  <ol>
-                    <li>
-                      <ins>
+                  according to [[XMLSCHEMA11-2]]<del cite="#change_5">,
+                    set <var>converted value</var>
+                    to the result of converting the
+                    <a>lexical form</a>
+                    to a JSON <a>number</a>.
+                  </del><ins cite="#change_5">:
```
anatoly-scherbakov, line 5491: :thinking: I thought `<ins>` cannot wrap an `<ol>`, can it?

gkellogg: The validator does not flag it, and I don't know why it shouldn't be able to wrap `ol`.

anatoly-scherbakov:
Thanks for reviews!

I still can't merge the PR though :thinking: 

image

gkellogg:
Needs a review from @pchampin.

tests/fromRdf/0027-out.jsonld
```diff
@@ -0,0 +1,60 @@
...
+    {
+      "@id": "http://example.com/number-native",
+      "http://example.com/example": [
+        {
+          "@type": "http://www.w3.org/2001/XMLSchema#integer",
+          "@value": 1
+        },
+        {
+          "@type": "http://www.w3.org/2001/XMLSchema#decimal",
+          "@value": 1.1
```
pchampin, line 38: This causes issues, because it round-trips to `"1.1E0"^^xsd:decimal`, which is ill-formed (xsd:decimal does not support scientific notation in its lexical space)

pchampin, line 38:
```suggestion
```
This must be removed for two reasons:

* the spec only tells to convert `xsd:integer` and `xsd:double`, and for a good reason (see below), so this part of the test is actually not complying with the spec;

* `xsd:decimal`s should never be converted to JSON native number because they do not round-trip: this one is converted back to "1.1E0"^^xsd:decimal, which is ill-formed (xsd:decimal does not support scientific notation in its lexical space).

tests/fromRdf/0027-in.nq
```diff
@@ -0,0 +1,12 @@
+<http://example.com/boolean-native> <http://example.com/example> "true"^^<http://www.w3.org/2001/XMLSchema#boolean> .
+<http://example.com/boolean-native> <http://example.com/example> "false"^^<http://www.w3.org/2001/XMLSchema#boolean> .
+
+<http://example.com/boolean-object> <http://example.com/example> "True"^^<http://www.w3.org/2001/XMLSchema#boolean> .
+<http://example.com/boolean-object> <http://example.com/example> "False"^^<http://www.w3.org/2001/XMLSchema#boolean> .
+
+<http://example.com/number-native> <http://example.com/example> "1"^^<http://www.w3.org/2001/XMLSchema#integer> .
+<http://example.com/number-native> <http://example.com/example> "1.1"^^<http://www.w3.org/2001/XMLSchema#decimal> .
```
pchampin, line 8:
```suggestion
```

Must be removed (see my comment on `0027-out.jsonld` below).

pchampin (approved):
Sorry for the delay. I'm approving this PR because the spec text now looks good, and to avoid more delay in back-and-forth...

However, the new test must be amended per my comment, as currently it is both buggy and non-compliant.

index.html
```diff
@@ -5456,6 +5456,12 @@ <h3>Algorithm</h3>
             <li>Initialize <var>converted value</var> to <var>value</var>.</li>
             <li>Initialize <var>type</var> to <code>null</code></li>
             <li>If <a data-link-for="JsonLdOptions">useNativeTypes</a> is <code>true</code>
+              <div class="candidate correction" id="change_5">
+                <span class="marker">Candidate Correction 5</span>
+                <p>This changes the behavior of using native numbers when <a data-link-for="JsonLdOptions">useNativeTypes</a> is `true`.
```
TallTed, line 5461:
```suggestion
                <p>This changes behavior when using native numbers where <a data-link-for="JsonLdOptions">useNativeTypes</a> is `true`.
```

index.html
```diff
@@ -5477,10 +5483,31 @@ <h3>Algorithm</h3>
...
+                  according to [[XMLSCHEMA11-2]]<del cite="#change_5">,
+                    set <var>converted value</var>
+                    to the result of converting the
+                    <a>lexical form</a>
+                    to a JSON <a>number</a>.
+                  </del><ins cite="#change_5">:
+                    <ol>
+                      <li>
+                        Attempt to convert the <a>lexical form</a> to a <a>JSON number</a>
+                        according to
```
TallTed, line 5495:
```suggestion
                        according to the
```

TallTed (changes requested):
Tiny language fixes

anatoly-scherbakov:
@TallTed  @pchampin thanks for the reviews!

@pchampin I've applied the suggestions; indeed, the spec makes no mention of `xsd:decimal`, I do not exactly recall why I even added it to the test.

TallTed (approved):

Issue #630, opened 2025-02-01
Inverse context creation, 4.3.2.3): what are the keys and values in active context?


ZLghkA:
Hello,

In Inverse Context Creation (4.3.2), step 3 iterates over “key term and value term definition in the active context”. What does it mean? In 4.1, an active context is a product type, and the term definitions is one of the fields, as an array of term definitions.

Best regards,

Vivien

w3cbot:
This was discussed during the [#json-ld meeting on 26 February 2025](https://www.w3.org/2025/02/26-json-ld-minutes.html#989d).

<details><summary><i>View the transcript</i></summary>

<h4><a href="https://github.com/w3c/json-ld-api/issues/630" rel="noopener noreferrer">w3c/json-ld-api#630</a></h4>
<p><cite>&lt;gb&gt;</cite> <strong><a href="https://github.com/w3c/json-ld-api/issues/630" rel="noopener noreferrer">Issue 630</a></strong> Inverse context creation, 4.3.2.3): what are the keys and values in active context? (by ZLghkA)</p>

<hr /></details>

pchampin:
Hi Vivien,

you correctly assumed that this iteration of key-values concerns the *terms definitions* of the active context (as hinted by the name of the value). But...

> the term definitions is one of the fields, as an array of term definitions.

This is indeed confusing. This should say that this is a map. And the [definition](https://www.w3.org/TR/json-ld11-api/#dfn-term-definition) of "term definition" in the terminology section should account for their internal representation described  in [Sec 4.1](https://www.w3.org/TR/json-ld11-api/#context-processing-algorithm). Currently, it only describes their JSON/INFRA representation (which is applicable to "local context", but not to "active context", and yet the definition of "active context" points there).

vivien35af:
Hello! 

I have just noticed that json-ld syntax has definitions for a context and expanded term definition (9.15, 9.15.1) that look very similar to the active context and term definition. Are they intended to be the same thing?

Best regards,
Vivien

PR #638, opened 2025-02-27
improve 'term definition' definition

pchampin:
I'm not entirely satisfied with the addition in common/terms.html, as it seems too detailed for the terminology section. But...

...as pointed out in the comment of #630,
the API spec references to this definition in multiple places, and many of those links are actually for instances of the internal representation, *not* for instances of the strings or maps that represent term definition in JSON.

So the alternative would be to introduce a new name for those "internal term definition" and patch all the algorithms... which seems laborious and error-prone.

@gkellogg I don't remember what's the process for synchronizing the files in common with other specs.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/638.html" title="Last updated on Jun 4, 2025, 5:02 PM UTC (b037802)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/638/55504a9...b037802.html" title="Last updated on Jun 4, 2025, 5:02 PM UTC (b037802)">Diff</a>

### DIFF
--- a/common/terms.html
+++ b/common/terms.html
@@ -383,7 +383,7 @@
     Its value is either a string (<dfn data-cite="JSON-LD11#dfn-simple-term-definition" data-lt="simple term definition|simple term" data-lt-noDefault>simple term definition</dfn>),
     expanding to an <a>IRI</a>,
     or a map (<a>expanded term definition</a>).<br/>
-    For <a data-cite="JSON-LD11-API#context-processing-algorithms">context processing</a>, <a>term definitions</a>' values are converted internally to a dedicated data structure that is easier to process.
+    For <a data-cite="JSON-LD11-API#context-processing-algorithms">context processing</a>, <a>term definition</a> values are converted internally to a dedicated data structure that is easier to process.
   </dd>
   <dt class="changed"><dfn data-cite="JSON-LD11#dfn-type-map">type map</dfn></dt><dd class="changed">
     A <a>type map</a> is a <a>map</a> value of a <a>term</a>
--- a/common/terms.html
+++ b/common/terms.html
@@ -382,8 +382,9 @@
     as a key, type, or elsewhere that a string is interpreted as a vocabulary item.
     Its value is either a string (<dfn data-cite="JSON-LD11#dfn-simple-term-definition" data-lt="simple term definition|simple term" data-lt-noDefault>simple term definition</dfn>),
     expanding to an <a>IRI</a>,
-    or a map (<a>expanded term definition</a>).<br/>
-    For <a data-cite="JSON-LD11-API#context-processing-algorithms">context processing</a>, <a>term definition</a> values are converted internally to a dedicated data structure that is easier to process.
+    or a map (<a>expanded term definition</a>).
+    <ins cite="#change_api_638"><br/>
+    For <a data-cite="JSON-LD11-API#context-processing-algorithms">context processing</a>, <a>term definition</a> values are converted internally to a dedicated data structure that is easier to process.</ins>
   </dd>
   <dt class="changed"><dfn data-cite="JSON-LD11#dfn-type-map">type map</dfn></dt><dd class="changed">
     A <a>type map</a> is a <a>map</a> value of a <a>term</a>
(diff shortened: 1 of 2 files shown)


common/terms.html
```diff
@@ -382,7 +382,8 @@
     as a key, type, or elsewhere that a string is interpreted as a vocabulary item.
     Its value is either a string (<dfn data-cite="JSON-LD11#dfn-simple-term-definition" data-lt="simple term definition|simple term" data-lt-noDefault>simple term definition</dfn>),
     expanding to an <a>IRI</a>,
-    or a map (<a>expanded term definition</a>).
+    or a map (<a>expanded term definition</a>).<br/>
+    For <a data-cite="JSON-LD11-API#context-processing-algorithms">context processing</a>, <a>term definitions</a>' values are converted internally to a dedicated data structure that is easier to process.
```
TallTed, line 386:
```suggestion
    For <a data-cite="JSON-LD11-API#context-processing-algorithms">context processing</a>, <a>term definition</a> values are converted internally to a dedicated data structure that is easier to process.
```

gkellogg:
> @gkellogg I don't remember what's the process for synchronizing the files in common with other specs

Just needs to be copied. The GH Action checks out common and compares the results.

pchampin:
This one probably also needs to be formatted as a candidate amendment...

dlongley (approved):

common/terms.html
```diff
@@ -383,6 +383,8 @@
     Its value is either a string (<dfn data-cite="JSON-LD11#dfn-simple-term-definition" data-lt="simple term definition|simple term" data-lt-noDefault>simple term definition</dfn>),
     expanding to an <a>IRI</a>,
     or a map (<a>expanded term definition</a>).
+    <ins cite="#change_api_638"><br/>
+    For <a data-cite="JSON-LD11-API#context-processing-algorithms">context processing</a>, <a>term definition</a> values are converted internally to a dedicated data structure that is easier to process.</ins>
```
gkellogg, line 387: The problem with this is that that will show up in syntax and framing, as well.

index.html
```diff
@@ -1071,10 +1071,19 @@ <h2>RDF Serialization/Deserialization</h2>
...
+        Clarify the notion <a>term definition</a>.
+        For more information, refer to <a href="https://github.com/w3c/json-ld-api/issues/630">issue 630</a>.
+      </p>
+    </div>
+
     <p>The <a>active context</a> consists of:</p>
     <ul>
       <li>the active <a>term definitions</a> which specify how
-        keys and values have to be interpreted (<a>array</a> of <a>term definitions</a>),</li>
+        keys and values have to be interpreted (<del cite="#change_api_638"><a>array</a></del><ins cite="#change_api_638"><a>map</a></ins> of <a>term definitions</a>),</li>
```
gkellogg, line 1086: We probably need to say what the map key is for each entry (the term itself, of course).

index.html
```diff
@@ -1071,10 +1071,19 @@ <h2>RDF Serialization/Deserialization</h2>
       using information provided by the <a>active context</a>. This
       section describes how to produce an <a>active context</a>.</p>
 
+    <div id="change_api_638" class="candidate correction">
+      <span class="marker">Candidate Correction 8</span>
+      <p>Change the type of <a>term definition</a> from <a>array</a> to <a>map</a> in <a>active context</a>,
+        to be consistent with the way it is used in the algorithms.
+        Clarify the notion <a>term definition</a>.
```
TallTed, line 1078:
```suggestion
        Clarify the notion of "<a>term definition</a>".
```

common/terms.html
gkellogg: We'll need issues for those other specs to have their own version of this update.

gkellogg: I created https://github.com/w3c/json-ld-syntax/issues/461 to track this in Syntax, so I think we can probably merge this now.

pchampin:
@gkellogg I actually don't see any markup issue there; the only reported issue is the inconsistency of the common files between the different repos, which is to be expected.

PR #637, opened 2025-02-27
fix #633

pchampin:
This reorders and factorizes some parts of the 'Inverse Content Creation' algorithm.
It does not change the behaviour of the algorithm,
except for one corner case:
if the active context has no default language but has a default base direction,
the default combined tag is now _{dir}.
In the previous version, it was @none_{dir}.
I find this more consistent with steps 3.13.4 and 3.15.1,
and I suspect that this former discrepency was actually a bug.

NB: I put this PR on top of #636 because they affect the same part of the spec,
and because I don't expect #636 to be controversial anyway.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/637.html" title="Last updated on Feb 28, 2025, 8:22 AM UTC (fbd89a5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/637/55504a9...fbd89a5.html" title="Last updated on Feb 28, 2025, 8:22 AM UTC (fbd89a5)">Diff</a>

PR #639, opened 2025-02-28
fix #633, in a better way

### LINKED ISSUE #633: Inverse context creation algorithm: could steps 3.16) and 3.17) be merged?
Hello,

In the inverse context creation algorithm (4.3.2), steps 3.16) and 3.17) look suspiciously similar. Could they be merged? For instance:

3.16.1) Create new variable default-lang-dir.
3.16.2) If active context has a default base direction, set default-lang-dir to the concatenation of default-language and default base direction, separated by '_'.
3.16.3) Otherwise, set default-lang-dir to default-language.
3.16.4) If language map does not have a default-lang-dir entry, create one and set its value to the term being processed.
3.16.5) (verbatim old 3.16.3)
3.16.6) (verbatim old 3.16.4)

Remove 3.17).

What do you think?

Best regards,

Vivien

pchampin:
and using the markup for candidate corrections

This PR deprecates the old PR #637 because
- the old PR did not use candidate correction markup
- the old PR accidentally removed a part of the algorithm that should stay (that's what you get for writing PRs at 1am...)
- the old PR introduced a behavior change in the edge case where default language was undefined, but default base direction was, on the assumption that the old behavior was a bug. On second thought, I'd rather keep this for a separate issue if this is really a bug (if it is, it might also affect other algorithms, e.g. compaction).
- the changes in this PR are simpler, while achieving the same result

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/639.html" title="Last updated on Jun 15, 2025, 8:22 PM UTC (a04703d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/639/525f503...a04703d.html" title="Last updated on Jun 15, 2025, 8:22 PM UTC (a04703d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2007,7 +2007,7 @@
           <div id="change_pr639" class="candidate correction">
             <span class="marker">Candidate Correction 8</span>
             <p>Simplify the algorithm by handling the <a>default base direction</a>
-              once and for all in this step, rather than in each iteration at the end of the next step.
+              once and for all in this step, rather than <a href="#alg-inv-defaul-base-dir">in each iteration</a>.
               For more information, refer to <a href="https://github.com/w3c/json-ld-api/issues/633">issue 633</a>. 
             </p>
           </div>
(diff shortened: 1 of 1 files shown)


gkellogg (approved):
We may want to add a reference to step 3.16 so that the ins/del isn't lost to the viewer.

PR #647, opened 2025-03-25
Fix fromRdf#tli03 test case definition in manifest

marcelotto:
The `input` and `expect` are the same as in `tli02`, while `fromRdf/li03-in.nq` and `fromRdf/li03-out.jsonld` are unused. Seems like a copy-and-paste error.

### DIFF
--- a/tests/fromRdf-manifest.jsonld
+++ b/tests/fromRdf-manifest.jsonld
@@ -410,8 +410,8 @@
       "@type": ["jld:PositiveEvaluationTest", "jld:FromRDFTest"],
       "name": "t0008 as interpreted for 1.1. ",
       "purpose": "List of lists",
-      "input": "fromRdf/li02-in.nq",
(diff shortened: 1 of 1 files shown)


anatoly-scherbakov:
Hi @marcelotto, indeed it seems a copy-paste error, thanks for this fix!

gkellogg (approved):

marcelotto:
The `stefanzweifel/git-auto-commit-action` limitation hitting again: https://github.com/w3c/rdf-canon/pull/201#issuecomment-1973619627

gkellogg:
> The `stefanzweifel/git-auto-commit-action` limitation hitting again: [w3c/rdf-canon#201 (comment)](https://github.com/w3c/rdf-canon/pull/201#issuecomment-1973619627)

If someone figures out how to not trigger that part of the workflow on remote repositories it would clean this up, but it gets run when merged, so the necessary files end up getting updated.

Issue #648, opened 2025-04-05
Error in fromRdf/0027-out.jsonld


marcelotto:
The new `fromRdf#t0027` test expects its result to be wrapped in a `@graph` container. Maybe I'm missing something, but that seems wrong to me.

anatoly-scherbakov:
Hi @marcelotto, thanks for looking into this!

1. I reviewed this against [8.4 Serialize RDF as JSON-LD Algorithm
](https://w3c.github.io/json-ld-api/#serialize-rdf-as-json-ld-algorithm), I think you are right. Will check this out.

2. [8.6 Data Round Tripping](https://w3c.github.io/json-ld-api/#data-round-tripping) seems to override conversion of RDF literals into `value objects` indeed.

I will prepare a PR, we will maybe discuss this on the next meeting.

anatoly-scherbakov:
For reference, I think this is the PR https://github.com/w3c/json-ld-api/pull/625

w3cbot:
This was discussed during the [#json-ld meeting on 09 April 2025](https://www.w3.org/2025/04/09-json-ld-minutes.html#a9be).

<details><summary><i>View the transcript</i></summary>

<h4><a href="https://github.com/w3c/json-ld-api/issues/648" rel="noopener noreferrer">w3c/json-ld-api#648</a></h4>
<p><cite>&lt;gb&gt;</cite> <strong><a href="https://github.com/w3c/json-ld-api/issues/648" rel="noopener noreferrer">Issue 648</a></strong> Error in fromRdf/0027-out.jsonld (by marcelotto)</p>
<p><cite>anatoly-scherbakov:</cite> when I was converting literals into JSON-LD<br>
<span>… when they used native types, I was converting them to Value Objects</span><br>
<span>… but that's not what the spec is going for</span><br>
<span>… the spec says to use native JSON types</span><br>
<span>… and I made a PR to address that</span></p>
<p><a href="https://github.com/w3c/json-ld-api/pull/650" rel="noopener noreferrer">w3c/<wbr>json-ld-api#650</a></p>
<p><cite>&lt;gb&gt;</cite> <strong><a href="https://github.com/w3c/json-ld-api/pull/650" rel="noopener noreferrer">Pull Request 650</a></strong> #648 `0027-out.jsonld`: `<code>@graph</code>` and value objects (by anatoly-scherbakov)</p>
<p><cite>gkellogg:</cite> the first sentence was about `<code>@graph</code>`<br>
[...]

PR #650, opened 2025-04-06
#648 `0027-out.jsonld`: `@graph` and value objects

anatoly-scherbakov:


tests/fromRdf/0027-out.jsonld
```diff
@@ -1,56 +1,45 @@
...
-      ]
-    }
-  ]
-}
+[
+  {
+    "@id": "http://example.com/boolean-native",
+    "http://example.com/example": [
+      true,
+      false
```
gkellogg, line 6: The algorithm still places these into value objects.

```suggestion
      {"@value": true},
      {"@value": false}
```

tests/fromRdf/0027-out.jsonld
```diff
@@ -1,56 +1,45 @@
...
+      {
+        "@type": "http://www.w3.org/2001/XMLSchema#boolean",
+        "@value": "False"
+      }
+    ]
+  },
+  {
+    "@id": "http://example.com/number-native",
+    "http://example.com/example": [
+      1
```
gkellogg, line 25:
```suggestion
      {"@value": 1}
```

gkellogg (changes requested):
If using native values, the RDF to Object Conversion algorithm still places these in value objects.

pchampin (changes requested):
I believe tha the changes in fromRdf-manifest.html should be reverted.
Otherwise, I approve the changes to 0027-out.jsonld .

PR #652, opened 2025-04-10
Add test report for JSON-LD.ex

marcelotto:
This PR adds a test report for the new JSON-LD.ex version 1.0, which implements the JSON-LD 1.1 specification.

Tests marked with `"specVersion": "json-ld-1.0"` were skipped. The only failing tests require either generalized RDF (not supported by RDF.ex) or the HTML Content Algorithm (not yet implemented).

Some tests produce correct values but in a different order. These are marked as passing with `earl:mode earl:semiAuto`.


gkellogg (approved):

marcelotto:
I noticed that on https://json-ld.org/, JSON-LD.ex is not yet listed as supporting JSON-LD 1.1. 
Is there a specific process to request such an update, or could you direct me to the appropriate contact?

gkellogg:
> I noticed that on https://json-ld.org/, JSON-LD.ex is not yet listed as supporting JSON-LD 1.1. Is there a specific process to request such an update, or could you direct me to the appropriate contact?

A PR in https://github.comjson-ld/json-ld.org should do it.

PR #665, opened 2025-08-18
replace sequence by mf:includes based path

MaillPierre:
The [main manifest](https://w3c.github.io/json-ld-api/tests/manifest.jsonld) uses "sequence" to list the other sub-manifests used for the test suite. 
"sequence" uses the property [mf:entries](https://www.w3.org/2001/sw/DataAccess/tests/test-manifest#entries), which is described as being used to link manifests to entries. To link manifests to other manifests, [mf:include](https://www.w3.org/2001/sw/DataAccess/tests/test-manifest#include) should be used.

This pull request adds "inclusion" to the context as a keyword similar to "sequence" but using mf:include. It also replaces "sequence" by "inclusion" in the main manifest.

### DIFF
--- a/tests/context.jsonld
+++ b/tests/context.jsonld
@@ -14,7 +14,7 @@
     "input":           { "@id": "mf:action", "@type": "@id" },
     "option":          { "@type": "@id"},
     "sequence":        { "@id": "mf:entries", "@type": "@id", "@container": "@list" },
-    "inclusion":        { "@id": "mf:includes", "@type": "@id", "@container": "@list" },
+    "inclusion":        { "@id": "mf:include", "@type": "@id", "@container": "@list" },
     "redirectTo":      { "@type": "@id"},
 
     "name":                 "mf:name",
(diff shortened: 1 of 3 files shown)


davidlehn:
- Where is the spec for the vocab?  The use here and in related specs might be a bit fuzzy.  I assume everyone processes these manifests as plain JSON, at least initially, to bootstrap having a JSON-LD processor.
- I recall the data mistakenly used "sequence" when it was first written, and was later mapped to "entries", but kept as the same name to avoid having to change every test suite and implementation.
- I always assumed "entries" values could be either a test or another manifest.  Is a manifest an entry?  I think that makes it a nice tree structure to process and switch based on type.  And if it's a string, assume you dereference that and process what you get.  Just like other JSON-LD.  But I've never looked at docs on any of this!
- Why "inclusion" instead of "includes"?
- If this changes, it will likely break every single implementation.  Might need to consider when trading off letting it stay slightly incorrect and documenting the issue.

pchampin:
@davidlehn I don't know if there *is* a spec for the test vocab, but the source is rather clear:

  http://www.w3.org/2001/sw/DataAccess/tests/test-manifest

`entries` is a list of entries, `included` is a list of manifests.
So I think this change makes sense.

pchampin (approved):

PR #662, opened 2025-06-14
(closes #656) convert `"1"^^xsd:boolean` and `"0"^^xsd:boolean` to native values

anatoly-scherbakov:
image

- **#656 Specify 0 and 1 as valid Boolean literals in RDF to Object conversion algorithm**
- **#656 Update fromRdf test 0027 to reflect 0 and 1 as valid Boolean values**

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/662.html" title="Last updated on Jun 20, 2025, 6:32 PM UTC (754c657)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/662/f32a438...754c657.html" title="Last updated on Jun 20, 2025, 6:32 PM UTC (754c657)">Diff</a>

Issue #656, opened 2025-05-08
convert "1"^^xsd:boolean and "0"^^xsd:boolean to native values


pchampin:
In the [RDF to Object conversion algorithm], when the flag `useNativeTypes` is true, literals of type `xsd:boolean`, `xsd:integer` and `xsd:double` are converted (when possible) to the corresponding JSON value.

Since `1` and `0` are both part of the [lexical space](https://www.w3.org/TR/xmlschema11-2/#boolean-lexical-mapping) of `xsd:boolean`, it would make sense to recognize them in step 2.4.2 of the [RDF to Object conversion algorithm], rewording it as follows:

* Otherwise, if the [datatype IRI](https://www.w3.org/TR/rdf11-concepts/#dfn-datatype-iri) of <var>value</var> equals `xsd:boolean`, set <var>converted value</var> to `true` if the [lexical form](https://www.w3.org/TR/json-ld11-api/#dfn-lexical-form) of value matches `"true"`<ins> or `"1"`</ins>, or to false if it matches `"false"`<ins> or `"0"`</ins>. If it matches neither, set <var>type</var> to `xsd:boolean`.

(NB: I also added quotes around the lexical values for clarity)

[RDF to Object conversion algorithm]: https://www.w3.org/TR/json-ld11-api/#x8-5-2-algorithm

PR #669, opened 2025-10-29
Improve test fromRdf#t0027.

davidlehn:
- Improving the test coverage for use useNativeTypes algorithm in 8.5.2 2.4.
- Remove the "True" and "False" boolean tests that can cause confusion that those words are special values.
- Add tests using a value of "notnative" typed to boolean, integer, and double. These test that the type will be preserved when a native conversion does not happen.

I was fixing jsonld.js for fromRdf#t0027 and thought the tests could use some work to increase covearge.

https://github.com/w3c/json-ld-api/blob/main/tests/fromRdf/0027-in.nq
https://github.com/w3c/json-ld-api/blob/main/tests/fromRdf/0027-out.jsonld

I initially mistakenly tried to add specific support for the "True"/"False" tests:
```nq
<http://example.com/boolean-object> <http://example.com/example> "True"^^<http://www.w3.org/2001/XMLSchema#boolean> .
<http://example.com/boolean-object> <http://example.com/example> "False"^^<http://www.w3.org/2001/XMLSchema#boolean> .
```
But those are, I think, just testing that any not-native value is passed through even though it may not match semantics of the type.  I think some other value text should be used to avoid confusion.  For consistency, it probably makes sense to have similar tests for integers and doubles.  jsonld.js was dropping the type for non-integers typed as integers.

There might be other edge cases that should be tested here too. This does seem like an incremental improvement.

Possible relevant references:
https://w3c.github.io/json-ld-api/#algorithm-16
https://github.com/w3c/json-ld-api/issues/555
https://github.com/w3c/json-ld-api/issues/656
https://github.com/w3c/json-ld-api/pull/619
https://github.com/w3c/json-ld-api/pull/625
https://github.com/w3c/json-ld-api/pull/662

### DIFF
--- a/tests/fromRdf/0027-in.nq
+++ b/tests/fromRdf/0027-in.nq
@@ -6,6 +6,7 @@
 
 <http://example.com/boolean-object> <http://example.com/example> "True"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 <http://example.com/boolean-object> <http://example.com/example> "False"^^<http://www.w3.org/2001/XMLSchema#boolean> .
+# NB: the two values above ("True" and "False") are *not* in the lexical space of xsd:bool, so they should be treated as any other invalid lexical form (like "notnative" below)
 <http://example.com/boolean-object> <http://example.com/example> "notnative"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 
 <http://example.com/number-native> <http://example.com/example> "1"^^<http://www.w3.org/2001/XMLSchema#integer> .
--- a/tests/fromRdf/0027-in.nq
+++ b/tests/fromRdf/0027-in.nq
@@ -4,6 +4,8 @@
 <http://example.com/boolean-number> <http://example.com/example> "1"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 <http://example.com/boolean-number> <http://example.com/example> "0"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 
+<http://example.com/boolean-object> <http://example.com/example> "True"^^<http://www.w3.org/2001/XMLSchema#boolean> .
+<http://example.com/boolean-object> <http://example.com/example> "False"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 <http://example.com/boolean-object> <http://example.com/example> "notnative"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 
 <http://example.com/number-native> <http://example.com/example> "1"^^<http://www.w3.org/2001/XMLSchema#integer> .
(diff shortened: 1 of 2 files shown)


filip26:
I see it now, the original test suite only considered `"true"` and `"false"` as valid values for `xsd:boolean`, but since `xsd:boolean` also accepts `"0"` and `"1"`, the test was later updated accordingly.

 I would keep the `"True"` and `"False"` tests, as they ensure that no implementation is performing a case-insensitive comparison like:

 `"true".equalsIgnoreCase(object)`

which could incorrectly coerce capitalized `"True"` or `"False"` into the native boolean type. These values are not valid according to the `xsd:boolean` definition and therefore should remain as-is, without conversion.

davidlehn:
> I would keep the `"True"` and `"False"` tests, as they ensure that no implementation is performing a case-insensitive comparison like: [...]

Ok.  More tests are always good.  Hopefully others don't get confused like I did and try to special case those values.

anatoly-scherbakov (approved):

dlongley (approved):

niklasl (approved):
LGTM.

Note though, that in RDF 1.2, the requirement in RDF 1.1 ([3.3, bullet point 2 b](https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal)) to produce graphs containing ill-typed values is relaxed from a MUST to a SHOULD in the upcoming RDF 1.2 ([3.4.2, bullet point 3.2](https://www.w3.org/TR/rdf12-concepts/#literal-value)). So an RDF 1.2 processor *might* drop those `True`|`False`|`notnative` tokens since they aren't in their respective value spaces. (But they *shouldn't*...)

tests/fromRdf/0027-in.nq
```diff
@@ -6,9 +6,13 @@
 
 <http://example.com/boolean-object> <http://example.com/example> "True"^^<http://www.w3.org/2001/XMLSchema#boolean> .
 <http://example.com/boolean-object> <http://example.com/example> "False"^^<http://www.w3.org/2001/XMLSchema#boolean> .
+<http://example.com/boolean-object> <http://example.com/example> "notnative"^^<http://www.w3.org/2001/XMLSchema#boolean> .
```
pchampin, line 10: I propose to add a comment to explain the rationale of testing "True" and "False",
and defusing the misunderstanding that they should be treated specifically.

```suggestion
# NB: the two values above ("True" and "False") are *not* in the lexical space of xsd:bool, so they should be treated as any other invalid lexical form (like "notnative" below)
<http://example.com/boolean-object> <http://example.com/example> "notnative"^^<http://www.w3.org/2001/XMLSchema#boolean> .
```

pchampin (approved):
NB: my proposed change is just a suggestion, to respond to @davidlehn 's concern.
But I'm happy with the PR as it is.

TallTed (approved):
With PA's comment

PR #676, opened 2026-01-14
Remove `<ins>` and `<del>` markup in preparation for 1.2

niklasl:
This removes all `<ins>` and `<del>` markup in the document. There may still be artefacts left that relate to these, so we may need to think about the scope of this PR (see comments).

If I got it right, we want this in preparation for JSON-LD 1.2 working drafts, along with other important changes (title, correct "What's new in 1.2", change use of `class="changed"`...). Do we want these other changes in separate PR:s, or a bigger branch for all the prep work?

***
<a href="https://pr-preview.s3.amazonaws.com/niklasl/json-ld-api/pull/676.html" title="Last updated on Jun 3, 2026, 4:49 PM UTC (c18662d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/676/590f78d...niklasl:c18662d.html" title="Last updated on Jun 3, 2026, 4:49 PM UTC (c18662d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5479,7 +5479,7 @@
                   
                     of <var>value</var> matches <code>"true"</code> or <code>"1"</code>, or <code>false</code>
                     if it matches <code>"false"</code> or <code>"0"</code>.
-                  </ins>
+                  
                   If it matches neither,
                   set <var>type</var> to <code>xsd:boolean</code>.</li>
                 <li>Otherwise, if the
--- a/index.html
+++ b/index.html
@@ -1083,7 +1083,7 @@
     <p>The <a>active context</a> consists of:</p>
     <ul>
       <li>the active <a>term definitions</a> which specify how
-        keys and values have to be interpreted (<del cite="#change_api_638"><a>array</a></del><ins cite="#change_api_638"><a>map</a></ins> of <a>term definitions</a>),</li>
+        keys and values have to be interpreted (<a>map</a> of <a>term definitions</a>),</li>
       <li>the current <dfn data-lt="context-base-iri" data-lt-noDefault>base IRI</dfn> (<a>IRI</a>),</li>
       <li class="changed">the <dfn>original base URL</dfn> (<a>IRI</a>),</li>
       <li class="changed">an <dfn data-lt="context-inverse" data-lt-noDefault>inverse context</dfn> (<a>inverse context</a>),</li>
@@ -1094,7 +1094,7 @@
         used when a non-propagated <a>context</a> is defined.</li>
     </ul>
 
-    <p>Each <a>term definition</a><ins cite="#change_api_638">'s value</ins> consists of:</p>
+    <p>Each <a>term definition</a>'s value consists of:</p>
     <ul>
       <li>an <dfn>IRI mapping</dfn> (<a>IRI</a>),</li>
       <li class="changed">a <dfn>prefix flag</dfn> (<a>boolean</a>),</li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2133,32 +2131,11 @@ <h3>Algorithm</h3>
...
-                <li>If <var>language map</var> does not have an `@none` <a>entry</a>,
-                  create one and set its value to the <a>term</a>
-                  being processed.</li>
-                <li>If <var>type map</var> does not have an `@none` <a>entry</a>,
-                  create one and set its value to the <a>term</a>
-                  being processed.</li>
-              </ol>
-              </del>
+              <em>(this step was removed by a <a href="#change_pr639">candidate correction</a>)</em>
             </li>
```
niklasl, line 2135: Should this list item be fully removed now?

index.html
```diff
@@ -5515,30 +5488,30 @@ <h3>Algorithm</h3>
...
-                        of <var>value</var>.
-                      </li>
-                    </ol>
-                  </ins>
+                  according to [[XMLSCHEMA11-2]],
+                  set <var>converted value</var>
+                  to the result of converting the
+                  <a>lexical form</a>
+                  to a JSON <a>number</a>.
+                  :
```
niklasl:
```suggestion
                  Then:
```
?

niklasl:
I cannot assign reviewers, so pinging @pchampin @BigBlueHat @anatoly-scherbakov @iherman to check if this is the wanted change and direction.

index.html
davidlehn: That `del` text should be removed.  Probably should be this:
```suggestion
                  according to [[XMLSCHEMA11-2]]:
```

davidlehn:
What's the process here and timeline for this change?  It seems the ins/del markup is useful during the entire 1.2 development, at least in the editors draft.  Is the intent to only remove the text for publication, but otherwise retain until 1.2 is finalized?  Asking because if the markers are only used short term, then why use at all rather than regular text or semantic diffs?

iherman:
> What's the process here and timeline for this change? It seems the ins/del markup is useful during the entire 1.2 development, at least in the editors draft. Is the intent to only remove the text for publication, but otherwise retain until 1.2 is finalized? Asking because if the markers are only used short term, then why use at all rather than regular text or semantic diffs?

It is perfectly possible for editors to use these as part of their work but, afaik, it is hardly ever done. With the lack of proper tools, managing them is the editors' nightmare. (I know it from personal experience.)

The only place where it is required per process is when a Recommendation is "maintained" by the WG, i.e., when a Rec is changed without going back to the traditional WD->CR->AC Vote->Rec route. The "living standard" case, as it were. Because, when there are non-editorial substantive changes, the AC still has to approve the changes, the ins/del tool is used for that sole purpose: get the AC vote on the changes. I guess this is why they are part of the 1.2 drafts in the first place. However, under the new charter, 1.2 will undergo the traditional process and, therefore, the usage of those tags is now moot, and they should disappear.

(That is the reason why the "living standard" model essentially failed in the W3C: it is a pain in the backside. Other groups choose never to publish a Rec, but leave the specification in a perpetual CR stage.)

index.html
pchampin: I think so

pchampin (changes requested):
one `<del>` tag has been treated as an `<ins>`, line 5518 and following (@davidlehn also noted this in a comment).

Apart from that, looking good.

The next step would be to remove the divs with `class="candidate correction"`, and move their text to the corresponding item in section "Change logs".

index.html
niklasl: Yes, certainly! (Accidentally did `dst` instead of `dat` in Vim 😉 .)

pchampin (approved):
I suggest that we merge this PR.
The blocking issues on my side have been fixed (thanks @niklasl),
and the rest can be done afterwards, namely

* remove completely the step 3.16 in the "Inverse Context Creation" algorithm
   (currently it just contains the text "this step has been removed)

* remove all the candidate correction blocks, and move their text into the corresponding items in the Changelog section

index.html
```diff
@@ -5504,8 +5477,8 @@ <h3>Algorithm</h3>
                   <var>converted value</var> to <code>true</code> if the
                   <a>lexical form</a>
                   
-                    of <var>value</var> matches <code>"true"</code><ins cite="#change_5"> or <code>"1"</code></ins>, or <code>false</code>
-                    if it matches <code>"false"</code><ins cite="#change_5"> or <code>"0"</code></ins>.
+                    of <var>value</var> matches <code>"true"</code> or <code>"1"</code>, or <code>false</code>
+                    if it matches <code>"false"</code> or <code>"0"</code>.
                   </ins>
```
anatoly-scherbakov:
```suggestion
                  
```

This `</ins>` seems unmatched?

anatoly-scherbakov (approved):

PR #686, opened 2026-04-11
tests: add expand #t0132 (@base vs property keys)

anatoly-scherbakov:
## Summary

Adds **ExpandTest** `#t0132`: with a context that sets only `@base` (and `@version` 1.1), an unmapped simple property key like `name` is **dropped**, while a relative `@type` token still resolves against that base ([IRI Expansion](https://www.w3.org/TR/json-ld11-api/#iri-expansion)).

## Files

- `tests/expand/0132-in.jsonld`, `tests/expand/0132-out.jsonld`
- `tests/expand-manifest.jsonld` — new sequence entry after `#t0131`
- `tests/expand-manifest.html` — regenerated (Haml + HtmlBeautifier; `bundle exec rake` was not runnable here without libyaml-dev; same pipeline as `tests/Rakefile`)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/pull/686.html" title="Last updated on Apr 17, 2026, 7:40 AM UTC (8fbed33)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/json-ld-api/686/590f78d...8fbed33.html" title="Last updated on Apr 17, 2026, 7:40 AM UTC (8fbed33)">Diff</a>

PR #702, opened 2026-08-03
expansion: add test for default language

### LINKED ISSUE #700: expansion: string @value must not get @language/@direction if type mapping exists
I ended up with a bug in my implementation where Step 5 of value expansion (https://www.w3.org/TR/json-ld11-api/#algorithm-5) always happened if the value was a string, even if a type mapping was present. The test suite lacks coverage for this case, so this bug went unnoticed for almost a year.

A simple reproducer:

```json
{
  "@context": {
    "typed": { "@id": "http://ex/typed", "@type": "http://ex/dt" },
    "@language": "und"
  },
  "typed": "v"
}
```

Expected:
```json
[
  {
    "http://ex/typed": [
      {
        "@type": "http://ex/dt",
        "@value": "v"
      }
    ]
  }
]
```

Due to my bug I ended up with an `"@language": "und"` on `http://ex/typed` as well.

The Pleroma/Akkoma lineage of fediverse instances, as well as Mobilizon, all smash a `{"@language": "und"}` in their context.

daenney:
This adds a test to check that when the context has a default language and we're expanding a string value that has a type mapping, the default language is not added.

It ensures Step 5 in Value Expansion is only performed for string values in the absence of a type mapping.

Fixes: #700

### DIFF
--- a/tests/expand-manifest.jsonld
+++ b/tests/expand-manifest.jsonld
@@ -988,6 +988,14 @@
       "expect": "expand/0131-out.jsonld",
       "option": {"specVersion": "json-ld-1.1"}
     }, {
+      "@id": "#t0133",
+      "@type": [ "jld:PositiveEvaluationTest", "jld:ExpandTest" ],
+      "name": "String value with type mapping and default @language",
+      "purpose": "Verify that in the presence of a type mapping a string value does not receive the default @language from context",
+      "input": "expand/0133-in.jsonld",
+      "expect": "expand/0133-out.jsonld",
+      "option": {"specVersion": "json-ld-1.1"}
+    }, {
       "@id": "#tc001",
       "@type": ["jld:PositiveEvaluationTest", "jld:ExpandTest"],
       "name": "adding new term",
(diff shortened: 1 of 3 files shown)


daenney:
I opted to name this one t0133 to avoid clashing with #686.

daenney:
@anatoly-scherbakov It seems I need to apply to join the WG as an invited expert to be able to join and then sign what the IPR commitments thing is?

anatoly-scherbakov (approved):

anatoly-scherbakov:
@daenney thanks for contributing this!

- I have marked this PR as non-substantive to prevent the `ipr` bot from complaining,
- There was another CI failure, I fixed it in another PR,
- and then I rebased this PR on top of mine to test that CI is now 🟢, sorry for messing with your PR and hope that this does not cause inconvenience!

Now we'll need both of these PRs approved by someone with write access, I added people as reviewers to both.

daenney:
No worries at all, thanks for doing all this!

anatoly-scherbakov:
@daenney I am afraid we accidentally closed this :/ PR stacks over forks do not behave as I intended. Would you be open to raise it again? I am sorry!

TallTed (commented):
lgtm

anatoly-scherbakov:
Or otherwise I can look into recreating it

PR #705, opened 2026-08-05
expand: certain keywords are dropped

daenney:
Step 13.4 of the expansion algorithm handles the case where a key, after IRI expansion, matches a JSON-LD keyword. It enumerates a number of keywords that are specifically handled, and ends with 13.4.17:

> Continue with the next key from element.

From my understanding of it, that should mean that any key which expanded into a keyword we did not just explicitly handle is not added to the expanded document, and therefor its value is never expanded either so can't be present in the expanded document. In effect, we're dropping that key.

One interesting tidbit, this fails in the JSON-LD playground. I think my reading of the spec is correct, so I believe this is an oversight in `jsonld.js`. But I'm not entirely sure. The expanded result on the playground is:

```json
[
  {
    "@base": [
      "base"
    ],
    "@container": [
      "@language"
    ],
    "@json": [
      {}
    ],
    "@none": [
      "none"
    ],
    "@prefix": [
      "prefix"
    ],
    "@preserve": [
      "preserve"
    ],
    "@version": [
      "1.1"
    ],
    "@vocab": [
      "hello"
    ]
  }
]
```


daenney:
Assuming my reading is correct, should the various frame expansion keywords be added to this test as well?

anatoly-scherbakov (approved):
If I understand the point correctly then the keywords in the test should be ignored specifically for the reason that they *[aren't keywords](https://www.w3.org/TR/json-ld11-api/#iri-expansion)* in the [document-root context](https://www.w3.org/TR/json-ld11/#keywords).

Here's a classification of these.

| Keyword | Classification | Valid JSON-LD 1.1 use |
| --- | --- | --- |
| `@base` | [Context keyword](https://www.w3.org/TR/json-ld11/#keywords) | Context definition |
| `@container` | [Term-definition keyword](https://www.w3.org/TR/json-ld11/#keywords) | Expanded term definition |
| `@import` | [Context keyword](https://www.w3.org/TR/json-ld11/#keywords) | Context definition |
| `@json` | [Type keyword](https://www.w3.org/TR/json-ld11/#keywords) | Value of `@type` |
| `@none` | [Map index sentinel](https://www.w3.org/TR/json-ld11/#keywords) | Index, id, language, or type map |
| `@prefix` | [Term-definition keyword](https://www.w3.org/TR/json-ld11/#keywords) | Expanded term definition |
| `@propagate` | [Context keyword](https://www.w3.org/TR/json-ld11/#keywords) | Context definition |
| `@version` | [Context keyword](https://www.w3.org/TR/json-ld11/#keywords) | Context definition |
| `@vocab` | [Context/type keyword](https://www.w3.org/TR/json-ld11/#keywords) | Context definition or term-definition `@type` |
| `@any` | [Algorithm sentinel](https://www.w3.org/TR/json-ld11-api/#iri-compaction) | Inverse-context processing |
| `@null` | [Algorithm sentinel](https://www.w3.org/TR/json-ld11-api/#iri-compaction) | Inverse-context processing |
| `@preserve` | [Framing-processing keyword](https://www.w3.org/TR/json-ld11-framing/#framing-algorithm) | Framing/compaction processing |

Could you please also regenerate and commit `tests/expand-manifest.html`:

```sh
cd tests
bundle exec rake expand-manifest.html
```

This is the generated-file rule in [`tests/Rakefile`](https://github.com/w3c/json-ld-api/blob/main/tests/Rakefile).

Non-blocking: adding `@protected` would complete the coverage of context/term-definition-only keywords.

daenney:
@anatoly-scherbakov all done I think. I added ~~propagate~~ protected for good measure.

daenney:
Sorry, I confugred propagate and protected there. Propagate was already there, I added protected as suggested.

Something I did just notice, `propagate`, regardless of its value, is dropped by the JSON-LD playground. Protected is kept though no matter the value.

PR #703, opened 2026-08-03
compaction: empty array and candidate selection

daenney:
This ensures that in 12.7 in value compaction the empty array is passed through, as it affects candidate selection in step 7.3 of IRI compaction.

PR #707, opened 2026-08-12
compaction: empty array and candidate selection

BigBlueHat:
This ensures that in 12.7 in value compaction the empty array is passed
through, as it affects candidate selection in step 7.3 of IRI
compaction.

---

This is a "redo" of #703 which got closed accidently (with another apology to @daenney)

### DIFF
--- a/tests/compact/0115-context.jsonld
+++ b/tests/compact/0115-context.jsonld
@@ -1,4 +1,6 @@
 {
-  "ex": "http://example.org/",
-  "ex:prop": { "@id": "http://example.org/prop", "@container": "@list" }
+  "@context": {
+    "ex": "http://example.org/",
+    "ex:prop": { "@id": "http://example.org/prop", "@container": "@list" }
+  }
 }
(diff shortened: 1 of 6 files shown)


anatoly-scherbakov (approved):

tests/compact/0115-context.jsonld
```diff
@@ -0,0 +1,4 @@
+{
+  "ex": "http://example.org/",
+  "ex:prop": { "@id": "http://example.org/prop", "@container": "@list" }
+}
```
BigBlueHat, line 6:
```suggestion
{
  "@context": {
    "ex": "http://example.org/",
    "ex:prop": { "@id": "http://example.org/prop", "@container": "@list" }
  }
}
```

tests/compact/0115-context.jsonld
```diff
@@ -0,0 +1,4 @@
+{
```
niklasl, line 1: I believe this is not a correct context document; they too need `@context` at top level.

BigBlueHat:
@daenney was this just a missing test? Or was there something you expected here that didn't happen?

daenney:
@BigBlueHat Just a missing test. I borked it in my implementation, but the test suite passed without raising an issue. I figured it would be nice to have a test.

niklasl (approved):

anatoly-scherbakov (approved):

PR #716, opened 2026-09-06
Add additional tests for @propagate

daenney:
The test suite currently doesn't exercise the handling for `@propagate` on a root context document, it only has tests for it in property scoped contexts. This adds tests that cover `@propagate` handling at the root.


daenney:
I'm actually a little confused about this. In Context Processing, the presence of `@propagate` is checked in step 2, before array-ification in step 4 and any further processing:

> If local context is an object containing the member `@propagate`, its value MUST be [boolean](https://infra.spec.whatwg.org/#boolean) true or false, set propagate to that value.

So, if we start with a context document that is an array:
```json
{
  "@context": [{"@propagate": false}]
}
```

Then in step 2, the condition is false. The context is an array, not an object, so we skip it. `propagate` now remains true (as that is its default), so the rollback step 3 sets up never happens. Now we hit step 4, and arrayify the context document and then continue processing. We'll still hit 5.11 to validate the value of `@propagate`, but it never affects `propagate`.

Both in my implementation and in jsonld.js the `@propagate` value is checked after arrayification, by looking at the first member if that one is a map.

https://github.com/digitalbazaar/jsonld.js/blob/7a93609e475edeebef372ba17ae8e3508b4a324f/lib/context.js#L94-L99

PyLD does too:

https://github.com/digitalbazaar/pyld/blob/517894bfc9c57901aae14d01c1b780abb8b3a136/lib/pyld/jsonld.py#L3287-L3291

dotnetRDF does not, and honours the exact "only if it's an object" wording in the spec: https://github.com/dotnetrdf/dotnetrdf/blob/92948418f567c068ad6374e02e34589f60458972/Libraries/dotNetRdf.Core/JsonLd/Processors/ContextProcessor.cs#L87-L101

trld also honours the "only if it is an object": https://github.com/niklasl/trld/blob/e82bc73a3bec2c0625baf378767c2c26c7376b95/trld/jsonld/context.py#L156-L159

Ruby-rdf seems to also do the exact check, before arrayification: https://github.com/ruby-rdf/json-ld/blob/fe7a963fcb3ffeadb906cef0e2acf6a81d51b872/lib/json/ld/context.rb#L251.

So, a bunch of implementations disagree with each other :sweat_smile:, including some rather important ones. How do we resolve this?

daenney:
In the end I think the fact that it was on the root context or not doesn't really matter here. That was a bug on my side, related to a missing context clone. It didn't surface for a property-scoped context because the clone always happened there.

But a question remains for me: how is `@propagate` meant to work? My guess is both jsonld.js, PyLD and myself independently came to the same optimisation: we can just do the check after. But it would result in a different outcome when faced with an `@context` that is an array.

daenney:
I've been further looking at this, and I found this tidbit: https://www.w3.org/TR/json-ld11/#context-propagation

> Contexts included within an array must all have the same value for `@propagate` due to the way that rollback is defined in [JSON-LD 1.1 Processing Algorithms and API](https://www.w3.org/TR/json-ld11-api/).

This, best I can tell, isn't guaranteed by context processing. It would seem that what we actually need is:
* Arrayify the context.
* For every member that is an object, and for every member that is a remote context once the context document has been retrieved, check if `@propagate` is set to the same value on all context definitions. 

If this is not the case, an invalid local context is returned. I don't think we can return invalid `@propagate` value since the value itself may be fine.

From looking at both the existing tests and the JSON-LD 1.1 text, I get the impression `@propagate` was only envisioned on scoped contexts? And scoped contexts have only been envisioned to take the form of a context definition? But these preconditions are not something that's currently enforced by context processing.

anatoly-scherbakov (commented):
The new manifest entries refer to `c039`, `c040`, and `c041`, while the added files are named `c0039`, `c0040`, and `c0041`. As written, the test resources do not resolve.

There is also a separate specification question in `tc039`. Context Processing reads `@propagate` before arrayification, so the expected result currently reflects behavior implemented by some processors rather than the algorithm. The Context Propagation note exposes the ambiguity, but does not resolve it. A specification decision would be needed before this becomes a conformance test.

daenney:
Yeap, agreed. Should I close this one and raise it as an issue instead, or do we leave it open for discussion here?

anatoly-scherbakov:
I propose to continue discussion here and maybe raise this on a call.

